PR #14, opened 2015-09-21
ADD contract_journal module

angelmoya:


contract_journal/README.rst
```diff
@@ -0,0 +1,53 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+========================
+Contract Account Journal
```
pedrobaeza: Contract Invoice Journal

contract_journal/README.rst
```diff
@@ -0,0 +1,53 @@
...
+========================
+
+This module was written to extend the functionality of contracts to support to
+set an account journal on contract.
```
pedrobaeza: This module allows to set a sales account journal for creating the invoices on it.

contract_journal/models/contract.py
```diff
@@ -0,0 +1,30 @@
...
+            ('company_id', '=', company_id)]
+        return self.env['account.journal'].search(domain, limit=1)
+
+    journal_id = fields.Many2one(
+        'account.journal',
+        string='Journal',
+        default=_default_journal,
+        domain="[('type', '=', 'sale'),('company_id', '=', company_id)]")
+
+    def _prepare_invoice_data(self, cr, uid, contract, context=None):
```
pedrobaeza: Why don't put `@api.model` and you avoid old API?

contract_journal/models/contract.py
```diff
@@ -0,0 +1,30 @@
...
+        context = context or {}
+        invoice = super(AccountAnalyticAccount, self)._prepare_invoice_data(
```
pedrobaeza: s/invoice/invoice_vals (to highlight that it's a dictionary with vals, not a recordset)

contract_journal/models/contract.py
```diff
@@ -0,0 +1,30 @@
...
+            cr, uid, contract, context=context)
+        invoice.update({
```
pedrobaeza:
```
if contract.journal_id:
    invoice_vals['journal_id'] = contract.journal_id.id
```

contract_journal/models/contract.py
```diff
@@ -0,0 +1,30 @@
+# -*- coding: utf-8 -*-
+from openerp import models, fields, api
+
+
+class AccountAnalyticAccount(models.Model):
+    _inherit = 'account.analytic.account'
+
+    @api.model
+    def _default_journal(self):
+        company_id = self._context.get(
```
pedrobaeza: Use `self.env.context`, not private variable.

angelmoya:
Ok @pedrobaeza , work done

pedrobaeza:
Thanks! Maybe an improvement in tests is to create another journal, because sales_journal is the default one, so there's no possibility to check if the journal is set because of default or the module code.

Anyway, :+1:

angelmoya:
Thanks @pedrobaeza , test are doing with sales_journal and with other custom journal that is created in the test.

pedrobaeza:
Oh, sorry, I didn't see it. Perfect then!

dreispt:
:+1: +1:

PR #15, opened 2015-10-01
ADD contract_show_invoice module

angelmoya:
Add button in contract form to show all related invoices

### DIFF
--- a/contract_show_invoice/README.rst
+++ b/contract_show_invoice/README.rst
@@ -1,0 +1,58 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+=====================
+Contract Show Invoice
+=====================
+
+This module add a button in contracts to show their invoices.
+All invoices including lines for the Contract and other related invoices will be shown.
+
+Usage
+=====
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/110/8.0
+
+For further information, please visit:
+
+* https://www.odoo.com/forum/help-1
+
+Known issues / Roadmap
+======================
+
+* Add field with the total amount invoiced to the invoices button.
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues <https://github.com/OCA/contract/issues>`_.
+In case of trouble, please check there if your issue has already been reported.
+If you spotted it first, help us smashing it by providing a detailed and welcomed feedback
+`here <https://github.com/OCA/contract/issues/new?body=module:%20contract_show_invoice%0Aversion:%208.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
+
+Credits
+=======
+
+Contributors
+------------
+
+* Ángel Moya <angel.moya@domatix.com>
+
+
+Maintainer
+----------
+
+.. image:: http://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: http://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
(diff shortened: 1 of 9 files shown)


pedrobaeza:
Please call it _contract_show_invoice_ (without plural for consistency)

contract_invoices_button/models/invoice.py
```diff
@@ -0,0 +1,22 @@
...
+
+    @api.one
+    @api.depends('invoice_line.account_analytic_id')
+    def _analytic_account_ids(self):
+        self.analytic_account_ids = \
+            [x.account_analytic_id.id for x in self.invoice_line]
+
+    analytic_account_ids = fields.Many2many(
+        comodel_name='account.analytic.account',
+        relation='account_invoice_contract_rel',
```
pedrobaeza: Don't put relation, column1 and column2. In new API is not needed.

contract_invoices_button/models/invoice.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+class AccountInvoice(models.Model):
+    _inherit = 'account.invoice'
+
+    @api.one
+    @api.depends('invoice_line.account_analytic_id')
+    def _analytic_account_ids(self):
+        self.analytic_account_ids = \
+            [x.account_analytic_id.id for x in self.invoice_line]
```
pedrobaeza: This will also include manual created invoices, so it's not the best approach.

pedrobaeza:
I recommend you to change the approach to one that fills a many2one field to account.analytic.account in invoices where they are generated, and make the one2many on the contract. This requires a post-init-hook, to fill this many2one using the heuristic rule that you are using in the current compute method.

angelmoya:
@pedrobaeza I want to show all invoices related to contract, not only recurring invoices, also if one invoice are related with two invoices I want to show invoice in both contracts.

Maybe I could do other module with only related invoices, but is not scope of this module.

contract_invoices_button/models/invoice.py
pedrobaeza: Use then: `self.mapped('invoice_line.account_analytic_id').ids`

pedrobaeza:
OK then. I understood the module the other way.

pedrobaeza:
Check Travis and runbot

angelmoya:
I don't know about this error. It's about storing many2many computed field, in my server get this error, but after restart this error is not given.

I don't nkow what can I do.

Maybe set store false and create search function ??

angelmoya:
@pedrobaeza you could check #16 , this PR has a module contract_show_recurring_invoices with the other approach

pedrobaeza:
That's because your computed store many2many field, that it's trying to create new analytic lines. Try with the following code:

```
self.analytic_account_ids = self.mapped('invoice_line.account_analytic_id')
```

angelmoya:
OK, fixed

pedrobaeza:
:+1:

nhomar:
:+1:

contract_show_invoice/README.rst
```diff
@@ -0,0 +1,52 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+=====================
+Contract Show Invoice
+=====================
+
+This module add a button in contracts to show their invoices.
+
```
dreispt, line 11: Maybe we could be more precise. If I understood correctly:

```
All invoices including lines for the Contract will be shown.
```

contract_show_invoice/views/contract_view.xml
```diff
@@ -0,0 +1,31 @@
...
+            <field name="name">account.analytic.account.button.invoice
+            </field>
+            <field name="model">account.analytic.account</field>
+            <field name="inherit_id" ref="analytic.view_account_analytic_account_form" />
+            <field name="arch" type="xml">
+                <xpath expr='//div[@name="buttons"]' position='inside'>
+                    <button class="oe_stat_button" type="action" icon="fa-edit"
+                        name="%(contract_show_invoice.act_analytic_invoices)d"
+                        string="Invoices" help="Invoices related with this contract">
+                    </button>
```
dreispt, line 25: It would be nice for the smart button to include a computed field with the total amount invoiced.
If it's not implemented now, it's a idea to put in the roadmap section.

dreispt:
A couple of suggestions, but :+1:

angelmoya:
OK, updated README.rst

PR #16, opened 2015-10-02
ADD contract_show_recurring_invoice module

angelmoya:


contract_show_recurring_invoice/README.rst
```diff
@@ -0,0 +1,52 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+===============================
+Contract Show Recurring Invoice
+===============================
+
+This module add a button in contracts to show their recurring invoices.
```
pedrobaeza: s/add/adds

contract_show_recurring_invoice/__init__.py
```diff
@@ -0,0 +1,11 @@
+# -*- coding: utf-8 -*-
+from . import models
+
+
+def copy_recurring_invoice(cr, registry):
+    """Copy recurring invoice on contract."""
+    cr.execute("UPDATE account_invoice "
```
pedrobaeza: Use a compose UPDATE:
```
UPDATE account_invoice
SET contract_id=aaa.id
FROM account_analytic_account aaa
WHERE aaa.code = account_invoice.origin
```
More efficient and you avoid undetermined cases when there's no match.

contract_show_recurring_invoice/views/contract_view.xml
```diff
@@ -0,0 +1,31 @@
...
+        </record>
+
+        <record id="account_analytic_account_button_recurring_invoice" model="ir.ui.view">
+            <field name="name">account.analytic.account.button.invoice
+            </field>
+            <field name="model">account.analytic.account</field>
+            <field name="inherit_id" ref="analytic.view_account_analytic_account_form" />
+            <field name="arch" type="xml">
+                <xpath expr='//div[@name="buttons"]' position='inside'>
+                    <button class="oe_stat_button" type="action" icon="fa-edit"
```
pedrobaeza: Maybe you should put also (or only) a oe_link button next to "create invoice" button to have it on the same part of the form.

angelmoya: But "create invoice" is only visible for technical users

pedrobaeza: I wouldn't put this as technical. It's very useful and used everyday. Put it with the same permission as the button view invoice in sales orders.

angelmoya: I don't put "create invoice" button as technical, it's default functionality. So I don't put view invoices in same place.

Maybe could be usefull module to set this button to other groups.

pedrobaeza: OK, I understood you bad. I tought you say your button is technical, not the other. Anyway, I would put it next to it, but always visible. If technical features is enabled, you'll see both buttons. If not, you'll only see "Show invoices" button

angelmoya: Ok, I will change

angelmoya:
Ok, button changed

pedrobaeza:
Some comments yet to be changed

angelmoya:
Ok @pedrobaeza sorry for the noise, just updated

contract_show_recurring_invoice/__init__.py
```diff
@@ -0,0 +1,10 @@
...
+               "SET contract_id=aaa.id "
+               "FROM account_analytic_account aaa "
+               "WHERE aaa.code = account_invoice.origin")
```
pedrobaeza: Put a line feed at the end

dreispt: We want to update all Contract Invoices every time a worker (re)starts?

dreispt: Add a filter to only update invoices that need it, not all.

contract_show_recurring_invoice/views/contract_view.xml
```diff
@@ -0,0 +1,33 @@
...
+    <data>
+
+        <record id="act_recurring_invoices" model="ir.actions.act_window">
+            <field name="context">{'search_default_contract_id':
+                [active_id],
+                'default_contract_id': active_id}
+            </field>
+            <field name="name">Invoices</field>
+            <field name="res_model">account.invoice</field>
+            <field name="domain">[('contract_id','=',active_id)]</field>
```
dreispt: I think the filter is defined twice: the domain "hard" filter and the context "search_default" soft filter.
Maybe you can drop the domain filter.

contract_show_recurring_invoice/views/contract_view.xml
```diff
@@ -0,0 +1,33 @@
...
+        </record>
+
+        <record id="account_analytic_account_button_recurring_invoice"
+            model="ir.ui.view">
+            <field name="name">account.analytic.account.button.invoice
+            </field>
+            <field name="model">account.analytic.account</field>
+            <field name="inherit_id" ref="account_analytic_analysis.account_analytic_account_form_form" />
+            <field name="arch" type="xml">
+                <button name="recurring_create_invoice" position='after'>
```
dreispt: `"after"` instead of `'after'`

contract_show_recurring_invoice/views/contract_view.xml
```diff
@@ -0,0 +1,33 @@
...
+                    <button attrs="{'invisible': [('recurring_invoices','!=',True)]}"
+                        name="%(contract_show_recurring_invoice.act_recurring_invoices)d"
+                        string="⇒ show invoices" class="oe_link" type="action"/>
```
dreispt, line 26: It would be nice for this to be a smart button.
If you're not into it, add that to the Roadmap, so maybe someone else can pick it up and implement that.

contract_show_recurring_invoice/__init__.py
angelmoya: This sql is executed in post install, so all invoices with origin equals account analytic code must be updated, because contract_id is a new field.

contract_show_recurring_invoice/views/contract_view.xml
angelmoya, line 26: Ok, added to roadmap

contract_show_recurring_invoice/views/contract_view.xml
angelmoya: Ok, done

contract_show_recurring_invoice/__init__.py
dreispt: Yes, you're right.

contract_show_recurring_invoice/views/contract_view.xml
angelmoya: Ok, done

dreispt:
LGTM :+1:

pedrobaeza:
:+1:

PR #20, opened 2015-10-16
[8.0][ADD] contract_payment_mode module

angelmoya:


contract_payment_mode/__openerp__.py
```diff
@@ -0,0 +1,40 @@
...
+        'account_analytic_analysis',
+        'account_payment_partner'],
+    'category': 'Sales Management',
+    'license': 'AGPL-3',
+    'data': [
+        'views/contract_view.xml',
+    ],
+    'test': ['test/contract_payment_mode.yml'],
+    'installable': True,
+    'auto_install': False,
```
pedrobaeza: You can put auto_install=True, because this a nice-to-have in case you installed the other 2.

contract_payment_mode/models/contract.py
```diff
@@ -0,0 +1,34 @@
...
+class AccountAnalyticAccount(models.Model):
+    _inherit = 'account.analytic.account'
+
+    payment_mode_id = fields.Many2one(
+        'payment.mode',
+        string='Payment Mode',
+        domain="[('type', '=', 'sale')]")
+
+    @api.multi
+    def on_change_partner_id(self, partner_id, name, context=None):
```
pedrobaeza: With @api.multi, context is not passed as argument.

angelmoya:
Good remark @pedrobaeza , work done and ready for merge

pedrobaeza:
:+1: if Travis goes green

contract_payment_mode/models/contract.py
```diff
@@ -0,0 +1,33 @@
...
+    _inherit = 'account.analytic.account'
+
+    payment_mode_id = fields.Many2one(
+        'payment.mode',
+        string='Payment Mode',
+        domain="[('type', '=', 'sale')]")
+
+    @api.multi
+    def on_change_partner_id(self, partner_id, name):
+        res = super(AccountAnalyticAccount, self).on_change_partner_id(
```
dreispt, line 15: add `@api.ensure_one()` ?

dreispt:
One suggestion, but it's :+1:

contract_payment_mode/models/contract.py
angelmoya, line 15: Sure I can add ensure_one ? this method is called by old api method , from the view , and I think when it was creating new records, ids is an empty array

dreispt, line 15: I wrote wrong: it's `self.ensure_one()` instead.
It's a safeguard since you assume in the rest of the code that you're dealing with a singleton.
But maybe it redundant for on_change. What do you think @pedrobaeza ?

angelmoya, line 15: In this method self is not called, I only use parner_id parameter, I think is right, but @pedrobaeza can make the point clear.

dreispt, line 15: You're right. Forget it then.

pedrobaeza, line 15: For the records, old API on_change (triggered by UI), although decorated with @api.multi, should be multi-record safe, because you mustn't use self (or make a browse in old API) to get current values, that are only loaded on client-side. That's why you pass extra arguments with the values you need to work with (and the problem that the module _web_context_tunnel_ solves for adding more arguments).

PR #22, opened 2016-01-27
[8.0][IMP][contract_payment_mode] Copy partner payment mode to contracts when installing

antespi:
This improvement copies payment mode (configured in customer) to contracts when installing the addon.

### DIFF
--- a/contract_payment_mode/__init__.py
+++ b/contract_payment_mode/__init__.py
@@ -1,2 +1,6 @@
 # -*- coding: utf-8 -*-
+# © 2016 Antiun Ingenieria S.L. - Antonio Espinosa
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
 from . import models
+from .hooks import post_init_hook
--- a/contract_payment_mode/hooks.py
+++ b/contract_payment_mode/hooks.py
@@ -13,7 +13,10 @@ def post_init_hook(cr, registry):
     with api.Environment.manage():
         env = api.Environment(cr, SUPERUSER_ID, {})
         m_contract = env['account.analytic.account']
-        contracts = m_contract.search([('type', '=', 'contract')])
+        contracts = m_contract.search([
+            ('type', '=', 'contract'),
+            ('payment_mode_id', '=', False),
+        ])
         if contracts:
             _logger.info('Setting payment mode: %d contracts' %
                          len(contracts))
--- a/contract_payment_mode/hooks.py
+++ b/contract_payment_mode/hooks.py
@@ -1,0 +1,24 @@
+# -*- coding: utf-8 -*-
+# © 2016 Antiun Ingenieria S.L. - Antonio Espinosa
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openerp import api, SUPERUSER_ID
+import logging
+
+_logger = logging.getLogger(__name__)
+
+
+def post_init_hook(cr, registry):
+    """Copy payment mode from partner to the new field at contract."""
+    with api.Environment.manage():
+        env = api.Environment(cr, SUPERUSER_ID, {})
+        m_contract = env['account.analytic.account']
+        contracts = m_contract.search([('type', '=', 'contract')])
+        if contracts:
+            _logger.info('Setting payment mode: %d contracts' %
+                         len(contracts))
+        for contract in contracts:
+            payment_mode = contract.partner_id.customer_payment_mode
+            if payment_mode:
+                contract.payment_mode_id = payment_mode.id
+        _logger.info('Setting payment mode: Done')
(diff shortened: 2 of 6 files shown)


contract_payment_mode/__init__.py
```diff
@@ -1,2 +1,27 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
 from . import models
+from openerp import api, SUPERUSER_ID
+import logging
+
+_logger = logging.getLogger(__name__)
+
+
+def copy_payment_mode(cr, registry):
```
pedrobaeza: Please put this into a hooks.py file according conventions: https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#installation-hooks

contract_payment_mode/__init__.py
```diff
@@ -1,2 +1,27 @@
...
+        if contracts:
+            _logger.info('Setting payment mode: %d contracts' %
+                         len(contracts))
+        for contract in contracts:
+            payment_mode = contract.partner_id.customer_payment_mode
+            if payment_mode:
+                contract.write({
+                    'payment_mode_id': payment_mode.id,
+                })
+        _logger.info('Setting payment mode: Done')
```
pedrobaeza: This is always log independently if there are contracts or not

contract_payment_mode/__init__.py
antespi: OK, this is new for me. I've copy-pasted from other OCA addon ... Good to know, thanks

contract_payment_mode/__init__.py
```diff
@@ -1,2 +1,27 @@
...
+        env = api.Environment(cr, SUPERUSER_ID, {})
+        m_contract = env['account.analytic.account']
+        contracts = m_contract.search([('type', '=', 'contract')])
+        if contracts:
+            _logger.info('Setting payment mode: %d contracts' %
+                         len(contracts))
+        for contract in contracts:
+            payment_mode = contract.partner_id.customer_payment_mode
+            if payment_mode:
+                contract.write({
```
pedrobaeza: You can write the compact form: `contract.payment_mode_id = payment_mode.id`

antespi:
@pedrobaeza fixes done, thanks.

contract_payment_mode/__init__.py
antespi: That's the point :smile:

pedrobaeza: But the other is not logged always.

sergio-teruel:
I tested in runbot and :+1:

contract_payment_mode/hooks.py
```diff
@@ -0,0 +1,24 @@
...
+    with api.Environment.manage():
+        env = api.Environment(cr, SUPERUSER_ID, {})
+        m_contract = env['account.analytic.account']
+        contracts = m_contract.search([('type', '=', 'contract')])
+        if contracts:
+            _logger.info('Setting payment mode: %d contracts' %
+                         len(contracts))
+        for contract in contracts:
+            payment_mode = contract.partner_id.customer_payment_mode
+            if payment_mode:
```
yajo, line 25: Shouldn't this be `if payment_mode and not contract.payment_mode_id:`?

pedrobaeza, line 25: It can't be, because this module adds the field payment_mode_id.

yajo, line 25: Oops

yajo:
:+1:

pedrobaeza:
Coveralls is dropping a lot because of the UNIT TEST in Travis: https://github.com/OCA/contract/blob/8.0/.travis.yml#L35. I suggest to remove it. Another option is to test the method itself (which by other way, is not a bad idea), that is create a payment mode, a partner with the payment mode, a contract without payment mode, call that method, and test if the contract now have payment mode.

yajo:
Good idea!

rafaelbn:
After https://github.com/OCA/contract/pull/22#issuecomment-176638727 for me is :+1:

pedrobaeza:
Any news on this?

pedrobaeza:
@antespi, can you try to make the test and we merge this?

antespi:
Test added and coveralls is green :smile:

contract_payment_mode/tests/test_contract_payment_init.py
```diff
@@ -0,0 +1,40 @@
...
+            'name': 'Test contract',
+            'type': 'contract',
+            'partner_id': self.partner.id,
+            'payment_mode_id': self.payment_mode.id,
+        })
+        self.assertEqual(self._contract_payment_mode_id(contract.id),
+                         self.payment_mode.id)
+
+        contract.payment_mode_id = False
+        self.assertEqual(self._contract_payment_mode_id(contract.id), False)
```
pedrobaeza, line 36: You can do here `self.assertFalse()` instead, but it's not so important

pedrobaeza:
:+1: 

Thanks. Merging.

PR #24, opened 2016-02-22
[ADD] contract_recurring_invoicing_marker

pedrobaeza:
# 
# Markers for contract recurring invoices

This module allows to include some markers on the lines of your recurring
invoices definition inside the contract so that the generated invoice lines
will contain a dynamic text depending on these markers.

These markers are the supported ones:
- #START#: Start date of the invoiced period.
- #END# End date of the invoiced period.
# Usage

On a contract (_Sales > Sales > Contracts_), mark "Generate recurring invoices
automatically" for enabling the creation of recurring invoices.

In the "Invoice Lines" section, you can add lines with products. In the
_Description_ field, you can now add any of the markers mentioned before
between the rest of the text.

When you invoice this contract (automatically or manually), your invoice
will contain the corresponding text that replaces the marker.
# Known issues / Roadmap
- Add more markers, like #START_MONTH# or #END_MONTH#.


contract_recurring_invoicing_marker/__openerp__.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
+# (c) 2016 Serv. Tecnol. Avanzados - Pedro M. Baeza
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+{
+    'name': 'Markers for contract recurring invoices',
+    'version': '8.0.1.0.0',
+    'category': 'Contract Management',
+    'author': 'Serv. Tecnol. Avanzados - Pedro M. Baeza, '
+              'Odoo Community Association (OCA)',
```
sergio-incaser: Add license key

sergio-teruel:
Code review done, while to runbot, but :+1:

pedrobaeza:
@sergio-incaser, done

rafaelbn:
I cannot test it, runbot fails :cry: 

@gurneyalex could you help us with runbot please?

contract_recurring_invoicing_marker/models/account_analytic_account.py
```diff
@@ -0,0 +1,41 @@
...
+        next_date = fields.Date.from_string(
+            contract.recurring_next_date or fields.Date.today())
+        interval = contract.recurring_interval
+        old_date = next_date
+        if contract.recurring_rule_type == 'daily':
+            new_date = next_date + relativedelta(days=interval - 1)
+        elif contract.recurring_rule_type == 'weekly':
+            new_date = next_date + relativedelta(weeks=interval, days=-1)
+        else:
+            new_date = next_date + relativedelta(months=interval, days=-1)
```
dreispt, line 23: Nothing wrong with this if block, but using a dict mapping the relativedeltas would be nicer.

contract_recurring_invoicing_marker/models/account_analytic_account.py
```diff
@@ -0,0 +1,41 @@
...
+        if 'old_date' in self.env.context and 'next_date' in self.env.context:
+            lang_obj = self.env['res.lang']
+            contract = line.analytic_account_id
+            lang = lang_obj.search(
+                [('code', '=', contract.partner_id.lang)])
+            date_format = lang.date_format or '%d/%M/%Y'
+            res['name'] = res['name'].replace(
+                '#START#', self.env.context['old_date'].strftime(date_format))
+            res['name'] = res['name'].replace(
+                '#END#', self.env.context['next_date'].strftime(date_format))
```
dreispt: Isolating the replace logic in a helper private method would make it easier to extend.

contract_recurring_invoicing_marker/views/account_analytic_account_view.xml
```diff
@@ -0,0 +1,20 @@
...
+<openerp>
+    <data>
+
+        <record id="account_analytic_account_form_form" model="ir.ui.view">
+            <field name="name">Contract form (with markers)</field>
+            <field name="model">account.analytic.account</field>
+            <field name="inherit_id" ref="account_analytic_analysis.account_analytic_account_form_form"/>
+            <field name="arch" type="xml">
+                <field name="recurring_invoice_line_ids" position="after">
+                    <group string="Legend (for the markers inside invoice lines description)">
```
dreispt: Add a `name` for extensibility.

dreispt:
Some improvement suggestions, but LGTM.

antespi:
Code review, I agree with @dreispt remarks. :+1: after fixing them

contract_recurring_invoicing_marker/models/account_analytic_account.py
pedrobaeza, line 23: It can't be done because the relativedelta contains a variable interval depending on the contract, and I think that nest the dictionary inside the method is uglier.

pedrobaeza:
@dreispt, thanks for your review. I have taken your suggestions into account except one that I have commented.

rafaelbn:
Rebuilding for testing

rafaelbn:
After 1 hour, still pending:

rafaelbn:
OK, runbot is not working, all the rebuilds are getting in queue in pending state
cc @gurneyalex

pedrobaeza:
Can you review again this module for merging it?

rafaelbn:
I have tested and I don't this this is working as expected. Check one contract, 3 invoices of diferent months, same description label.

pedrobaeza:
Several things:
- The start and end are not taken from the contract start and end dates, but the invoiced period.
- You have something strange on the date format of the language configuration, and thus the month is missing.

Please review again taking this into consideration.

rafaelbn:
- I tested in runbot: http://3142616-24-7df454.runbot1.odoo-community.org/web?db=3142616-24-7df454-base
- Date format is Odoo standard , I didn't change it.
- Check that I write #START# on top of '----------' and #END# bellow. In the invoice's description it's #END# on top and #START# bellow. This is one mistake. Second mistake it that today is March and  this means that it must write 03/01/2016, after '----------', after 03/31/2016

pedrobaeza:
The second statement is not correct. The date is taken from the "Date of next invoice", and you can see that the next invoice is already on may (because you have clicked several times on create invoice). I will check the other thing.

pedrobaeza:
@rafaelbn, the problem in the date format was because the contact hasn't any language, and the default was badly put. I have changed it.

But the other thing is only that you are not seeing the month, so please check again.

rafaelbn:
:+1: thanks!

pedrobaeza:
Merging then

Thanks to all reviewers.

PR #29, opened 2016-04-06
[FIX] contract_payment_mode: get bank account from partner

jabibi:
in recurring invoice


pedrobaeza:
What is the rationale over the change? You have also the assignation 2 times. Please correct.

Can you also sign the CLA?

jabibi:
partner_bank_id field should be the Bank Account from partner, not from the payment mode that normally is main company's bank account. You are right, the code is wrong, I will push again. I dont know how to do in the same PR so there will be an other one, sorry

I just send the Contributor License Agreement signed

pedrobaeza:
For making it in the same PR, you have to make `git commit --amend --all` and then `git push -f`

jabibi:
sorry, I open do the new PR. I'll remember that for next time

PR #31, opened 2016-04-13
[IMP] contract_show_invoice: more lightweight solution

sbidoul:
In order to implement the button, it's not necessary to have a heavy stored computed field. So I propose to remove it.


pedrobaeza:
Of course it looks good better! :+1: 

Can you propose the same for v9?

sbidoul:
@pedrobaeza I removed the unneeded context from the action too.

I'll port it to v9 when this one is merged.

sbidoul:
I also removed the test which was based on the unneeded computed field.

sbidoul:
Runbot is red because of the missing xslxwriter python dependency

jabibi:
Code review: it looks good for me. Clean and fresh according to new api :+1:

PR #35, opened 2016-09-09
[ADD] contract_variable_quantity:

pedrobaeza:
# 
# Variable quantity in contract recurrent invoicing

With this module, you will be able to define in recurring contracts some
lines with variable quantity according a provided formula.
# Configuration
- Go to Sales > Configuration > Contracts > Formulas (quantity).
- Define any formula based on Python code that stores at some moment a
  float/integer value of the quantity to invoice in the variable 'result'.
  
  You can use these variables to compute your formula:
  - _env_: Environment variable for getting other models.
  - _context_: Current context dictionary.
  - _user_: Current user.
  - _line_: Contract recurring invoice line that triggers this formula.
  - _contract_: Contract whose line belongs to.
  - _invoice_: Invoice (header) being created.
# Usage

To use this module, you need to:
- Go to Sales -> Contracts and select or create a new contract.
- Check _Generate recurring invoices automatically_.
- Add a new recurring invoicing line.
- Select "Variable quantity" in column "Qty. type".
- Select one of the possible formulas to use (previously created).


pedrobaeza:
cc @Tecnativa

sergio-teruel:
Tested in runbot... Great!! :+1:

contract_variable_quantity/models/contract.py
```diff
@@ -0,0 +1,66 @@
...
+        ], required=True, default='fixed', string="Qty. type")
+    qty_formula_id = fields.Many2one(
+        comodel_name="contract.line.qty.formula", string="Qty. formula")
+
+
+class ContractLineFormula(models.Model):
+    _name = 'contract.line.qty.formula'
+
+    name = fields.Char(required=True)
+    code = fields.Text(required=True, default="result = 0")
```
dreispt, line 47: We should have here the usage explanation (from the README), in the `help` tooltip.

contract_variable_quantity/models/contract.py
```diff
@@ -0,0 +1,66 @@
...
+            'env': self.env,
+            'context': self.env.context,
+            'user': self.env.user,
+            'line': self.env['account.analytic.invoice.line'],
+            'contract': self.env['account.analytic.account'],
+            'invoice': self.env['account.invoice'],
+        }
+        try:
+            safe_eval(
+                self.code.strip(), eval_context, mode="exec", nocopy=True)
```
dreispt, line 61: I'm curious: why didn't you just use something like `vals['quantity'] = safe_eval(...) or 0`?
It probably would be simpler/more intuitive for end users.

dreispt:
LGTM :+1:

contract_variable_quantity/models/contract.py
pedrobaeza: You already have it in the screen explicitly. I'll add an screenshot in the README to show it.

contract_variable_quantity/models/contract.py
pedrobaeza: Because you can have more than one line in that case (expression vs block code). It's more comfortable (and sometimes, the only possibility, although you know you can make a lot with one-liners) to have the possibility to prepare some data through several lines and then assign the result you want.

contract_variable_quantity/models/contract.py
dreispt: OK, no need to bother with the screenshot.

contract_variable_quantity/models/contract.py
dreispt: OK

contract_variable_quantity/views/contract_view.xml
```diff
@@ -0,0 +1,90 @@
...
+                    <div class="oe_title">
+                        <h1>
+                            <field name="name" placeholder="Name"/>
+                        </h1>
+                    </div>
+                    <group string="Code">
+                        <div style="margin-top: 4px;">
+                            <field name="code" nolabel="1"/>
+                            <h3>Help with Python expressions.</h3>
+                            <p>You have to insert valid Python code block that stores at some moment a float/integer value of the quantity to invoice in the variable 'result'.</p>
```
dreispt, line 45: Here are the instructions I missed!

contract_variable_quantity/models/contract.py
pedrobaeza: Added anyway! :wink:

rafaelbn:
Tested in runbot :+1:

PR #37, opened 2016-09-15
ADD contract_account_banking_mandate module

elacunza:
To define the banking mandate for recurring invoices

PR #38, opened 2016-09-15
ADD 8.0 contract account banking mandate v2

elacunza:
Follow up to #37, incorrectly asked to merge to 9.0 branch. This module is for 8.0 branch as it is now.


pedrobaeza (approved):

dreispt:
This just adds the field.
I was expecting to see also some logic to associate the mandate with the invoice's payment form.
At least the invoice report should also be changed to include that mandate information.

elacunza:
@dreispt This adds the field and fills in the field in the recurring invoice when it is generated.
The invoice report should be changed in account_banking_mandate, because that is the module that adds the mandate field to invoices.

dreispt:
OK :+1:

susport (approved):
👍

carlosdauden (approved):

carlosdauden:
:+1:

PR #40, opened 2016-11-12
10.0 contract

StefanRijnhart:


pedrobaeza:
Travis is failing saying that `decimal_precision` is not imported. Maybe dependency chain has changed? Can you check it?

StefanRijnhart:
No worries @pedrobaeza. That is why it's work in progress...

StefanRijnhart:
Now ready for review. Because the 10.0 branch had not received any changes I simply merged in the latest changes of 9.0. I hope that is alright.
Only substantial change is that the sale groups are now no longer available anymore without the sale module, so I transferred the permissions to the accounting groups. Right now, the accounting user groups can only read the analytic invoice line configuration, but maybe they should have all permissions instead?

contract/README.rst
```diff
@@ -21,16 +22,27 @@ To use this module, you need to:
 
 #. Go to Sales -> Contracts and select or create a new contract.
```
susport: Contracts menu is not in Sales, it is in Invoicing menu.

susport (commented):
Validate that Date of Next Invoice is greater than the current date.

contract/README.rst
StefanRijnhart: Thanks, I have realigned this description with reality.

StefanRijnhart:
@susport I think being able to invoice retroactively could be useful in some cases so I would rather not implement the check you suggested.

contract/__manifest__.py
```diff
@@ -5,19 +5,19 @@
 
 {
     'name': 'Contracts Management recurring',
-    'version': '9.0.1.0.0',
+    'version': '10.0.1.1.0',
```
pedrobaeza: 10.0.1.0.0, as the first 10.0 version

contract/models/contract.py
```diff
@@ -145,11 +155,34 @@ def _onchange_recurring_invoices(self):
             self.recurring_next_date = self.date_start
 
     @api.model
+    def get_relalive_delta(self, recurring_rule_type, interval):
```
pedrobaeza: s/relalive/relative

contract/models/contract.py
```diff
@@ -181,68 +214,62 @@ def _prepare_invoice_line(self, line, invoice_id):
...
+            'origin': self.name,
+            'company_id': self.company_id.id,
+            'contract_id': self.id,
+            'user_id': self.partner_id.user_id.id,
         })
         # Get other invoice values from partner onchange
         invoice._onchange_partner_id()
-        return invoice._convert_to_write(invoice._cache)
+        return invoice._convert_to_write(
+            {name: invoice[name] for name in invoice._cache})
```
pedrobaeza: Why changing this?

StefanRijnhart: Because in the case of empty values for many2one records, the cache contains an empty tuple instead of an empty recordset. The invoice record contains the correct value.

Before I added this change, this line broke the test. I took the modification from the Odoo core addons where it is used in several places.

pedrobaeza: Uhm, that's something ugly that doesn't happen on v9. Another hack to take into account...

contract/models/contract.py
StefanRijnhart: Done

contract/__manifest__.py
StefanRijnhart: Done

pedrobaeza (approved):

lasley (approved):
LGTM thanks @StefanRijnhart

pedrobaeza:
@StefanRijnhart, can you please squash last 4 commits into one before proceeding to merge?

StefanRijnhart:
Thanks for the reviews. Discovered a bug though.

@pedrobaeza about the squashing, that would remove credits so I'd rather not.

StefanRijnhart:
@pedrobaeza Oh, the last four commits. Will do!

pedrobaeza:
No, what I want you to do is to squash your migration commits together.

pedrobaeza:
We have commented at the same time :wink:

lasley (changes requested):
`contract_variable_quantity` is included here, but not installable.

StefanRijnhart:
Ready for merging, rebased and all. The bug I experienced was related to another call to convert_to_write. However, the problem with this method was fixed upstream in https://github.com/odoo/odoo/commit/51680426, so I could revert my changes to this call.

pedrobaeza:
I'm relief to see that we don't need to introduce a new hack.

pedrobaeza:
Merging this PR. @lasley, please rebase for your PR about variable_quantity.

rafaelbn:
I see some changes in the menus, are they intentionally?

StefanRijnhart:
@rafaelbn Yes, sale menus are no longer available in Odoo 10.0 without the additional dependency on the sale module so I moved the menus to the accounting menu

PR #41, opened 2016-11-13
ADD more markers #START_MONTH# and #END_MONTH#

gonxi:
ADD more markers, #START_MONTH# and #END_MONTH#

### DIFF
--- a/contract_recurring_invoicing_marker/__openerp__.py
+++ b/contract_recurring_invoicing_marker/__openerp__.py
@@ -6,9 +6,8 @@
     'name': 'Markers for contract recurring invoices',
     'version': '8.0.1.0.0',
     'category': 'Contract Management',
-    'author': 'Serv. Tecnol. Avanzados - Pedro M. Baeza, '
+    'author': 'Serv. Tecnol. Avanzados - Pedro M. Baeza, bisnesmart - Gonzalo Borrás,'
               'Odoo Community Association (OCA)',
-              'bisnesmart - Gonzalo Borras',
     'website': 'http://www.serviciosbaeza.com',
     'depends': [
         'account_analytic_analysis',
--- a/contract_recurring_invoicing_marker/__openerp__.py
+++ b/contract_recurring_invoicing_marker/__openerp__.py
@@ -8,6 +8,7 @@
     'category': 'Contract Management',
     'author': 'Serv. Tecnol. Avanzados - Pedro M. Baeza, '
               'Odoo Community Association (OCA)',
+              'bisnesmart - Gonzalo Borras',
     'website': 'http://www.serviciosbaeza.com',
     'depends': [
         'account_analytic_analysis',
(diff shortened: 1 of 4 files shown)


pedrobaeza (changes requested):
Remove 'text' word in your additions.

gonxi:
ok!

StefanRijnhart:
How about allowing for a template string passed to strftime instead of adding more and more custom substitutions?

lasley:
I second that idea. Playing off the current format, we could just embed the string into the markers. Maybe something like `#START:%B %Y#`, then just split by `:` & use the internal date format if included. Otherwise fallback to the date format arg in order to provide back compatibility.

ovnicraft:
@StefanRijnhart i start working with `contract` module so you recomendation will have locale issue ?  `datetime` is locale aware but IDK about `strftime`

lasley:
@ovnicraft - By locale, do you mean the ability to put months in as different languages?

contract_recurring_invoicing_marker/__openerp__.py
```diff
@@ -6,7 +6,7 @@
     'name': 'Markers for contract recurring invoices',
     'version': '8.0.1.0.0',
     'category': 'Contract Management',
-    'author': 'Serv. Tecnol. Avanzados - Pedro M. Baeza, '
+    'author': 'Serv. Tecnol. Avanzados - Pedro M. Baeza, bisnesmart - Gonzalo Borrás,'
```
elicoidal, line 9: usually we put companies in that section as the contributors are in the README

pedrobaeza: Yeah, the exception was me, that my name was my company. But this is no more, as we have now Tecnativa, so the correct is: `Tecnativa, bisnesmart,`

pedrobaeza: Anyway, I'm not sure if this contribution is enough to raise Gonzalo as co-author. @lasley mentioned an interesting rule: if the contribution is about 20-30% of new/refactored code, it deserves to be here. If not, then it should go to contributors section on README.

lasley: Keep in mind that the author key is a limited field, and I'm pretty sure it's not even the full 255 chars of the field.

I'll submit a draft change to our contribution guidelines today on this. I've been testing the rule internally, but quite a few heavy contributors seemed to like it when they saw it. Figure it's time to make the suggestion official now that it's public 😄

gonxi: For me it is not a problem to appear here or in the readme. I only added my name by convention

StefanRijnhart:
@ovnicraft Good point. Maybe use Babel's format_date instead. http://babel.pocoo.org/en/latest/dates.html

rafaelbn:
Hi @gonxi , please could you review travis in order to get it green? 😄

pedrobaeza:
Any news about this. If requested changes are not performed, I will proceed to close it for cleaning this.

pedrobaeza:
Closing as no further answer.

PR #42, opened 2016-11-30
[IMP][10.0] contract: Add templating

lasley:
This PR adds template functionality to `contract`.

Depends:
- [x] #40


rafaelbn:
Hi @lasley , contracts had templates in v8, have you recover that functionality?

lasley:
It seems I did, yeah. Had no idea it was a pre-existing feature 😆

rafaelbn:
I ask, because we have customers in v8 using them and we will need it fits for the migration. cc @pedrobaeza

lasley:
An overview of my changes:
* Move most field declarations & onchanges from `account.analytic.account` to new model `account.analytic.contract`
* _inherit `account.analytic.contract` from `account.analytic.account`
* Add onchange on new field `contract_template_id` in `account.analytic.account` to set fields from selected `account.analytic.contract` 

Works quite nicely and didn't take much code, I was pretty happy. Hopefully it fits migration 😉  - I can backport to 9 fairly easy.

contract/models/account_anayltic_account.py
```diff
@@ -0,0 +1,189 @@
+# -*- coding: utf-8 -*-
```
lasley: File name typo

lasley:
Alright this is ready for review

contract/models/account_analytic_contract.py
```diff
@@ -0,0 +1,61 @@
...
+    @api.model
+    def _default_journal(self):
+        company_id = self.env.context.get(
+            'company_id', self.env.user.company_id.id)
+        domain = [
+            ('type', '=', 'sale'),
+            ('company_id', '=', company_id)]
+        return self.env['account.journal'].search(domain, limit=1)
+
+    @api.onchange('partner_id')
```
tedsalmon: `partner_id` is not defined as a field of this model.

contract/models/account_analytic_contract.py
```diff
@@ -0,0 +1,61 @@
+# -*- coding: utf-8 -*-
```
tedsalmon: No `name` field or `name_get` is defined, resulting in oddly named entries.

tedsalmon (changes requested):
A couple changes requested :)

contract/models/account_anayltic_account.py
```diff
@@ -3,151 +3,47 @@
 # © 2014 Angel Moya <angel.moya@domatix.com>
```
tedsalmon: This file is misspelled. `s/account_anayltic_account.py/account_analytic_account.py`

lasley:
Does anyone have any thoughts on moving contracts into a subheader of its own? Right now it's under Invoicing => Sales => Contracts, but I'd like to maybe add an Invoicing => Contracts with the following:
* Contracts - Same thing we have now
* Invoices - Invoice view filtered to show only those generated from contracts
* Templates - View for the contract templates that I added here (there's not a view right now)

pedrobaeza:
I prefer to not create another "Invoices" entry. Better to have a filter like "From contracts) in the existing one.

You need to have Contracts on Sales, because salesmen are the one who manage them. About templates, I think it's better to have it under Sales > Configuration > Contracts > Templates.

contract/__manifest__.py
```diff
@@ -10,6 +10,7 @@
     'license': 'AGPL-3',
     'author': "OpenERP SA,"
```
lasley: Should this `OpenERP SA` be `Odoo SA`?

pedrobaeza: I'm not sure if they have switched already their legal part. They don't put author now in Odoo 10 manifests, so no clue...

lasley: Hmmm best to play it on the safe and leave it as is then. Thanks @pedrobaeza

lasley:
@tedsalmon - thanks for the review

@pedrobaeza - I somewhat followed your suggestion on the menu placement. `contract` doesn't have any `sale` depenencies though, so I didn't want to bring that in just for the menu. Instead I just placed the "Contract Templates" in the Invoicing Configuration instead, which I think hits the intent well enough. 

I'm torn on whether it should be above or below "Settings". I settled on below, but am definitely open to other placements.

contract/views/account_analytic_account_view.xml
```diff
@@ -1,22 +1,11 @@
...
-        <field name="view_id" ref="account.invoice_tree" />
-        <field name="search_view_id" ref="account.view_account_invoice_filter"/>
-    </record>
-
     <record id="account_analytic_account_recurring_form_form" model="ir.ui.view">
         <field name="name">account.analytic.account.invoice.recurring.form.inherit</field>
         <field name="model">account.analytic.account</field>
         <field name="inherit_id" ref="analytic.view_account_analytic_account_form"/>
-        <field eval="40" name="priority"/>
+        <field name="mode">primary</field>
```
lasley, line 8: Note- setting the mode to primary seemed more appropriate here. In the previous case this view was being appended to all `account.analytic.account` views, which I don't believe was the intent. Please correct me if I'm wrong on that assumption.

lasley:
I noticed that Runbot was failing here, and it turns out to be failing on master too. Turns out there is a warning coming from `contract_variable_quantity` due to it adding a menu into a module it doesn't own.

I leveraged this error and moved the menu into the core `contract` module. This is the new home of the `Contract Template` menu now, so we just nailed two birds with one stone.

That explanation said, please excuse my changing two modules in 9d24dea

pedrobaeza:
About the menu moved out from Sales, I think we can re-add that menu in product_contract, as that module needs the sale one. What do you think?

lasley:
Oh I actually didn't mean to move the original Contract link from Invoicing => Sales. You made a perfectly valid point previously with the fact that sales reps are the ones working in contracts. Not sure where the menu went TBH; it's still in the code.

lasley:
Alright menu added back - I accidentally left the new menu as the other one's ID, so it was being overwritten. Ready for review!

tedsalmon (approved):
👍  LGTM Now :)

lasley:
Can a PSC please review?

pedrobaeza (approved):

PR #46, opened 2016-12-02
[MIG][9.0] contract_journal

lasley:
* Bump versions
* Update readme template
* Rename view and use odoo tags
* Remove deprecated yml
* Add tests
* Update invoice vals generate for v9 module

I'll handle v10 as well - that's really the version I need it on, just seemed like a dick move to skip a version in the migrations 😛


contract_journal/tests/test_contract_journal.py
```diff
@@ -0,0 +1,52 @@
...
+        """ It should inject the journal into invoice vals """
+        res = self.contract._prepare_invoice()
+        self.assertEqual(res['journal_id'], self.journal.id)
+
+    def test_default_journal_company(self):
+        """ It should return journal of correct company """
+        res = self.env['acount.analytic.account']._default_journal_id()
+        self.assertEqual(res.company_id, self.env.user.company_id)
+
+    def test_default_journal_type(self):
```
pedrobaeza, line 36: Join this test with previous one

pedrobaeza (changes requested):
Check Travis result

contract_journal/views/contract.xml
```diff
@@ -0,0 +1,40 @@
+<?xml version="1.0"?>
+<odoo>
+
+    <record id="account_analytic_account_journal_form" model="ir.ui.view">
+        <field name="name">account.analytic.account.journal.form</field>
+        <field name="model">account.analytic.account</field>
+        <field name="inherit_id" ref="analytic.view_account_analytic_account_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='template_id']" position="before">
```
pedrobaeza, line 9: The error is here

pedrobaeza (changes requested):

contract_journal/views/contract.xml
lasley: Admittedly I should be testing my 9.0 submissions better, but mannnnn switching over to the v9 instance from the v10 in my local is a real pain 😦 . I swear one of these days I'm going to create us some sort of `odooenv`

lasley:
I'm dumb though, I just found this in the new contract. I guess I've been looking at it the whole time; derp

PR #49, opened 2017-01-12
[ADD] product_contract: Create module

tedsalmon:
* Add contract functionality to `product.templates`
* Add logic to create contracts from `sale.order` that contains contract products.

Depends: 

~~- [x] #42~~

cc @lasley

### DIFF
--- a/product_contract/README.rst
+++ b/product_contract/README.rst
@@ -1,0 +1,66 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+================
+Product Contract
+================
+
+This module adds support for products to be linked to contract templates.
+It also adds functionality to automatically create a contract, from the template,
+when a ``sale.order`` contains a product that implements a contract.
+
+Usage
+=====
+
+To use this module, you need to:
+
+#. Go to Sales -> Products and select or create a product.
+#. Check "Is a contract" and select the contract template related to the
+   product
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/110/10.0
+
+Known issues / Roadmap
+======================
+
+* None
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/contract/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
+help us smash it by providing detailed and welcomed feedback.
+
+Credits
+=======
+
+Images
+------
+
+* Odoo Community Association: `Icon <https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.svg>`_.
+
+Contributors
+------------
+
+* Ted Salmon <tsalmon@laslabs.com>
+
+
+Maintainer
+----------
+
+.. image:: https://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit https://odoo-community.org.
--- a/product_contract/__manifest__.py
+++ b/product_contract/__manifest__.py
@@ -1,0 +1,23 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 LasLabs Inc.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    'name': 'Product Contract',
+    'version': '10.0.1.0.0',
+    'category': 'Contract Management',
+    'license': 'AGPL-3',
+    'author': "LasLabs, "
+              "Odoo Community Association (OCA)",
+    'website': 'https://laslabs.com',
+    'depends': [
+        'contract',
+        'product',
+        'sale',
+    ],
+    'data': [
+        'views/product_template_view.xml',
+    ],
+    'installable': True,
+    'application': False,
+}
--- a/product_contract/models/product_template.py
+++ b/product_contract/models/product_template.py
@@ -1,0 +1,23 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 LasLabs Inc.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, fields, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = 'product.template'
+
+    is_contract = fields.Boolean('Is a contract')
+    contract_template_id = fields.Many2one(
+        comodel_name='account.analytic.contract',
+        string='Contract Template',
+    )
+
+    @api.onchange('is_contract')
+    def _change_is_contract(self):
+        """ Clear the relation to contract_template_id when downgrading
+        product from contract
+        """
+        if not self.is_contract:
+            self.contract_template_id = False
--- a/product_contract/models/sale_order.py
+++ b/product_contract/models/sale_order.py
@@ -1,0 +1,27 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 LasLabs Inc.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, models
+
+
+class SaleOrder(models.Model):
+    _inherit = 'sale.order'
+
+    @api.multi
+    def action_confirm(self):
+        """ If we have a contract in the order, set it up """
+        for rec in self:
+            order_lines = self.mapped('order_line').filtered(
+                lambda r: r.product_id.is_contract
+            )
+            for line in order_lines:
+                contract_tmpl = line.product_id.contract_template_id
+                contract = self.env['account.analytic.account'].create({
+                    'name': '%s Contract' % rec.name,
+                    'partner_id': rec.partner_id.id,
+                    'contract_template_id': contract_tmpl.id,
+                })
+                line.contract_id = contract.id
+                contract.recurring_create_invoice()
+        return super(SaleOrder, self).action_confirm()
--- a/product_contract/models/sale_order_line.py
+++ b/product_contract/models/sale_order_line.py
@@ -1,0 +1,14 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 LasLabs Inc.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class SaleOrderLine(models.Model):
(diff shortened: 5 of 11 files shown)


product_contract/models/product_template.py
```diff
@@ -0,0 +1,22 @@
...
+        comodel_name='account.analytic.contract',
+        string='Contract Template',
+    )
+
+    @api.onchange('is_contract')
+    def _change_is_contract(self):
+        """ Clear the relation to contract_template_id when downgrading
+        product from contract """
+        if not self.is_contract:
+            self.contract_template_id = None
```
pedrobaeza: Set it to `False`, instead of `None`, that is the standard "null" value in Odoo.

product_contract/models/sale_order.py
```diff
@@ -0,0 +1,27 @@
...
+
+from odoo import api, models
+
+
+class SaleOrder(models.Model):
+    _inherit = 'sale.order'
+
+    @api.multi
+    def action_done(self):
+        ''' If we have a contract in the order, set it up '''
```
pedrobaeza: Use triple double quote for docstring according PEP257.

product_contract/models/sale_order.py
```diff
@@ -0,0 +1,27 @@
...
+        for rec in self:
```
pedrobaeza: Save one indentation level doing:
```
for rec in self.filtered(lambda x: x.state != 'done'):
```

product_contract/tests/__init__.py
```diff
@@ -0,0 +1,6 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 LasLabs Inc.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from . import test_product_template
+from . import test_sale_order
```
pedrobaeza: Line at the end

product_contract/models/sale_order.py
```diff
@@ -0,0 +1,27 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, models
+
+
+class SaleOrder(models.Model):
+    _inherit = 'sale.order'
+
+    @api.multi
+    def action_done(self):
```
pedrobaeza: The creation of the contract should be on `action_confirm`, not `action_done`

product_contract/models/sale_order.py
pedrobaeza: Of course, this one has no sense if you change the method where it acts.

product_contract/README.rst
```diff
@@ -0,0 +1,53 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+================
+Product Contract
+================
+
+This module adds support for products to be linked to contracts. It also adds
```
pedrobaeza: Instead of contracts, you should talk about contract templates.

product_contract/README.rst
```diff
@@ -0,0 +1,53 @@
...
+functionality to automatically create a contract when a ``sale.order`` contains
+a product that implements a contract.
```
pedrobaeza: Shouldn't you add a many2one in the sales order line to keep a reference to the created contract? Even an smart-button like the one that `project` adds to the order can be good.

pedrobaeza (changes requested):

product_contract/tests/__init__.py
lasley: Preferences => File Saving => Ensure File End With EOL Marker (assuming you're still using Komodo)

tedsalmon:
Hey @pedrobaeza,

Thanks for the review! I've addressed all the points you've made :)

Thanks!

product_contract/models/sale_order.py
```diff
@@ -9,19 +9,19 @@ class SaleOrder(models.Model):
...
-                    contract_tmpl = line.product_id.contract_template_id
-                    contract = self.env['account.analytic.account'].create({
-                        'name': '%s Contract' % rec.name,
-                        'partner_id': rec.partner_id.id,
-                        'contract_template_id': contract_tmpl.id,
-                    })
-                    contract.recurring_create_invoice()
-        super(SaleOrder, self).action_done()
+            order_lines = rec.order_line.filtered(
+                lambda r: r.product_id.is_contract==True
```
pedrobaeza: No need to put `==True`. Every Python expression is evaluated as boolean when needed

product_contract/models/sale_order.py
```diff
@@ -9,19 +9,19 @@ class SaleOrder(models.Model):
...
-                for line in order_lines:
-                    contract_tmpl = line.product_id.contract_template_id
-                    contract = self.env['account.analytic.account'].create({
-                        'name': '%s Contract' % rec.name,
-                        'partner_id': rec.partner_id.id,
-                        'contract_template_id': contract_tmpl.id,
-                    })
-                    contract.recurring_create_invoice()
-        super(SaleOrder, self).action_done()
+            order_lines = rec.order_line.filtered(
```
pedrobaeza: No need to do two loops. You can do:
```
order_lines = self.mapped('order_lines').filtered(lambda r: r.product_id.is_contract)
```

tedsalmon: Thank you! This is considerably more attractive to look at than my solution :)

product_contract/models/sale_order.py
lasley: Also just a note - that was nasty PEP8 too. No spaces between the `==` and the vars? What are we, back in PHP or something? 😛

product_contract/README.rst
```diff
@@ -0,0 +1,53 @@
...
+Known issues / Roadmap
+======================
+
+* None
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/{project_repo}/issues>`_. In case of trouble, please
```
lasley: Bad URI

product_contract/__manifest__.py
```diff
@@ -0,0 +1,24 @@
...
+{
+    'name': 'Product Contract',
+    'version': '10.0.1.0.0',
+    'category': 'Contract Management',
+    'license': 'AGPL-3',
+    'author': "LasLabs, "
+              "Odoo Community Association (OCA)",
+    'website': 'https://laslabs.com',
+    'depends': [
+        'base',
```
lasley: Only time you need to depend on base is when you don't have any other depends. One can assume that all modules follow this principle, so at some point you will hit a base depends once you hit the bottom of the depends stack.

product_contract/models/product_template.py
```diff
@@ -0,0 +1,22 @@
...
+
+    is_contract = fields.Boolean('Is a contract')
+    contract_template_id = fields.Many2one(
+        comodel_name='account.analytic.contract',
+        string='Contract Template',
+    )
+
+    @api.onchange('is_contract')
+    def _change_is_contract(self):
+        """ Clear the relation to contract_template_id when downgrading
```
lasley, line 19: This docblock is out of spec. Take a look at [PEP-257](https://www.python.org/dev/peps/pep-0257/)

product_contract/models/sale_order.py
```diff
@@ -0,0 +1,26 @@
...
+        for line in order_lines:
+            contract_tmpl = line.product_id.contract_template_id
+            contract = self.env['account.analytic.account'].create({
+                'name': '%s Contract' % rec.name,
+                'partner_id': rec.partner_id.id,
+                'contract_template_id': contract_tmpl.id,
+            })
+            line.contract_id = contract.id
+            contract.recurring_create_invoice()
+        super(SaleOrder, self).action_confirm()
```
lasley: always return result from super. Otherwise you're breaking inheritance chain. Keep in mind that if a method doesn't return anything, it is still implicitly returning `None`. Knowing that it's returning `None` requires looking at the parent though, which is a no-no in modular design

product_contract/models/sale_order_line.py
```diff
@@ -0,0 +1,14 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class SaleOrderLine(models.Model):
+    _inherit = 'sale.order.line'
+
+    contract_id = fields.Many2one(
+        'account.analytic.account',
```
lasley: No need to change, but going forward I prefer that you statically define all keys in field definitions. `comodel_name=`

lasley (approved):
Thanks @tedsalmon - few comments but otherwise 👍

product_contract/README.rst
```diff
@@ -0,0 +1,53 @@
...
+   :alt: License: AGPL-3
+
+================
+Product Contract
+================
+
+This module adds support for products to be linked to contract templates.
+It also adds functionality to automatically create a contract, from the template,
+when a ``sale.order`` contains a product that implements a contract.
+
```
lasley, line 12: Missing usage section with Runbot link

lasley:
Please also make your dependencies a checklist. This will show up in the high level views as a task list that gets completed, allowing for reviewers to more easily spot PRs that are ready (and ones they should review first if they're interested in something):

image

You check the task off once the branch has been merged & you have rebased onto the production head. Now everyone is able to tell the progress without having to view the PR 🚀 

(PS I'm trying to figure out good wording for this for our Contribution docs. I'm settling with internal training first in order to figure out the verbiage after explaining a few times)

tedsalmon:
Hey @lasley,

I've addressed all your concerns :D

lasley:
You didn't handle my concern regarding the dependencies noted in https://github.com/OCA/contract/pull/49#issuecomment-273251203

tedsalmon:
@lasley - I think I have now. I felt stupid trying to figure out how to use the task list *eyeroll*

lasley:
@tedsalmon #42  has been merged, please rebase & update PR description

tedsalmon:
@lasley,

Done! I also had to fix the tests and some functionality that broke along the way.

product_contract/__manifest__.py
lasley: Remove base depend

product_contract/tests/__init__.py
lasley: NL EOF

lasley (changes requested):

tedsalmon:
Fixed -- Thanks!

lasley (approved):
Thanks @tedsalmon

lasley:
@pedrobaeza would you mind checking that @tedsalmon has attended to your review comments when you get a chance?

pedrobaeza (approved):
Squash commits at your taste and this one is good to go

lasley:
Squashed!

rafaelbn:
Hi @tedsalmon , sorry for writing here but I'm unable to create a contract from sale order

```Add logic to create contracts from sale.order that contains contract products```

I've recorded a video: https://drive.google.com/file/d/1iwH27-uwvzSZhV2koIcEq8BiYrSkkRu-/view?usp=sharing

Maybe could you give me a tip?

lasley:
@rafaelbn - looks broken to me. Seems we already have #138 for it as well. I'll take a look soon.

PR #51, opened 2017-01-20
|8.0] Add module contract_invoice_start_end_dates

alexis-via:
This is a sort of "glue" module between account_analytic_analysis and account_cutoff_prepaid.

### DIFF
--- a/contract_invoice_start_end_dates/README.rst
+++ b/contract_invoice_start_end_dates/README.rst
@@ -1,0 +1,54 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+================================
+Contract Invoice Start End Dates
+================================
+
+This module sets the *Start Date* and *End Date* on recurring invoice lines generated from a contract.
+
+The fields *Start Date* and *End Date* on invoice lines are added by the module *account_cutoff_prepaid* for Odoo v8 (module *account_invoice_start_end_dates* on Odoo v9+) hosted on the OCA project `account-closing <https://github.com/OCA/account-closing>`_.
+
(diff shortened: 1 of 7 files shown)


alexis-via:
Travis failure is caused by the module contract_invoice_merge_by_partner, not mine.

pedrobaeza:
Main branch was green. I have restarted the build to see if it's a change in upstream. If not, your module is choking with that one: https://travis-ci.org/OCA/contract/builds/191821821

bealdav:
Hi @alexis-via,
Thanks for your module.
It makes me think, that only automatically checked rules, are good rules, others are bad.
At worst, it only should be guidelines, no more
I even have convinced odoo sa to follow this rule by arguing but that's not enough.
https://github.com/odoo/odoo/pull/15070#issuecomment-273793585
Then if I have not convinced you, maybe it's because this rule is not so really good.
We all accomplish a lot of work all last five years in oca, not so easy to maintain this code 
version after version. A lot of reasons to be tired.
Odoo sa from its side have done a lot of work to improve legacy code. 
But legacy code is the future, why `res.users` ? That's the question of the newcomers in odoo ecosystem. Why not ? Whatever ? What is this question ? Go ahead to build some legacy code, this is the future.
I believe to know their is good reading on this matter there http://www.catb.org/esr/writings/cathedral-bazaar
Nothing is perfect/consistent nowhere. No problem with that.
I think now it's time to remove bad rules.
Please make the PR, to remove rules on module naming, I'll vote for it, for sure.

Thanks a lot

rafaelbn:
Hello! Is this module the same as  contract_recurring_invoicing_marker ? Thanks

pedrobaeza:
No, it isn't. Please read initial comment.

pedrobaeza:
Is this still interesting or can we close it?

pedrobaeza:
Closing as no further notice.

PR #52, opened 2017-01-29
[10.0][IMP][contract] include journal onchange

ovnicraft:
Extend ability to include extra info from _onchange_journal_id

### DIFF
--- a/contract/models/contract.py
+++ b/contract/models/contract.py
@@ -247,8 +247,9 @@ class AccountAnalyticAccount(models.Model):
             'contract_id': self.id,
             'user_id': self.partner_id.user_id.id,
         })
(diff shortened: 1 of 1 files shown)


pedrobaeza:
I'm seeing that none of them are needed: https://github.com/odoo/odoo/blob/6a7c05112bf0c07ffa7dadfe76be08f3121fd4c8/addons/account/models/account_invoice.py#L331

ovnicraft:
I will check so i redefine _onchange_journal_id on my own module and doesnt run that code, will debug and give feedback here.

rafaelbn:
Hi @ovnicraft , could you give an update or close this PR if not needed please? Thanks! 😄

PR #57, opened 2017-03-28
[MIG][9.0] contract_invoice_merge_by_partner module

cubells:
- Update README
- Update views
- Update tests

cc @Tecnativa


luismontalba (approved):
Functionality tested on runbot (I've loaded the Sales Management application to access contract menu)

chienandalu (approved):
functionality tested :+1:

rafaelbn (approved):
Tested, thanks

mourad-ehm:
Hi @pedrobaeza, Do you know why this module has not been migrated in the recent versions ? 
We have a customer need to merge invoices generated by contracts (by some criteria). I didn't see that this need is covered by the contract modules of v14.

pedrobaeza:
We haven't used since this version, but it's still interesting and it's not covered in the base module. The architecture is not the same, so the migration won't be direct I'm afraid.

mourad-ehm:
Thank you @pedrobaeza for your response. 
We should propose a solution to our customer for merging contracts in the same way of merging invoices.
But we don't delete the the merged contracts but only cancel contract lines (by probably use a renew line method).
In fact, contracts can be merged a long time after their creation, and we want preserve history of invoicing of merged contracts. 
If it accepts solution, we w'll propose new module.

pedrobaeza:
I don't have clear the possible derivations of such thing, but putting that contract merging in a separate module will avoid debates.

PR #58, opened 2017-04-03
[IMP][10.0][contract] Fix not showing recurring contract after add templating

robyf70:
@lasley Please pull to fix not showing recurring invoices

### DIFF
--- a/contract/README.rst
+++ b/contract/README.rst
@@ -49,6 +49,7 @@ Known issues / Roadmap
 ======================
 
 * Recover states and others functional fields in Contracts.
+* Switch view type to primary when migrating to v11
 
 Bug Tracker
 ===========
--- a/contract/views/account_analytic_account_view.xml
+++ b/contract/views/account_analytic_account_view.xml
@@ -5,7 +5,7 @@
         <field name="name">account.analytic.account.invoice.recurring.form.inherit</field>
         <field name="model">account.analytic.account</field>
         <field name="inherit_id" ref="analytic.view_account_analytic_account_form"/>
-        <field name="mode">primary</field>
+        <field eval="40" name="priority"/>
         <field name="arch" type="xml">
             <group name="main" position="after">
                 <separator string="Recurring Invoices"


lasley:
@robyf70 - I created that view as primary on purpose, in order to not edit all inherited views. Can you please elaborate on where the recurring invoices are not showing?

robyf70:
I've installed the patch but I cannot see the recurring invoice check box. Reverting it back to priority and eval=40 it will show up again

lasley:
@robyf70 - thanks for the followup. To confirm you are not seeing the check in `Invoicing => [Sales] Contracts`, or are you in a different area maybe?

robyf70:
@lasley I've tried it from [Accounting] -> [Sales] -> Contracts

lasley:
Ok sweet, that's the same thing just with Analytic Accounting installed.

Hrmmm this must be something with upgrading vs. installing then, as I can't replicate this on Runbot. 

This change will negatively affect some other analytic account views, but that was also the functionality originally. I won't block seeing as I don't have a better solution, but would you mind adding a note to the Known Issues section in the ReadMe saying to switch the view type back to primary when migrating to v11?

pedrobaeza:
@lasley check possible missing field on XML records before and after the change. This kind of issues should come from this.

robyf70:
@lasley Done in bb0bfa9

pedrobaeza:
I commented yesterday a possible reason. Please explore that possibility before changing this.

robyf70:
@pedrobaeza Ah! Sorry!

rafaelbn:
Hi @robyf70 ,

For me it would be much more easy to test this if describing which is the error and how to get it like I did in https://github.com/OCA/contract/issues/68

Any way I cannot reproduce this as I'm unable to create a contract template because https://github.com/OCA/contract/issues/68

Let me know how do you reproduce the error please and how do you create the templates

Thanks!! 😄

robyf70:
Hi @rafaelbn,

I did some test with some new fresh DB but the problem doesn't show up any more. I've the problem in one db in production but I can investigate a bit more tomorrow.

rafaelbn:
Hi @robyf70 , I think we should close this PR. I'm unable to reproduce your error in runbot http://3294782-10-0-9bd1a5.runbot1.odoo-community.org

Please next time write down steps for reproducing you error in runbot

I you think the error exists please ping me in this PR with the steps, I will test it again.

Thanks! 😄

rafaelbn:
Hi @bmw95 reading https://github.com/OCA/contract/pull/62 are you able to explain how to reproduce this issue?  I saw that you depends you PR in this one https://github.com/OCA/contract/pull/62#issue-222277747

woodbrettm:
@rafaelbn It actually turned out to be an unrelated issue I was experiencing but thought it was related. Removed that bit from the PR so all good now.

PR #66, opened 2017-05-06
[10.0][IMP] Add variable qty to contract template view

lasley:
Allows templating of the variable qty lines by updating the `account.analytic.contract` view.

### DIFF
--- a/contract_variable_quantity/__manifest__.py
+++ b/contract_variable_quantity/__manifest__.py
@@ -4,7 +4,7 @@
 
 {
     'name': 'Variable quantity in contract recurrent invoicing',
-    'version': '10.0.1.0.1',
+    'version': '10.0.1.1.0',
     'category': 'Contract Management',
     'license': 'AGPL-3',
     'author': "Tecnativa,"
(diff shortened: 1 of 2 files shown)


contract_variable_quantity/__manifest__.py
```diff
@@ -4,7 +4,7 @@
 
 {
     'name': 'Variable quantity in contract recurrent invoicing',
-    'version': '10.0.1.0.0',
+    'version': '10.0.1.0.1',
```
pedrobaeza: Should change minor version, not revision

pedrobaeza (approved):
Please change version

lasley:
Thanks for the review @pedrobaeza - done

rafaelbn:
Hi @lasley , sorry but I'm not able to create contract templates. I tried https://github.com/OCA/contract/issues/68 and also same error if I go to Accounting -> Configuration -> Contracts -> Contracts templates

lasley:
@rafaelbn - looks like the issue is related to pricelists. For the purpose of this review, we can just not select a pricelist when creating a contract template in order to circumvent. 

I'll submit a separate PR for the fix in order to avoid expanding the scope of this feature with a bug fix.

rafaelbn (approved):
Thanks1

Issue #55, opened 2017-03-23
[RFC] contract_auto: Automatic Payment


lasley:
This module will provide the ability to accept automatic payments on contracts. It is a continuation of the discussion [here](https://github.com/OCA/contract/issues/45#issuecomment-286616680). 

I was originally trying to abstract this to work directly on invoices, but I cannot think of a good way to do that while still maintaining a streamlined workflow. As such, this RFC is limited in scope to contracts only (unless someone has a brilliant idea).

It will add a m2o on `account.analytic.account` to relate to a `payment.token` that should be used for automatic payment. The `payment.token` will be domain restricted to children of the `partner_id` of the `account.analytic.account`.

If a `payment.token` is defined on the `account.analytic.account`, the invoice will be validated immediately after creation. A payment will then be generated using the v10 [server2server interface](https://github.com/odoo/odoo/blob/10.0/addons/payment/models/payment_acquirer.py#L587) against the `payment.token` that is defined.

It will also add an m2o on `account.analytic.account` to relate to a `mail.template`. If a template is defined, it will be emailed to the customer automatically when the invoice is validated.

It will also add an option at the contract and contract template level to require the use of automatic payment on the contract. If this option is selected, a `payment.token` will be required for the contract.

Sounds simple enough I think. Anyone have any ideas of how to abstract this beyond contracts, or some existing code I can base on?

lasley:
Can a PSC please tag as `question` and `10.0` milestone? Thanks!

vonpupp:
I would like to help with this, if possible.

I successfully developed a payment-gateway (PagSeguro) integration module however I am still learning Odoo development. What do you guys think/propose?

lasley:
@vonpupp - This feature should be independent of the payment gateway. Assuming that PagSeguro has a concept of payment tokens, we should be good to go. 

I believe we've started dev on this, I'll let @bmw95 comment on progress (LABS-411)

vonpupp:
Thank you @lasley. Is this dev open/available? Is there any way to collaborate or be able to check how things are going and follow the code/progress? I would like to help.

lasley:
@vonpupp - The ticket only moved to in progress a few days ago, so I'm not sure if there is anything of substance yet (and there's not a branch linked to our internal ticket at the moment). We always develop in the open though, so that answer is most certainly a yes. 

[...]

Issue #65, opened 2017-05-03
[RFC] website_sale_contract: Contract Signing


lasley:
This module will inject a `Contract Signing` step in the e-commerce process, immediately before payment. The user will skip this stage if there is not a product in their cart set to `is_contract`, similar to how we do for [prescriptions in medical](https://github.com/LasLabs/vertical-medical/blob/release/10.0/website_sale_medical_prescription/controllers/main.py#L17). It will depend on [`product_contract`](https://github.com/OCA/contract/tree/10.0/product_contract) and [`website_portal_contract`](https://github.com/OCA/contract/issues/45).

In order to support contract signatures, a Binary field `signature` will be added to `account.analytic.account` along with a `signatory` char field to indicate who signed it.

During this stage, the first unsigned contract in the user's cart will be displayed (using the `website_template_id` provided in `website_portal_contract`). Below that, a freeform field will be provided to type the signatory name, as well as a [`jSignature`](https://github.com/odoo/odoo/search?utf8=%E2%9C%93&q=signature+jSignature&type=) for the actual signing. 

Clicking `Accept` will continue either loop the above operation for the next unsigned contract in the cart, or proceed to payment.

The gotcha here is that the contract hasn't actually been created until after the sale is completed, which is after payment during e-commerce. This means that we will need to temporarily store the signature information, then assign it correctly during the `action_done` when the contract is actually created.

This module will also add the signer's information into the contract's detailed view in the website portal (provided in `website_portal_contract`), as well as show it in the backend for employees viewing the contract through the standard form view.

Anyone see something I'm missing, or a better way to go about this?

pedrobaeza:
:+1: for me

lasley:
Closing to track in #88

Issue #45, opened 2016-12-02
[RFC] website_portal_contract: Website Contract Management


lasley:
This module will provide website views for customer management of their existing contracts. 

It will add a new field `website_template_id` to the `account.analytic.contract` model (and subsequently the `account.analytic.account` model). This field will act in the same way that the quote template works for website sales. A default template will be provided.

It will provide a new section in the website `My Account`, similar to the existing `Your Quotes`, `Your Orders`, etc & will be called `Your Contracts`.

In the `Your Contracts` listing, the following fields will be displayed in the table:

* Display Name
* Debit
* Credit
* Balance

Clicking a contract in the listing will take you to the website template associated with that contract.

rafaelbn:
It's more website_portal_contract, isn't it?

lasley:
@rafaelbn - Yup that's a better name for sure

dreispt:
Add to the wishlist a "Your Issues" feature, and "Open an Issue"

pedrobaeza:
Isn't that already on website_issue?

dreispt:
You're probably right. Ignore that then.

vonpupp:
Using a module such as this would it be possible to automatically charge a credit card? Such as a subscription monthly payment?

lasley:
@vonpupp - That is not within scope right now, but autopay is part of my long-term plans. That will likely be in a separate module though in order to allow for usage outside of this context.

vonpupp:
Great @lasley. I am not aware of any module right now that allows recurrent payments. Perhaps the closest might be the [memberships] module. I think it would be very nice to have something similar to the Odoo enterprise feature [subscriptions].

There is also an effort on the [donation] module to achieve recurrent payments. Wouldn't it be possible to create some sort of "fit all" recurrent payment module that integrates with contracts, memberships and donations?

[memberships]: https://github.com/OCA/vertical-association/issues/19
[subscriptions]: https://www.odoo.com/page/subscriptions
[donation]: https://github.com/OCA/donation/issues/22

pedrobaeza:
AFAIK Odoo enterprise doesn't handle either automatic recurrent payment. And this is very closely related with payment gateways (Paypal and so on) to allow this option.

vonpupp:
I have not tried this myself but Fabien [stated] that it is supported on the Enterprise version @pedrobaeza. I believe this feature would be very important and would boost Odoo adoption IMHO.

[...]

PR #67, opened 2017-05-06
[10.0] [ADD] contract_auto: Automatic Payment For Contracts

lasley:
Contract Auto (RFC #55)
=============

This module allows for the configuration of automatic payments on invoices that are created by a contract.

Usage
=====

Enable Automatic Payment
------------------------

* Navigate to a customer contract
* Check the `Auto Pay?` box to enable automatic payment
* Configure the options as desired
* Set the `Payment Token` to the payment token that should be used for automatic payment

Automatic Payment Settings
--------------------------

The following settings are available at both the contract and contract template level:

| Name | Description |
|------|-------------|
| Invoice Message | Message template that is used to send invoices to customers upon creation. |
| Payment Retry Message | Message template that is used to alert a customer that their automatic payment failed for some reason and will be retried. |
| Payment Fail Message | Message template that is used to alert a customer that their automatic payment failed and will no longer be retried. |
| Auto Pay Retries | Amount of times to attempt an automatic payment before discontinuing and removing the payment token from the contract/account payment method. |
| Auto Pay Retry Hours | Amount of hours that should lapse until retrying failed payments. |

Payment Token
-------------

A valid payment token is required to use this module. These tokens are typically created during the `website_sale` checkout process, but they can also be created manually at the acquirer.

A payment token can be defined in one of two areas:

* Contract - Defining a payment token in the contract will allow for the use of this token for automatic payments on this contract only.
* Partner - Defining a payment token in the partner will allow for the use of this token for automatic payments on all contracts for this partner that do not have a payment token defined.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,26 +7,26 @@ python:
 
 addons:
   apt:
-# only add the two lines below if you need wkhtmltopdf for your tests
-#    sources:
-#      - pov-wkhtmltopdf
+    sources:
+      - pov-wkhtmltopdf
     packages:
       - expect-dev  # provides unbuffer utility
       - python-lxml  # because pip installation is slow
       - python-simplejson
       - python-serial
       - python-yaml
-#      - wkhtmltopdf  # only add if needed and check the before_install section below
+      - wkhtmltopdf
 
 # set up an X server to run wkhtmltopdf.
-#before_install:
-#  - "export DISPLAY=:910.0"
-#  - "sh -e /etc/init.d/xvfb start"
+before_install:
+  - "export DISPLAY=:910.0"
+  - "sh -e /etc/init.d/xvfb start"
 
 env:
   global:
   - VERSION="10.0" TESTS="0" LINT_CHECK="0" TRANSIFEX="0"
   - TRANSIFEX_USER='transbot@odoo-community.org'
+  - WKHTMLTOPDF_VERSION="0.12.4"
   - secure: ArnbVaF5+ry6zVysZ7HA9xcnQrodo8FsXXcon9yINTYRfDC9QEr+/mTLAPRl+lVLdtWV2GuGuX0vPYzBxFWpY3LR0BpKqXzx0G51s94zR2WWEmizYFzFhpAuIxoU4CYNckHFSUaPAQJhwB/pYx9+H/W6bMjG/VnZBq+AmBJ2Kh0=
 
   matrix:
--- a/contract_payment_auto/README.rst
+++ b/contract_payment_auto/README.rst
@@ -1,0 +1,89 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+=====================
+Contract Auto Payment
+=====================
+
+This module allows for the configuration of automatic payments on invoices that are created by a contract.
+
+Usage
+=====
+
+Enable Automatic Payment
+------------------------
+
+* Navigate to a customer contract
+* Check the `Auto Pay?` box to enable automatic payment
+* Configure the options as desired
+* Set the `Payment Token` to the payment token that should be used for automatic payment
+
+Automatic Payment Settings
+--------------------------
+
+The following settings are available at both the contract and contract template level:
+
+| Name | Description |
+|------|-------------|
+| Invoice Message | Message template that is used to send invoices to customers upon creation. |
+| Payment Retry Message | Message template that is used to alert a customer that their automatic payment failed for some reason and will be retried. |
+| Payment Fail Message | Message template that is used to alert a customer that their automatic payment failed and will no longer be retried. |
+| Auto Pay Retries | Amount of times to attempt an automatic payment before discontinuing and removing the payment token from the contract/account payment method. |
+| Auto Pay Retry Hours | Amount of hours that should lapse until retrying failed payments. |
+
+Payment Token
+-------------
+
+A valid payment token is required to use this module. These tokens are typically created during the `website_sale` checkout process, but they can also be created manually at the acquirer.
+
+A payment token can be defined in one of two areas:
+
+* Contract - Defining a payment token in the contract will allow for the use of this token for automatic payments on this contract only.
+* Partner - Defining a payment token in the partner will allow for the use of this token for automatic payments on all contracts for this partner that do not have a payment token defined.
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/110/10.0
+
+Known issues / Roadmap
+======================
+
+* None
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/contract/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
+help us smash it by providing detailed and welcomed feedback.
+
+Credits
+=======
+
+Images
+------
+
+* Odoo Community Association: `Icon <https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.svg>`_.
+
+Contributors
+------------
+
+* Dave Lasley <dave@laslabs.com>
+
+
+Maintainer
+----------
+
+.. image:: https://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit https://odoo-community.org.
--- a/contract_payment_auto/__init__.py
+++ b/contract_payment_auto/__init__.py
@@ -1,0 +1,5 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 LasLabs Inc.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from . import models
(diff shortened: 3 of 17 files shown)


vonpupp:
Thank you @lasley. Yesterday I had some problems during installation (XML issues) but now I could install it fine. I just created a contract and I will have to wait to see if everything goes fine. I wonder a couple of things:
(1) Wouldn't be interesting to define a payment-gateway to be used on a contract?
(2) How do I define a credit card number for the person who is the contract owner?

What if the customer decides to switch to another payment method which involves switching to another payment-gateway? (this scenario is related to my first question).

I would like to test the module using stripe, which implements s2s methods, however I don't understand how a specific person will be charged for a contract. Do you know how to debug this to try to understand better the process? What first comes to my mind is to save the customer on the stripe backend and link a credit card number as primary so it can be automatically retrieved during the payment, but I will have to test that.

lasley:
The payment token represents the credit card, which is linked to a partner and acquirer. Changing either is as easy as changing the payment token. 

The catch here is that the credit card already has to be saved to the partner's account as a payment token. This happens during the website_sale checkout process if the customer checks the option to remember their card. You can also create a token at your acquirer manually and add into Odoo for use.

rafaelbn:
Hi @lasley ,

Please could you add a usage section detailed in order to know what to test?

After seeing this PR I have read again https://github.com/OCA/contract/issues/55 and https://github.com/OCA/contract/issues/45#issuecomment-286616680 but I have some doubts about how it works.

You want that contract set invoices paid base on? I read "This module adds logic to automatically pay invoices that are created by a contract" but I don't understand why you want invoices paid without knowing if they are paid o not really with a transfer, bank card, paypal or similar.

I'm interested.

Thanks

lasley:
@rafaelbn - readme updated & instructions added to the main comment of this PR. Let me know if this doesn't clarify.

vonpupp:
@lasley, thank you very much for your guidance; It makes total sense now! Stripe is missing this important feature (saving the credit card). I am trying the module using authorize, however even though I check the "remember the CC details", it doesn't seem to be remembered at all. Truth to be told I am still struggling with authorize itself (the payment gets processed properly but the return URL doesn't seem to work, I need to further dig into this). I assume that the payment.token is created prior redirecting to authorize.net website and the failed return to the odoo website has nothing to do with it (not 100% sure though).

I tried to create a payment token manually by going (with developer mode) to Accounting / Configuration / Payments / Saved Payment Data, however I am not sure how to fill the acquirer Ref. I guess it is the token which maps to the record with the payment-gateway website record (some sort of id).

Unfortunately I am having some troubles to test this out, since as for today, the only two acquirers that I am aware that support remembering the payment details are authorize and ogone. Ogone is a nightmare to setup (I need more time) and authorize doesn't seem to remember anything.

If I achieve something with authorize I will be more than happy to document it along with the readme (or in a separated folder)

lasley:
@vonpupp - the payment token memorization definitely works with a proper Authorize setup, so I would assume it's just something in the configuration. Make sure that you're using a sandbox account set to live mode in the Authorize console, and that you set the acquirer to sandbox mode in Odoo.

Regarding Stripe, I see all of the server to server methods implemented in the v10 acquirer, such as the [token creation](https://github.com/odoo/odoo/blob/10.0/addons/payment_stripe/models/payment.py#L174)

vonpupp:
Thank you very much @lasley. Yes indeed the problem is on my side. Apparently the payment information gets saved once it succeeds. Authorize.net does not seem to be documented. I have found [this issue], which pretty much describes my problem. I have configured the URLs accordingly however the problem persists (Default Receipt URL is <URL>/shop/confirmation and Default Relay Response URL is <URL>/payment/authorize/return).

I configured also the silent URL to be a [request bin], to check out the POST parameters coming back from authorize.net so I try to replicate the request using curl. The sad part is that I always get a 404 on odoo when trying to access route `/payment/authorize/return/`, I don't know why though so I need to investigate the issue further.

Sorry that I haven't been helpful in testing your module. Hopefully I will manage to solve all this issues soon. Unfortunately Stripe do not allow to remember the payment details.

[this issue]: https://github.com/odoo/odoo/issues/14007
[request bin]: http://requestb.in/137j5x21?inspect

lasley:
> Unfortunately Stripe do not allow to remember the payment details.

It does though, I linked the token creation above. I definitely see it taking credit card numbers, then turning it into a payment token. Maybe try skipping the process entirely & just use [`stripe_create` directly](https://github.com/odoo/odoo/blob/10.0/addons/payment_stripe/models/payment.py#L174)

lasley:
This is ready for review if a PSC would please mark as such. I just need to fix a few missing test holes, but no functional code

vonpupp:
@lasley thank you for your patience! I finally could make authorize to work. I had to destroy the server and recreate it again, something very very weird was going on with the controller routes and 404 error responses (I think I messed up some python dep).

I have tested stripe with the regular process and I noticed that the "remember" check is on a different place. It resides on the stripe.js form shown to the user and when checked it asks for the phone number. On future purchases, the phone number is used to receive a code which authenticates that user within Stripe and allows the user to push the `pay now` button.

I am manually creating a payment.token however I am not sure what to use on the acquirer ref field. I guess is the customer id which I can get from the Stripe interface `cus_xxx`. Could you please clarify this? I couldn't understand properly, sorry. On every credit card charge on the autopay feature from contracts, doesn't Stripe send you an SMS with a code?

Thank you very much once again for your patience and help. I am learning a lot. Hopefully I will be able to implement this at the NGO I am a volunteering for.

lasley:
TBH I'm not sure how Stripe works, but I think you're confusing Stripe's automatic payment service with that of what I am doing here. We are not utilizing the automatic payment functionality of any of the acquirers, but are instead using the saved payment tokens to charge cards that are already known.

vonpupp:
I don't think so but now that authorize.net works, I can finally test the module, understand it better and compare it properly with stripe =)

I was thinking of other use cases, i.e: send payment email notification with a `pay now` button (useful for Paypal), or attaching a banking slip (money order, "boleto" in portuguese, not sure about the translation). I will try to add these features when I am more odoo-fluent.

vonpupp:
@lasley, Excuse me to annoy you again. I would like to please confirm something.

I have noticed that the invoices even when paid, they still remain as open state.
2017-05-12-212009_2720x1024_scrot

Here is a screenshot of the payment confirmations.
2017-05-12-211954_2720x1024_scrot

Should the invoice status be changed to paid/done automatically, or this is the expected behavior and then some sort of conciliation process should take care of those status change? Thank you very much.

lasley:
@vonpupp - I don't think any of that is controlled by my module. There are acquirer settings that would dictate what happens with the invoice after payment.

vonpupp:
Thank you @lasley. I guess you are referring to the option: "Authorize & capture the amount, confirm the SO and auto-validate the invoice on acquirer confirmation". This work ok when using the regular ecommerce flow (which creates a S.O), but does not seem to validate the invoice when used in contracts (maybe because no S.O is created). Just for your information.

I am personally very happy with the module and I want to thank you very much for your efforts. I will try to dig into the Stripe and "remember the credit card" issue.

Thank you once again and big kudos!

lasley:
@vonpupp - Ah yup you're right about that. So we'd need some sort of option in the acquirer added to perform similar functionality. That's a good module idea - you should make an RFC for it (and/or write it if you have time).

Sweet so this module works? If so, can you please add an approval in the "Files Changed" tab & just note that it was from a functional standpoint only (unless you also reviewed the code).

vonpupp (approved):
LGTM

vonpupp:
@lasley. I approved as requested.

Regarding the option in the acquirer to automatically change the status, do you think it would be usable by all the acquirers or is something that needs to be done per-acquirer basis? Could you point me into a sane starting point, please?

Perhaps writing a subscription module that would create a new S.O. instead of an invoice, so the Odoo workflow can be followed normally would be interesting. This might be easier to maintain in the future (speculations though). Any thoughts? With my little Odoo knowledge, writing this subscription module will take me quite a long time and I would rather first put into production the Odoo I am preparing for the NGO first so that I can later focus on this (they need it quite urgently). But it something I would like to try in the future.

lasley:
> Could you point me into a sane starting point, please?

Totally. First, I think you should make an RFC [like this one](https://github.com/OCA/contract/issues/65) so we can move the conversation into an appropriate context.

Second, take a look at how it happens in [website_portal_sale](https://github.com/odoo/odoo/blob/5bd8f97fa2fd572614f321a747b3cac140bb8a1d/addons/website_portal_sale/models/payment.py#L61). It'll definitely work for all acquirers, we're just modifying invoice statuses. Our implementation will be much simpler.

> Perhaps writing a subscription module that would create a new S.O. instead of an invoice, so the Odoo workflow can be followed normally would be interesting. 

Generating a sale per contract invoicing would not be ideal. A sale is a representation of an agreement to purchase something - in the context of a contract, that's the agreement to purchase the recurring billing. Invoices are related specifically to requesting and subsequently receiving payment from customers.

vonpupp:
Amazing, @lasley. Thank you very much!

I created #70 for this purpose. Please feel free to give your impressions. Just to be clear, you provided the `website_portal_sale` as an example, but I don't need to modify nothing there, right?

Currently I am working on the missing parts of the `payment_stripe` module, it seems to miss some parts to properly work with s2s. Afterwards I will dig into #70 by my own to try to contribute :). Then I would like to have a look at the #45 which seems to be the natural progression of your work here.

rafaelbn:
Hi, could I propose `contract_payment_auto` ? contract_auto is ambiguous IMHO. 

Testing I get and error when filling Payment token in the partner check, Do you know why?:

lasley:
@rafaelbn - I agree with the rename.

Looks like I forgot to add `payment_token_ids` to the partner view. Should be good now!

rafaelbn:
Hi @lasley travis is red, could you take a look please?

lasley:
Alright we're 🍏 !

lasley:
Any PSC have thoughts on this one?

rafaelbn:
Hi @lasley I'm rebuilding for testing but it looks runbot is broken everywhere... 😞

lasley:
@rafaelbn - I rebuilt and it looks good now. must have been the space issue from earlier

rafaelbn (approved):
Tested 👍 

@lasley do you want to squash before merge?

PR #82, opened 2017-08-03
[9.0][FIX] contract_show_invoice: Error if create invoice with view context

carlosdauden:
@Tecnativa

### DIFF
--- a/contract_show_invoice/views/contract_view.xml
+++ b/contract_show_invoice/views/contract_view.xml
@@ -3,26 +3,13 @@
     <data>
 
         <record id="act_analytic_invoices" model="ir.actions.act_window">
-            <field name="context">{'search_default_analytic_account_ids':
-                [active_id], 'default_analytic_account_ids': active_id}</field>
             <field name="name">Invoices</field>
             <field name="res_model">account.invoice</field>
-            <field name="domain">[('analytic_account_ids','=',active_id)]</field>
-            <field name="search_view_id" ref="account.view_account_invoice_filter" />
-        </record>
-
-        <record id="action_contract_show_invoice_tree" model="ir.actions.act_window.view">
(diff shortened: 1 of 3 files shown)


pedrobaeza:
Why this change? It was OK before

contract_show_invoice/views/contract_view.xml
```diff
@@ -3,26 +3,14 @@
     <data>
 
         <record id="act_analytic_invoices" model="ir.actions.act_window">
-            <field name="context">{'search_default_analytic_account_ids':
-                [active_id], 'default_analytic_account_ids': active_id}</field>
```
carlosdauden, line 7: @pedrobaeza this context is useless because analytic_account_ids is a computed field and the search is done by domain

pedrobaeza: OK, but don't change the rest...

pedrobaeza: The search is needed anyway

carlosdauden:
This module would not be better in analytic branch?

pedrobaeza:
Well, yes. And with other name even...

contract_show_invoice/views/contract_view.xml
carlosdauden: Why is needed?

pedrobaeza: For filtering the invoices with that analytic contract. It's the way to do it.

dreispt:
> This module would not be better in analytic branch?

That's best done in a next version migration (v10).
You can put that on the README's roadmap if you like.

rafaelbn:
hi @pedrobaeza , please take a look again. Thanks!

pedrobaeza (approved):

PR #84, opened 2017-08-04
[9.0][FIX] contract: Invoice supplier form view from contract link

carlosdauden:
@Tecnativa

### DIFF
--- a/contract/tests/test_contract.py
+++ b/contract/tests/test_contract.py
@@ -15,7 +15,6 @@ class TestContract(SavepointCase):
             'name': 'Partner test',
             'customer': True,
         })
-        # cls.partner = cls.env.ref('base.res_partner_2')
         cls.product = cls.env.ref('product.product_product_2')
         cls.product.description_sale = 'Test description sale'
         cls.contract = cls.env['account.analytic.account'].create({
--- a/contract/tests/test_contract.py
+++ b/contract/tests/test_contract.py
@@ -11,7 +11,11 @@ class TestContract(SavepointCase):
     @classmethod
     def setUpClass(cls):
         super(TestContract, cls).setUpClass()
-        cls.partner = cls.env.ref('base.res_partner_2')
+        cls.partner = cls.env['res.partner'].create({
+            'name': 'Partner test',
+            'customer': True,
+        })
+        # cls.partner = cls.env.ref('base.res_partner_2')
         cls.product = cls.env.ref('product.product_product_2')
         cls.product.description_sale = 'Test description sale'
         cls.contract = cls.env['account.analytic.account'].create({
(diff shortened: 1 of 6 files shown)


pedrobaeza (changes requested):
Change module version

contract/views/res_partner_view.xml
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+    <record id="act_res_partner_2_contract" model="ir.actions.act_window">
+        <field name="name">Contracts</field>
+        <field name="res_model">account.analytic.account</field>
+        <field name="view_type">form</field>
+        <field name="view_mode">tree,form</field>
+        <field name="context">{
+            'search_default_partner_id': active_id,
```
pedrobaeza: This won't show contacts' contracts

carlosdauden: Ok. Yes now :wink:

contract/tests/test_contract.py
```diff
@@ -11,7 +11,11 @@ class TestContract(SavepointCase):
     @classmethod
     def setUpClass(cls):
         super(TestContract, cls).setUpClass()
-        cls.partner = cls.env.ref('base.res_partner_2')
+        cls.partner = cls.env['res.partner'].create({
+            'name': 'Partner test',
+            'customer': True,
+        })
+        # cls.partner = cls.env.ref('base.res_partner_2')
```
pedrobaeza: Don't comment, but remove

contract/views/contract.xml
```diff
@@ -85,6 +87,9 @@
             <field name="model">account.analytic.account</field>
             <field name="inherit_id" ref="analytic.view_account_analytic_account_search"/>
             <field name="arch" type="xml">
+                <field name="partner_id" position="attributes">
+                    <attribute name="operator">child_of</attribute>
```
pedrobaeza: Better to create another different line with this behavior, so that you can search by exact partner, or by the partner and its childs.

pedrobaeza (changes requested):
In tests, add a child, assign a contract, and check that the count is correct in both the child and the parent.

contract/views/contract.xml
carlosdauden: To search by exact partner you can use advanced search or group by partner.
Alternatives appears a strange code

pedrobaeza: But that's not the problem. The problem is that you are modifying the behavior of this search in all the account analytic views. Another alternative is to make a primary view and put it only on contract menu action, not in the rest.

rafaelbn:
@pedrobaeza I think @carlosdauden attend you comments, so please review. After this I will test it in runbot. thanks!

pedrobaeza:
I don't think so

carlosdauden:
@pedrobaeza changes done

pedrobaeza (approved):

pedrobaeza:
@rafaelbn please review

rafaelbn (approved):
tested in runbot 👍 

Please @carlosdauden could you make PR for v10 please?

PR #90, opened 2017-08-22
[10.0] contract: Be totally deterministic about which views to use + copyright

pedrobaeza:
* Having a primary view that is not explicitly declared to be used and w/o priority makes Odoo to choose between one of them randomly (well, not exactly, but kind of), so we put here which views to use.
* I have also put tree view as primary and put a large priority for not being selected on other actions that don't have these explicit views.
* A friendly name in views is also assigned.
* Copyrights in manifest updated.

cc @Tecnativa

### DIFF
--- a/contract/__manifest__.py
+++ b/contract/__manifest__.py
@@ -1,11 +1,15 @@
 # -*- coding: utf-8 -*-
-# © 2004-2010 OpenERP SA
-# © 2016 Carlos Dauden <carlos.dauden@tecnativa.com>
+# Copyright 2004-2010 OpenERP SA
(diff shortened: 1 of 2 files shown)


pedrobaeza:
@lasley this issue happened when you changed to templating system, and it was due to this. I hit this migrating a database.

rafaelbn (approved):
runbot tested 👍 
solved issue in production

lasley (approved):
TIL thanks @pedrobaeza 

I wonder if this fixes #58

pedrobaeza:
Yeah, totally. I was looking to that reference because I remembered  that someone had this problem also.

PR #91, opened 2017-08-22
[IMP] travis: Remove `UNIT_TEST` build

lasley:
TBH I'm not 100% why this is like this, but IMO it's just wasting Travis power by testing a subset of the same thing.


pedrobaeza (approved):
Definitively. This is an old remnant of a try to make isolated tests of each module, but there's not too much sense.

lasley:
> This is an old remnant of a try to make isolated tests of each module

This would be pretty awesome, but yeah totally not feasible - particularly with Runbot. I'll submit backports then

PR #92, opened 2017-08-22
[FIX] contract: Template lines handling

### LINKED ISSUE #80: Contract template is useless because it creates empty invoices
I have tried to use contract templates in my own instalation of odoo 10 and in oca runbot and that is what happens:

- I create a contract template with a product

- I use this contract template in a contract

- the resulting invoice is empty with no product

You can add the product directly to the contract and it works, but if you use the template , then the invoice is empty.

lasley:
* Update contract template lines handling to fix #80

cc @fgarcia-humanoide

### DIFF
--- a/contract/__manifest__.py
+++ b/contract/__manifest__.py
@@ -5,7 +5,7 @@
 
 {
     'name': 'Contracts Management - Recurring',
-    'version': '10.0.1.1.0',
+    'version': '10.0.1.1.1',
     'category': 'Contract Management',
     'license': 'AGPL-3',
     'author': "OpenERP SA, "
--- a/contract/__manifest__.py
+++ b/contract/__manifest__.py
@@ -5,7 +5,7 @@
 
 {
     'name': 'Contracts Management - Recurring',
-    'version': '10.0.1.1.1',
+    'version': '10.0.2.0.0',
     'category': 'Contract Management',
     'license': 'AGPL-3',
     'author': "OpenERP SA, "
--- a/contract/models/account_analytic_contract_line.py
+++ b/contract/models/account_analytic_contract_line.py
@@ -17,34 +17,3 @@ class AccountAnalyticContractLine(models.Model):
         required=True,
         ondelete='cascade',
     )
-
-    @api.multi
-    @api.onchange('product_id')
-    def _onchange_product_id(self):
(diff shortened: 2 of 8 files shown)


contract/models/account_analytic_invoice_line.py
```diff
@@ -16,23 +16,44 @@ class AccountAnalyticInvoiceLine(models.Model):
     _name = 'account.analytic.invoice.line'
 
     product_id = fields.Many2one(
-        'product.product', string='Product', required=True)
+        'product.product',
```
lasley, line 19: Reviewers - sorry I killed the diff here - I was having a hard time auditing the fields for required, etc. so had to clean up the spacing

pedrobaeza:
Why the coverage drop?

lasley:
I missed a spot!

contract/models/account_analytic_contract_line.py
```diff
@@ -0,0 +1,50 @@
...
+    analytic_account_id = fields.Many2one(
+        string='Contract',
+        comodel_name='account.analytic.contract',
+        required=True,
+        ondelete='cascade',
+    )
+
+    @api.multi
+    @api.onchange('product_id')
+    def _onchange_product_id(self):
```
pedrobaeza: Why this extra method is needed?

lasley: The `_onchange_product_id` in `account.analytic.invoice.line` uses the partner pretty heavily, which isn't available in the `contract.line`. This method performs similar operation, just with the user partner instead.

I guess alternatively I could just add some try/excepts in the invoice line. I'll try that

contract/__manifest__.py
```diff
@@ -5,7 +5,7 @@
 
 {
     'name': 'Contracts Management - Recurring',
-    'version': '10.0.1.1.0',
+    'version': '10.0.1.1.1',
```
pedrobaeza: Module version should raise at least to `10.0.1.2.0`, but this can be considered a huge change and it should be `10.0.2.0.0`

pedrobaeza (approved):

hurrinico (approved):
Great! :+1: 
Tested Locally works well!

PR #95, opened 2017-08-22
[10.0] Add modules agreement_account and agreement_sale

alexis-via:
See README.rst of the modules for more info. This PR is the result of a discusson that I started in the contributors mailing list on July 13th 2017 with subject "Contract without analytic".

### DIFF
--- a/agreement_account/__manifest__.py
+++ b/agreement_account/__manifest__.py
@@ -4,7 +4,7 @@
 
 {
     'name': u'Agreement (Account)',
-    'summary': "Adds a sale agreement object linked to a customer invoice",
+    'summary': "Adds an agreement object linked to an invoice",
     'version': '10.0.1.0.0',
     'category': 'Contract',
     'author': "Akretion,Odoo Community Association (OCA)",
@@ -13,8 +13,8 @@
     'depends': ['account'],
     'data': [
         'security/ir.model.access.csv',
-        'security/sale_agreement_security.xml',
-        'views/sale_agreement.xml',
+        'security/agreement_security.xml',
+        'views/agreement.xml',
         'views/account_invoice.xml',
         ],
     'demo': ['demo/demo.xml'],
--- a/agreement_account/__manifest__.py
+++ b/agreement_account/__manifest__.py
@@ -1,0 +1,22 @@
+# -*- coding: utf-8 -*-
+# © 2017 Akretion (Alexis de Lattre <alexis.delattre@akretion.com>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    'name': u'Agreement (Account)',
+    'summary': "Adds a sale agreement object linked to a customer invoice",
+    'version': '10.0.1.0.0',
+    'category': 'Contract',
+    'author': "Akretion,Odoo Community Association (OCA)",
+    'website': 'http://www.akretion.com',
+    'license': 'AGPL-3',
+    'depends': ['account'],
+    'data': [
+        'security/ir.model.access.csv',
+        'security/sale_agreement_security.xml',
+        'views/sale_agreement.xml',
+        'views/account_invoice.xml',
+        ],
+    'demo': ['demo/demo.xml'],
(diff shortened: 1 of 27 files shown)


agreement_account/__manifest__.py
```diff
@@ -0,0 +1,22 @@
+# -*- coding: utf-8 -*-
+# © 2017 Akretion (Alexis de Lattre <alexis.delattre@akretion.com>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    'name': u'Agreement (Account)',
```
bealdav, line 6: `u` required ?

bealdav (approved):

bealdav:
Nice work for me. Thanks Alexis :+1: 
Just a simple remark: on my side I think we could name the module `agreement` and the main model `agreement.agreement`. Then we could use it sale side and purchase side.

bealdav:
ping @elicoidal @jgrandguillaume

alexis-via:
@bealdav I took into account your remark: I renamed the model 'sale.agreement' to 'agreement'. Therefore, I also added a 'type' field (sale or purchase) on it and the field is now also displayed on supplier invoices.
I also updated all the modules impacted by this change in OCA/edi and OCA/l10n-france... it was not a small change !

pedrobaeza:
Please don't call any field `type`, as you are overwriting Python keyword. Call it `agreement_type` for example.

bealdav:
@pedrobaeza another option is to name `type_`

bealdav (approved):

carlosdauden (approved):

rousseldenis:
LGTM. 

@alexis-via  I ask myself if this could interfere with this :

https://github.com/odoo/odoo/blob/875dd9221f80ea1b60cd9ee8a005d71246734a2b/addons/purchase_requisition/models/purchase_requisition.py#L9

agreement_account/models/agreement.py
```diff
@@ -0,0 +1,49 @@
...
+
+from odoo import models, fields
+
+
+class Agreement(models.Model):
+    _name = 'agreement'
+    _description = 'Agreement'
+
+    code = fields.Char(
+        string='Code', required=True, copy=False)
```
rousseldenis: string is not needed anymore in that case (and below too).

agreement_account/models/agreement.py
```diff
@@ -0,0 +1,49 @@
...
+    active = fields.Boolean(string='Active', default=True)
+    signature_date = fields.Date(string='Signature Date')
+    out_invoice_ids = fields.One2many(
+        'account.invoice', 'agreement_id', string='Customer Invoices',
+        readonly=True, domain=[('type', 'in', ('out_invoice', 'out_refund'))])
+    in_invoice_ids = fields.One2many(
+        'account.invoice', 'agreement_id', string='Supplier Invoices',
+        readonly=True, domain=[('type', 'in', ('in_invoice', 'in_refund'))])
+
+    def name_get(self):
```
rousseldenis, line 35: Must decorate api.multi

rousseldenis (changes requested):
Code review. Minor comments.

Could we add maybe a date range for agreement validity (not required field)?

Could you add tests ?

agreement_account/models/agreement.py
bealdav: As far as I know all methods which inherit of models.Model are `api.multi` by default in v10 (dixit Antony Lesuisse @ last odoo days). Then we should only specify decorator when don't want this behavior.
It allows to making the code lighter. 
Sure it was different in previous versions.

alexis-via:
There are certainly many fields that could be added on the "agreement" object. It will depend very much on what you will use it for. I only added the very minimum here. I don't mind adding a dependancy on date_range and add such an optional field, but there are so many that can be added too... I tried to keep it minimal.
About tests, as you can see, these modules contain almost no code... so there is no real interesting things to test. The main objective of these modules is to store information that we need for electronic invoices.

PR #96, opened 2017-09-01
[9.0][ADD] contract purchase

funbaker:
Hello,

this pull requests adds contract functionalities for purchases and incoming invoices, which is archived by modifying the journal.

# Prerequisites
- [ ] #97 

# TODO
* Tests, Translations
* Rethinking about the placement of the `type` attribute in this module.
* Find a better way than overwriting `fields_get()` to change field string.


pedrobaeza:

funbaker:
@pedrobaeza it will. including the prerequisite PR.

PR #97, opened 2017-09-06
[9.0][IMP] Add type to analytic account / contract

funbaker:
Prerequisite for https://github.com/OCA/contract/pull/96

### DIFF
--- a/contract/models/contract.py
+++ b/contract/models/contract.py
@@ -102,6 +102,11 @@ class AccountAnalyticAccount(models.Model):
             ('company_id', '=', company_id)]
         return self.env['account.journal'].search(domain, limit=1)
 
+    contract_type = fields.Selection(
+        selection=[
+            ('sale', _('Sale'))
+        ], default='sale'
+    )
     pricelist_id = fields.Many2one(
         comodel_name='product.pricelist',
         string='Pricelist')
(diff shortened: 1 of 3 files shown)


pedrobaeza:

funbaker:
Yes, but i got one question about this.

Tests are failing because of ``ERROR openerp_test openerp.addons.contract_recurring_analytic_distribution.tests.test_contract_recurring_distribution: ` ValueError: Wrong value for account.analytic.account.type: 'contract'`` caused by https://github.com/OCA/contract/blob/9.0/contract_recurring_analytic_distribution/tests/test_contract_recurring_distribution.py#L24-L36 but i cannot find the place where ``account_analytic_account.type`` is defined.

Do you have an idea?

pedrobaeza:
You shouldn't use the word `type`, as it's a reserved word in Python. Use `contract_type`.

funbaker:
Will do, but isn't it used in odoo/odoo a lot?

pedrobaeza:
Yes, indeed, but it can cause several glitches. I don't know if the Travis problem is caused here by that or by anything else.

pedrobaeza:
@funbaker you can see that this has been adopted for v11 for generating purchases and finally without a new module. Are you interested in finishing this for v9 or should we close it?

funbaker:
In this case i'll rather backport it to v10 in a new PR.

PR #79, opened 2017-07-10
[ADD] contract_digitized_signature

cubells:
[WIP] Depends on
- [x] https://github.com/OCA/web/pull/666  
- [x] https://github.com/OCA/contract/pull/81

New module that allows to add digitized signature to contracts and send by mail.

cc @Tecnativa

PR #88, opened 2017-08-10
[10.0][ADD] website_sale_contract

woodbrettm:
- [x] Depends #62 

### TODO
- [ ] JS Tests
- [ ] Backend Tests
- [x] Contract step doesn't turn Green after done
- [ ] Add security
- [x] Ensure users can't go to unsigned contracts in UI using url.
- [x] Ensure users can't go to contracts in other checkout wizards using url.
- [x] Ensure fields highlighted in JS Signing Form when errors
- [x] Ensure proper fields are populated in account.analytic.account when converting temp contracts to account.analytic.account.
- [x] Ensure contracts are converted to account.analytic.accounts at correct stage (currently done during confirmation stage)
- [x] Fix pricing section in contract page.
- [x] The Confirmation Message in the contract signing modal is incomplete. (i.e. "I accept the terms of . . .", then does signature and types name.) 
- [x] Adjust top margin in mobile for contract header name

PR #102, opened 2017-10-09
[ADD] contract_digitized_signature

cubells:
New module that allows to add digitized signature to contracts and send by mail.

cc @Tecnativa

Related: #79


lasley:
Hmmmm I wonder if we need to think about integration of this against #88, which also has signature.

I'm curious though, what is the purpose of the signature field in the backend view? Is this instead for an employee to sign off on a contract completion?

cubells:
This module completes these others:
https://github.com/OCA/project/tree/9.0/project_task_digitized_signature
https://github.com/OCA/stock-logistics-workflow/tree/9.0/stock_picking_digitized_signature
https://github.com/OCA/sale-workflow/pull/504

The purpose is that commercial user with a tablet or similar, in customer home, shows him a document (project, sale, contract or picking) and he signs document. Company can now continue commercial flow.

cubells:
Customer signature is shown in printed document too.

pedrobaeza:
@lasley not everyone wants to install website part (in v11 it's not needed, but it does on v10).

@rafaelbn you can try now

pedrobaeza (approved):

andhit-r (approved):

pedrobaeza:
@rafaelbn give your last try.

rafaelbn:
Runbot says `502 Bad Gateway` I cannot test it.

By the way @cubells @pedrobaeza `codecov/patch — 85.71% of diff hit (target 100%)`

woodbrettm:
@lasley I could probs just add this module as a dependency for website_sale_contract. Most of the backend logic in website_sale_contract lies in the checkout wizard, so should not be an issue.

@cubells would it be possible to add a `signature_name` field to account.analytic.account? I think it would be helpful to know the name of the person that signed the contract. Would also follow what is done in website_quote.

contract_digitized_signature/README.rst
```diff
@@ -0,0 +1,72 @@
...
+Installation
+============
+
+This module depends on :
+
+* web_widget_digitized_signature
+
+This module is part of the OCA/web suite.
+
+Configuration
```
chienandalu: Is this section needed. It's the same instructions as in Usage

contract_digitized_signature/README.rst
```diff
@@ -0,0 +1,72 @@
...
+=====
+
+#. Go to **Sales > Sales > Contracts** and edit or create a contract.
+#. You will see a box ready to write a signature by your customer.
+#. Save Contract.
+#. Every time you modify signature, a message is saved in chatter to log any
+   changes on it.
+#. You can clear signature in any moment by clicking on the **Clear** link
+   located on the top right side of signature box.
+#. If you print Contract, signature was added to contact report .
```
chienandalu: If you print the contract the signature will be shown in the contract report.

chienandalu (approved):
Tested on runbot. Just two non blocking comments

lasley:
@bmw95 - if you can easily make our module compatible with this strategy, please do it. I'd rather not have a fractured ecosystem

rafaelbn:
Please @cubells could you answer to https://github.com/OCA/contract/pull/102#issuecomment-335993638. Thanks!

rafaelbn (approved):
Tested functionaly in runbot 👍

cubells:
@bmw95 @lasley 
I added a new field signature_name

Can you retest?

contract_digitized_signature/models/account_analytic_account.py
```diff
@@ -12,6 +12,11 @@ class AccountAnalyticAccount(models.Model):
     customer_signature = fields.Binary(
         string='Customer acceptance',
     )
+    signature_name = fields.Many2one(
```
lasley: IMO this should be a plain text field, possibly containing a capture of the name from the partner (not sure how your workflow works). Linking a signature to a dynamic record means that the actual data can change out from underneath the legal data

pedrobaeza: Well, you shouldn't modify contacts when there are changes, but deactivate them and create new ones. The problem with the plain text is that requires to fully type it. With this, you can select directly by direct search.

lasley: @pedrobaeza - none of my customers do that, so while yeah it's the best option, it's still not a legal field.

An alternative would be to add a `signatory_id` column that is the relation, with an  onchange to update the `signature_name`.

Furthermore- this creates an incompatibility with the way Odoo does it, which is the way we did it. The goal of adding this was to make our addons live in harmony. Adding this, but making it a relation, takes us further from that goal than simply not having the column

lasley: Also - to be a relation, it needs to be suffixed with `_id`, which is not the case with `signature_name`

pedrobaeza: OK, let's change it to a char field. At the end, we don't have that requirement.

cubells: done!

contract_digitized_signature/models/account_analytic_account.py
```diff
@@ -0,0 +1,30 @@
...
+    )
+    signature_name = fields.Char(
+        string='Signed by',
+    )
+
+    @api.model
+    def create(self, values):
+        contract = super(AccountAnalyticAccount, self).create(values)
+        if contract.customer_signature:
+            values = {'customer_signature': contract.customer_signature}
```
lasley, line 23: Can you add a test for this please?

lasley (approved):
Thanks @cubells. One note, non-functional change

contract_digitized_signature/models/account_analytic_account.py
cubells: @lasley done!

lasley:
Thanks @cubells

PR #99, opened 2017-09-20
[FIX] Contracts Management - Recurring: wrong value for account

primes2h:
- Added missing brackets to fix #59 

Thanks @carlosdauden

PR #103, opened 2017-10-10
[FIX] website_portal_contract: Add analytic_id to demo data

lasley:
* Add analytic_id reference to the invoice line data to resolve null column

Ref https://github.com/OCA/contract/pull/99#issuecomment-335532888


website_portal_contract/demo/account_analytic_account_demo.xml
```diff
@@ -13,8 +13,6 @@
         <field name="partner_id" ref="portal.demo_user0_res_partner" />
         <field name="recurring_invoices" eval="True" />
         <field name="website_template_id" ref="website_contract_template_default" />
-        <field name="recurring_invoice_line_ids"
```
pedrobaeza, line 16: Do you remove this?

lasley: Yup, [it's a O2M](https://github.com/OCA/contract/blob/10.0/contract/models/account_analytic_account.py#L26) so it is set automatically when the line is set to the analytic account.

pedrobaeza: Yeah, totally! It's important then the order of the XML loading in the manifest. Note that in it for avoiding problems changing that order.

lasley: Added a note in the manifest about the load order. 

Note that I didn't increase the version of the module because this is all just demo data issue.

pedrobaeza:
Travis is now red on OCB. I have restarted the job to see if it's something random.

lasley:
~~Oh no I hope we're not about to have cascading issues in MQT in v11~~

I'm dumb, wrong PR

pedrobaeza:
Do you know the problem here then?

lasley:
Honestly I don't. The issue is ping ponging back and forth between Odoo and OCB too.

So yeah my solution here is remove the broken test. Pretty sure it's implicitly covered, and the coverage shouldn't drop.

lasley:
Alright I'm going to merge this in the interest of fixing the build. The only changes outside of the test removal were already approved

PR #105, opened 2017-10-13
[IMP] contract: Improvements in tests

pedrobaeza:
* Use SavepointCase for making the setup only once for all tests.
* Make them inheritable, creating a base class with only the setup, so that it can be inherited without the need of executing all tests contained here each time you inherit it, and adding other class in the same module that inherits from the base class that actually performs the tests.
* Remove duplicated test method.


pedrobaeza:
Example of a module that benefits from this inheritable system: https://github.com/OCA/vertical-association/pull/39

Unfortunately, this is red due until #103 is fixed.

pedrobaeza:
@lasley please fix the branch and fast-track this when done

lasley:
Rebased - will fast track if green

lasley (approved):

PR #109, opened 2017-10-26
[10.0][ADD][WIP] Recurring Rule Type by Line

angelmoya:
Contracts Recurring Rule by Line
================================

This module extends functionality of contracts to be able to set diferents recurring
rule type by lines.

Usage
=====

To use this module, you need to:

#. Go to Accounting -> Contracts and select or create a new contract.
#. Check *Generate recurring invoices automatically*.
#. Now you can set recurring rule type by line.


angelmoya:
Module is working... but could be improbed...

Todo list:

- [ ] Create test
- [ ] Check default values
- [ ] Default value in lines from contract data for recurring type

angelmoya:
Also checked compatible with contract sale generate module

pedrobaeza:

PR #108, opened 2017-10-25
[10.0][MIG] contract_digitized_signature

woodbrettm:
Hi there,

Just doing a quick migration so this module can be used for website_sale_contract.

Thanks!
Brett

@lasley @rafaelbn

PR #112, opened 2017-10-30
10.0 mig contract digitized signature

lucode:
Hope this one is correct?


pedrobaeza:
Agh, sorry to say that someone else already migrated the module in #108. I didn't see it before. That one should be reviewed instead.

PR #115, opened 2017-11-06
[10.0][FIX] Fix Travis

pedrobaeza:
Trying to fix Travis result

PR #116, opened 2017-11-10
[FIX] contract_payment_auto, contract_payment_mode, contract_sale_generation, contract_variable_quantity: Fix wkhtmltopdf freeze

moylop260:
The test uses wkhtmltopdf, wkhtmltopdf uses html connection to get data from odoo.
If you have a instance starting and you try use it you will have a freezed browse, with wkhtmltopdf is the same.
So is required run this tests after start odoo and use a session, the session is created from HttpCase.SetUp method.

Fix https://github.com/OCA/maintainer-quality-tools/issues/509

In your local machine you can bypass this issue if you have a session stored (a previous /web/login before of run the tests), that is the reason that with the CI fails because anyone start a session or login instance before run the tests.

NOTE: Please, delete the branch if is merged
NOTE2: Please, merge with squash
NOTE3: There is a unrelated issue after run the tests correctly: https://github.com/OCA/contract/issues/117

### DIFF
--- a/contract/models/account_analytic_account.py
+++ b/contract/models/account_analytic_account.py
@@ -209,7 +209,10 @@ class AccountAnalyticAccount(models.Model):
         Create invoices from contracts
         :return: invoices created
         """
-        invoices = self.env['account.invoice']
+        # NOTE: Don't use a browse because here there are 2 types of records:
+        #       sale or invoice. A manual 'set' fix error of
+        #       Mixing apples and oranges: account.invoice().union(sale.order)
+        invoices = set()
         for contract in self:
             ref_date = contract.recurring_next_date or fields.Date.today()
             if (contract.date_start > ref_date or
@@ -228,7 +231,7 @@ class AccountAnalyticAccount(models.Model):
                 'force_company': contract.company_id.id,
             })
             # Re-read contract with correct company
-            invoices |= contract.with_context(ctx)._create_invoice()
+            invoices.add(contract.with_context(ctx)._create_invoice())
             contract.write({
                 'recurring_next_date': new_date.strftime('%Y-%m-%d')
             })
--- a/contract/models/account_analytic_account.py
+++ b/contract/models/account_analytic_account.py
@@ -206,7 +206,7 @@ class AccountAnalyticAccount(models.Model):
     @api.multi
     def recurring_create_invoice(self):
         """
-        Create invoices from contracts
+        Create invoices or sales from contracts
         :return: invoices created
         """
         # NOTE: Don't use a browse because here there are 2 types of records:
--- a/contract_payment_auto/tests/test_account_analytic_account.py
+++ b/contract_payment_auto/tests/test_account_analytic_account.py
@@ -6,14 +6,16 @@ import mock
 
 from contextlib import contextmanager
 
+import odoo.tests
 from odoo import fields
 from odoo.tools import mute_logger
-from odoo.tests.common import TransactionCase
 
 from ..models import account_analytic_account
 
 
-class TestAccountAnalyticAccount(TransactionCase):
+@odoo.tests.at_install(False)
+@odoo.tests.post_install(True)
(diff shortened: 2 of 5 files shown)


contract/models/account_analytic_account.py
```diff
@@ -206,10 +206,13 @@ def _create_invoice(self):
...
         """
-        Create invoices from contracts
+        Create invoices or sales from contracts
         :return: invoices created
         """
-        invoices = self.env['account.invoice']
+        # NOTE: Don't use a browse because here there are 2 types of records:
+        #       sale or invoice. A manual 'set' fix error of
+        #       Mixing apples and oranges: account.invoice().union(sale.order)
+        invoices = set()
```
pedrobaeza, line 215: This is very ugly because is caused by an inherited module. @angelmoya you have provoked here a real inconsistency with your module contract_sale_generation. You can't change the API for returning either sale orders or invoices. You MUST change your module for not returning the other things. There are other modules that depend on the API that says that it returns a recordset of invoices, not other thing. Please fix it, Ángel.

contract_payment_auto/tests/test_account_analytic_account.py
```diff
@@ -6,14 +6,16 @@
...
+import odoo.tests
 from odoo import fields
 from odoo.tools import mute_logger
-from odoo.tests.common import TransactionCase
 
 from ..models import account_analytic_account
 
 
-class TestAccountAnalyticAccount(TransactionCase):
+@odoo.tests.at_install(False)
```
pedrobaeza, line 16: Can we make a Wiki about these cases?

contract_payment_mode/tests/test_contract_payment.py
```diff
@@ -4,42 +4,44 @@
...
 from ..hooks import post_init_hook
 
 
-class TestContractPaymentInit(common.SavepointCase):
-    @classmethod
-    def setUpClass(cls):
-        super(TestContractPaymentInit, cls).setUpClass()
+@odoo.tests.post_install(True)
+@odoo.tests.at_install(False)
+class TestContractPaymentInit(odoo.tests.HttpCase):
```
pedrobaeza, line 13: I don't want to change to HttpCase, because it executes `setUp` for each test. I want `SavepointCase`

contract_variable_quantity/tests/test_contract_variable_quantity.py
```diff
@@ -2,27 +2,28 @@
...
 
 
-class TestContractVariableQuantity(common.SavepointCase):
-    @classmethod
-    def setUpClass(cls):
-        super(TestContractVariableQuantity, cls).setUpClass()
-        cls.partner = cls.env['res.partner'].create({
+@odoo.tests.at_install(False)
+@odoo.tests.post_install(True)
+class TestContractVariableQuantity(odoo.tests.HttpCase):
```
pedrobaeza, line 11: The same: it should be `SavepointCase`

contract/models/account_analytic_account.py
angelmoya: OK, I will check

contract_payment_mode/tests/test_contract_payment.py
moylop260: I’m agree with you
My focus here is to get a travis green in order to search the real guilty

check my last comment from

https://github.com/OCA/maintainer-quality-tools/issues/509#issuecomment-343481818

contract/models/account_analytic_account.py
angelmoya: I'm checking.... I tried to use contract code, and not rewrite all functionality... 
But finally if sale_generation is not part of base module and other modules adapt to it, then would be better to create as independent functionality isolating invoicing and sales generation.

angelmoya: #118

pedrobaeza:
Closing in favor of #115

PR #120, opened 2017-11-12
[10.0][FIX] Correct domain attribute in field journal_id

gorozcoh:
Original domain includes unknown value company_id. Throws error when selecting the journal.

### DIFF
--- a/contract/models/account_analytic_contract.py
+++ b/contract/models/account_analytic_contract.py
@@ -58,16 +58,17 @@ class AccountAnalyticContract(models.Model):
         'account.journal',
         string='Journal',
         default=lambda s: s._default_journal(),
-        domain=lambda s: s._domain_journal(),
+        domain="[('type', '=', 'sale'),('company_id', '=', company_id)]",
     )
-
-    @api.model
-    def _domain_journal(self):
-        company_id = self.env.context.get(
-            'company_id', self.env.user.company_id.id)
-        return [('type', '=', 'sale'), ('company_id', '=', company_id)]
+    company_id = fields.Many2one('res.company', string='Company', required=True,
+                                 default=lambda self: self.env.user.company_id)
 
     @api.model
     def _default_journal(self):
-        return self.env['account.journal'].search(self._domain_journal(), limit=1)
+        company_id = self.env.context.get(
+            'company_id', self.env.user.company_id.id)
+        domain = [
+            ('type', '=', 'sale'),
+            ('company_id', '=', company_id)]
+        return self.env['account.journal'].search(domain, limit=1)
 
--- a/contract/models/account_analytic_contract.py
+++ b/contract/models/account_analytic_contract.py
@@ -16,6 +16,11 @@ class AccountAnalyticContract(models.Model):
     NO_SYNC = [
         'name',
     ]
+    
+    def domain_journal_id(self):
+        company_id = self.env.context.get(
+            'company_id', self.env.user.company_id.id)
+        return [('type', '=', 'sale'), ('company_id', '=', company_id)]
 
     name = fields.Char(
         required=True,
@@ -58,7 +63,7 @@ class AccountAnalyticContract(models.Model):
         'account.journal',
         string='Journal',
         default=lambda s: s._default_journal(),
-        domain="[('type', '=', 'sale'),('company_id', '=', company_id)]",
+        domain=lambda self:self.domain_journal_id(),
     )
 
     @api.model
--- a/contract/models/account_analytic_contract.py
+++ b/contract/models/account_analytic_contract.py
@@ -60,8 +60,12 @@ class AccountAnalyticContract(models.Model):
         default=lambda s: s._default_journal(),
         domain="[('type', '=', 'sale'),('company_id', '=', company_id)]",
     )
-    company_id = fields.Many2one('res.company', string='Company', required=True,
-                                 default=lambda self: self.env.user.company_id)
+    company_id = fields.Many2one(
+        'res.company',
+        string='Company',
+        required=True,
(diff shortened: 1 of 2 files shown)


pedrobaeza:
The expression is correct as it was in principle. Please put steps to reproduce a problem.

lasley:
Not at a computer to test, but I’m having trouble recalling where the company_id field is on the form. Issue is probably that the field needs to be added to contract form as invisible. 


-Dave Lasley

gorozcoh:
Steps to reproduce:

1. Clean instance without demo data.
2. Install OCA/contract modules
3. Open view contract template and clic on create button
4. Try select journal
5. Client throws error related to unknown field company_id

pedrobaeza:
OK, then as said your PR should be one adding to the view the field company_id (which should have groups attribute with multi-company group assigned).

gorozcoh:
View account_analytic_contract_view_form is based in model account.analytic.contract and in this model there is not company_id field. I tried this before this PR and don't works. default_journal and domain_journal use the same domain, why not DRY:

```
    @api.model
    def _domain_journal(self):
        company_id = self.env.context.get(
            'company_id', self.env.user.company_id.id)
        return [('type', '=', 'sale'), ('company_id', '=', company_id)]

    @api.model
    def _default_journal(self):
        return self.env['account.journal'].search(self._domain_journal(), limit=1)
```

lasley:
IMO `account.analytic.contract` should define `company_id`.

gorozcoh:
The CLA was sent. The company_id field was added.

carlosdauden (approved):

pedrobaeza:
But including a `company_id` field should be linked to multi-company rules, isn't it? Shouldn't we do an `analytic_multi_company` module with that? 

@gorozcoh sorry for catching you with all this discussion in the middle, but I think we should do this the correct way

lasley:
We can slap multi-company on top of a pre-existing `company_id` col. IMO that's the best way to do it here in order to keep the scope of this module limited to providing contracts on top of the pre-existing `account.analytic.account` system. Multi company analytic accounting is out of scope here, both in the sense that it may be undesired for contracts but also the fact that contracts might be undesired for multi company analytic accounting.

For multi-company though, we would also need a `contract_multi_company` on top of the `analytic_multi_company` I think.

lasley (approved):
Thanks @gorozcoh - one comment inline. Not a blocker, just an idiom

contract/models/account_analytic_contract.py
```diff
@@ -60,6 +60,8 @@ class AccountAnalyticContract(models.Model):
         default=lambda s: s._default_journal(),
         domain="[('type', '=', 'sale'),('company_id', '=', company_id)]",
     )
+    company_id = fields.Many2one('res.company', string='Company', required=True,
```
lasley: Please match the style of the pre-existing fields, with one argument on each line and a trailing comma on the last one.

This style is primarily due to VCS (git) reasons - in order to change or add lines going forward, the diff is not as large & will only touch parts that need to be edited.

gorozcoh: Thanks lasley. I got it.

lasley (approved):
Thanks @gorozcoh 

@pedrobaeza - do you agree with my evaluation about multi-company being out of scope here? Or maybe I'm missing something?

pedrobaeza:
@lasley this still misses the multi-company rules, so I agree on having contract_multi_company (or include in `contract` module all the multi-company stuff together), but not doing things half.

lasley:
In this case though, account.analytic.account already has a company_id field, I just did not move it back into the contract model when I created it. The current operation of the module is purely on single company rules, so I don’t see the harm in continuing that motivation in a bug fix. Adding multi company to the pre-existing module is a full version upgrade, but I personally think the better option is a standalone module adding it. Not all deploys need the extra overhead of multi company.

pedrobaeza:
OK, go ahead with this PR, but multi-company part shouldn't be in a module apart, as Odoo is multi-company aware, so it shouldn't be something that you make apart.

lasley:
Thanks for the fix @gorozcoh 

@pedrobaeza - doesn't make much of a difference to me because I would probably use the multi-company and thus would install the additional module anyways.

Ok so in the instance of an ideal situation - how does a multi company contract even work? Or are you thinking just the template side of things, so that multiple companies can use one contract template, but the contracts themselves (aka account.analytic.account) are between one company and one partner still?

rafaelbn:
Hello @gorozcoh 😄 

You introduce this issue https://github.com/OCA/contract/issues/125 with this PR. Please could you review it? Thanks.

Before merging please test it at least creating a contract 😉 , our customer cannot create contracts now! 😮 

Thank you!

cc @carlosdauden

rafaelbn:
If any doubt... i test it in all last runbots available. Only fails in this PR

gorozcoh:
What is the use of contract_test.py if everything has to be tested manually? Adding a new field company_id was not my idea. Did you check my original PR?

lasley:
@rafaelbn - IMO this was on the reviewers and not @gorozcoh. I admittedly did not functionally review the change, and it sounds like ~Pedro~ Carlos didn't either. This is a QA process breakdown, and while yeah the developer could have caught it, that's the point of a review process. I'll handle this fix today.

@gorozcoh - In an ideal world, the tests would catch everything. This is why tests should be added as the module is built, and not after the fact. This is also why we strive for 100% test coverage in all modules, which is not the case here. In the case of our automated tests, testing is started from the template level. In @rafaelbn case, there is no template being created so the workflow is different. Cutting corners ftw basically

_Edit: fixed other reviewer name; it wasn't Pedro_

PR #122, opened 2017-11-16
[10.0][FIX] contract: propagate contract reference to invoices

cubells:
As It's done with sales and purchase orders, now contract's reference is propagated to invoices if is setted.

cc @Tecnativa

### DIFF
--- a/contract/models/account_analytic_account.py
+++ b/contract/models/account_analytic_account.py
@@ -232,14 +232,14 @@ class AccountAnalyticAccount(models.Model):
             self.company_id.currency_id
         )
         invoice = self.env['account.invoice'].new({
-            'reference': self.code,
+            'reference': self.name,
             'type': 'out_invoice',
             'partner_id': self.partner_id.address_get(
                 ['invoice'])['invoice'],
             'currency_id': currency.id,
             'journal_id': journal.id,
             'date_invoice': self.recurring_next_date,
-            'origin': self.name,
+            'origin': self.code,
             'company_id': self.company_id.id,
             'contract_id': self.id,
             'user_id': self.partner_id.user_id.id,
--- a/contract/__manifest__.py
+++ b/contract/__manifest__.py
@@ -9,7 +9,7 @@
 
 {
     'name': 'Contracts Management - Recurring',
-    'version': '10.0.4.1.1',
+    'version': '10.0.4.1.2',
     'category': 'Contract Management',
     'license': 'AGPL-3',
     'author': "OpenERP SA, "


contract/models/account_analytic_account.py
```diff
@@ -183,7 +183,7 @@ def _prepare_invoice(self):
             'currency_id': currency.id,
             'journal_id': journal.id,
             'date_invoice': self.recurring_next_date,
-            'origin': self.name,
+            'origin': self.code or self.name,
```
carlosdauden: Code is stored as reference
https://github.com/OCA/contract/pull/122/files#diff-47b79491db78ebebcb372014ce9c7bf3L179

If you want change this field, Don't you should change both?

carlosdauden (changes requested):

rafaelbn:
Hi @cubells ,

- This [field ](https://github.com/OCA/contract/pull/122/files#diff-47b79491db78ebebcb372014ce9c7bf3L179)`reference` in invoices is only for vendors so we should delete it here in contracts

- The field `name`in contracts should be in field `name` in the invoices
- the field `code` in contracts should be in field `origin`in the invoices

Check screenshot:

rafaelbn (changes requested):
Please @cubells review https://github.com/OCA/contract/pull/122#issuecomment-345319525

thanks

carlosdauden (commented):
Another field to consider is name in invoice model.
We can set invoice.name=contract.name.

You must also bear in mind that the only field that appears in the invoice tree view is origin and that the contracts haven't an automatic sequence that fills the code field.

rafaelbn:
Ping @cubells please

pedrobaeza:
@rafaelbn there's no agreement yet on this. For me, each data goes to the right place right now.

rafaelbn:
@pedrobaeza @carlosdauden @cubells do we agree in:

- This [field ](https://github.com/OCA/contract/pull/122/files#diff-47b79491db78ebebcb372014ce9c7bf3L179) reference in invoices is only for vendors so we should delete it here in contracts. In fact this field is hidden in customer invoices.... why should contract use it?

This is obvious. Isn't it?

rafaelbn:
In my opinion @carlosdauden @pedrobaeza @cubells :

- the name of the contract is **private** so customer must not know about it. Obviously some companies could want to so it (like us)
- the reference is that a reference for the customer that you use to reference you contract. Usualy the field origin is filled with the reference of the object which create it. This is the more general approach.

Actual situation in contract without this PR (absolutely with should fix https://github.com/OCA/contract/pull/122#issuecomment-346178836)



My proposal is https://github.com/OCA/contract/pull/122#issuecomment-345319525 . It doesn't have to be the best one... or yes.

The second approach that for me is good to:

- if contract has reference: write reference in origin of the invoice (only reference, reference + name I don't agree)
- If contract doesn't have reference: then write contract name

This second approach cover my proposal + our own use case 👍 🎉 

I will ask my pillow 😄

rafaelbn:
Hi,

I didn't get any answer but I will introduce more reasoning plus https://github.com/OCA/contract/pull/122#issuecomment-345319525 plus this **BUG** (https://github.com/OCA/contract/pull/122#issuecomment-345319525) plus https://github.com/OCA/contract/pull/122#issuecomment-346184891

- Contracts module comes from v8
- in v8 the reference of the contract go to origin in invoice
- If we change this in v9 and v10 and v11 when the customer come to a new version thay will have change the functionality which is logical. Must this customers from v8 make a custom module to work as they were working? or must a new customer which begin in v10 or v11 to work os module work or make a custom or it?

pedrobaeza:
What is the final call with this? Any way, please rebase on current 10.0 that is green

contract/models/account_analytic_account.py
```diff
@@ -232,14 +232,14 @@ def _prepare_invoice(self):
...
-            'reference': self.code,
+            'reference': self.name,
             'type': 'out_invoice',
             'partner_id': self.partner_id.address_get(
                 ['invoice'])['invoice'],
             'currency_id': currency.id,
             'journal_id': journal.id,
             'date_invoice': self.recurring_next_date,
-            'origin': self.name,
+            'origin': self.code,
```
rafaelbn, line 242: Please! This was correct, set  `'origin': self.name,` 
I still have this BUG from 1 year ago

contract/models/account_analytic_account.py
```diff
@@ -232,14 +232,14 @@ def _prepare_invoice(self):
             self.company_id.currency_id
         )
         invoice = self.env['account.invoice'].new({
-            'reference': self.code,
+            'reference': self.name,
```
rafaelbn, line 235: The correct is : `'name': self.code,`
Please change it , rebase and get this merge

rafaelbn: The field reference is an error from the beginning

rafaelbn (changes requested):
Pleaseeee pleaseee pleaseee :smile:

AaronHForgeFlow:
It seems this need a rebase @cubells

rafaelbn:
I spend here time to explain what is the more general approach but closing as blocked. :cry:

PR #128, opened 2017-11-30
[11.0][FIX] Set encoding for *.py files

katyukha:
In some python files copyright-sign is used, and if encodning is not set, module cannot be imported on some platforms.


pedrobaeza:
That's on purpose, as Python 3 doesn't need encoding. Closing as useless.

Issue #124, opened 2017-11-23
[10.0] Overriding field analytic_account_id in account.analytic.contract.line breaks link between account.analytic.invoice.line and account.analytic.account.


NL66278:
I have a module where I need to have partner_id as a reference field in account.analytic.invoice.line. This should not pose any problem, as account.analytic.invoice.line refers with account_analytic_id to account.analytic.account, which has partner id.

This is my code:
```python
  1 # -*- coding: utf-8 -*-                                                        
  2 # Copyright 2014-2017 Therp BV <https://therp.nl>.                             
  3 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).                
  4 from odoo import api, fields, models                                           
  5                                                                                
  6                                                                                
  7 class AccountAnalyticInvoiceLine(models.Model):                                
  8     _inherit = 'account.analytic.invoice.line'                                 
  9                                                                                
...                                                         
 23     partner_id = fields.Many2one(                                              
 24         comodel_name='res.partner',                                            
 25         related='analytic_account_id.partner_id',                              
 26         string='Partner') 
```

Unfortunately the code in contract/models/account_analytic_contract_line.py breaks the original analytic_account_id:
```python
  1 # -*- coding: utf-8 -*-                                                        
  2 # Copyright 2017 LasLabs Inc.                                                  
  3 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).                
  4                                                                                
  5 from odoo import fields, models                                                
  6                                                                                
  7                                                                                
  8 class AccountAnalyticContractLine(models.Model):                               
  9                                                                                
 10     _name = 'account.analytic.contract.line'                                   
 11     _description = 'Contract Lines'                                            
 12     _inherit = 'account.analytic.invoice.line'                                 
 13                                                                                
 14     analytic_account_id = fields.Many2one(                                     
 15         string='Contract',                                                     
 16         comodel_name='account.analytic.contract',                              
 17         required=True,                                                         
 18         ondelete='cascade',                                                    
 19     )
```
The result is this:
```
2017-11-23 10:41:23,674 13472 DEBUG odownbdem10 odoo.modules.loading: loading module publication (55/57)
2017-11-23 10:41:23,721 13472 ERROR odownbdem10 odoo.fields: Problem setting up related field partner_id for model account.analytic.contract.line
2017-11-23 10:41:23,721 13472 ERROR odownbdem10 odoo.modules.registry: Error setting up fields for model account.analytic.contract.line
[...]

PR #129, opened 2017-11-30
[10.0][IMP] New module contract_line_extended.

NL66278:
Restores link between contract line (account.analytic.invoice.line) and contract
(account.analytic.account).
Also adds reference fields partner_id, date_start and date_end to contract line.

### DIFF
--- a/contract_line_extended/models/account_analytic_invoice_line.py
+++ b/contract_line_extended/models/account_analytic_invoice_line.py
@@ -79,7 +79,7 @@ class AccountAnalyticInvoiceLine(models.Model):
         maintained and dates stay within limit of contract."""
         vals['contract_id'] = vals['analytic_account_id']
         result = super(AccountAnalyticInvoiceLine, self).create(vals)
-        new_line._limit_dates()
+        result._limit_dates()
         return result
 
     @api.multi
--- a/contract_line_extended/models/account_analytic_invoice_line.py
+++ b/contract_line_extended/models/account_analytic_invoice_line.py
@@ -1,0 +1,70 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Therp BV.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from odoo import api, fields, models
+
+
+SQL_SYNCHRONIZE1 = \
+    """UPDATE account_analytic_invoice_line
+ SET contract_id = analytic_account_id
+ WHERE contract_id IS NULL"""
+
+SQL_SYNCHRONIZE2 = \
+    """UPDATE account_analytic_invoice_line
+SET partner_id=subquery.partner_id,
+    date_start=subquery.date_start,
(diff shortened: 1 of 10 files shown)


pedrobaeza:
Please rebase on current 10.0 that is green

NL66278:
@pedrobaeza Rebase has been done. As well as commit squash. @lasley Everything is now green on travis and runbot.

contract_line_extended/models/account_analytic_invoice_line.py
```diff
@@ -0,0 +1,70 @@
...
+ WHERE subquery.id = account_analytic_invoice_line.id"""
+
+
+class AccountAnalyticInvoiceLine(models.Model):
+    _inherit = 'account.analytic.invoice.line'
+
+    # We already had analytic_account_id to refer to the contract to which
+    # this line belongs. Unfortunately this is overwritten by
+    # account.analytic.contract.line which (ab)uses the same field to refer
+    # to the contract template.
```
lasley, line 39: Hmmm this feels to me like I broke something with the new implementation while being lazy. Am I understanding correctly that after turning a contract template (`account.analytic.contract`) into an active contract (`account.analytic.account`), the `account.analytic.account.line`s that are created maintain a reference ID corresponding to the `account.analytic.contract.line` instead of being rewritten correctly? If so this, was not my intent and seems like an extension of #80

lasley:
@NL66278 - I looked deeper into this after reading #124, and I get it now. I didn't consider a related field with this implementation, but this makes sense in hindsight.

IMO it would be better to fix this on the base module though- Otherwise we're just designing around my stupidity.

NL66278:
@lasley Hi, something definitely broke. Well mistakes happen sometime, I have been guilty of a lot of them. I agree it would be best to solve it in the main module. The question is what the best approach would be. Personally I would prefer to merge contracts and templates. Then a template (which could have a boolean field marking it as a template), would have start date, partner and things like that optional. If it is preferred to keep contracts and templates separate, I think the inheritance should go from templates and template lines to contract and contract lines. And finally we might do something about the fact that templates are called contracts in the model name, which is mighty confusing.

lasley:
I agree it makes sense to combine the two. I think we could actually skip the boolean field and just consider a contract as a template if it does not have a partner defined on it. We don't want a partner in the templates for the most part, because it's the one getting sold to multiple partners. This would work out too, because then we'd gain the reporting of having the templates as a parent analytic account.

I think it makes sense to keep the explicit contract model, and try to pull as much of the contract logic out of the analytic account as possible. The mixing of that logic has always been kind of dirty in my mind anyways, and is what I kind of half designed around with the template. Hindsight is 20/20 🚀

NL66278:
I close this PR in favour of this one, that is a direct backport of what has already been merged in 11.0: https://github.com/OCA/contract/pull/172. Any extra things done in this PR might be part of an extra module.

PR #130, opened 2017-12-02
[10.0][FIX+IMP] contract: Improve usability and don't fail on wrong data

pedrobaeza:
* Cron create invoices masked for avoiding silent errors
* New constraints for assuring data consistency
* UI helps for entering consistent data
* Spanish translation
* Remove double company_id field on form

@Tecnativa

### DIFF
--- a/contract/models/account_analytic_account.py
+++ b/contract/models/account_analytic_account.py
@@ -51,6 +51,17 @@ class AccountAnalyticAccount(models.Model):
         index=True,
         default=lambda self: self.env.user,
     )
+    create_invoice_visibility = fields.Boolean(
+        compute='_compute_create_invoice_visibility',
+    )
+
+    @api.depends('recurring_next_date', 'date_end')
+    def _compute_create_invoice_visibility(self):
+        for contract in self:
+            contract.create_invoice_visibility = (
+                not contract.date_end or
+                contract.recurring_next_date <= contract.date_end
+            )
 
     @api.onchange('contract_template_id')
     def _onchange_contract_template_id(self):
@@ -75,14 +86,59 @@ class AccountAnalyticAccount(models.Model):
             )):
                 self[field_name] = self.contract_template_id[field_name]
 
-    @api.onchange('recurring_invoices')
-    def _onchange_recurring_invoices(self):
-        if self.date_start and self.recurring_invoices:
+    @api.onchange('date_start')
+    def _onchange_date_start(self):
+        if self.date_start:
             self.recurring_next_date = self.date_start
 
     @api.onchange('partner_id')
     def _onchange_partner_id(self):
         self.pricelist_id = self.partner_id.property_product_pricelist.id
+
+    @api.constrains('partner_id', 'recurring_invoices')
+    def _check_partner_id_recurring_invoices(self):
+        for contract in self.filtered('recurring_invoices'):
+            if not contract.partner_id:
+                raise ValidationError(
+                    _("You must supply a customer for the contract '%s'") %
+                    contract.name
+                )
+
+    @api.constrains('recurring_next_date', 'date_start')
+    def _check_recurring_next_date_start_date(self):
+        for contract in self.filtered('recurring_next_date'):
+            if contract.date_start > contract.recurring_next_date:
+                raise ValidationError(
+                    _("You can't have a next invoicing date before the start "
+                      "of the contract '%s'") % contract.name
+                )
+
+    @api.constrains('recurring_next_date', 'recurring_invoices')
+    def _check_recurring_next_date_recurring_invoices(self):
+        for contract in self.filtered('recurring_invoices'):
+            if not contract.recurring_next_date:
+                raise ValidationError(
+                    _("You must supply a next invoicing date for contract "
+                      "'%s'") % contract.name
+                )
+
+    @api.constrains('date_start', 'recurring_invoices')
+    def _check_date_start_recurring_invoices(self):
+        for contract in self.filtered('recurring_invoices'):
+            if not contract.date_start:
+                raise ValidationError(
+                    _("You must supply a start date for contract '%s'") %
+                    contract.name
+                )
+
+    @api.constrains('date_start', 'date_end')
+    def _check_start_end_dates(self):
+        for contract in self.filtered('date_end'):
+            if contract.date_start > contract.date_end:
+                raise ValidationError(
+                    _("Contract '%s' start date can't be later than end date")
+                    % contract.name
+                )
 
     @api.multi
     def _convert_contract_lines(self, contract):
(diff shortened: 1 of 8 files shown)


pedrobaeza:
Let's see if this one is the definitive one for total satisfaction, hehe

contract/models/account_analytic_account.py
```diff
@@ -75,15 +85,60 @@ def _onchange_contract_template_id(self):
             )):
                 self[field_name] = self.contract_template_id[field_name]
 
-    @api.onchange('recurring_invoices')
-    def _onchange_recurring_invoices(self):
-        if self.date_start and self.recurring_invoices:
+    @api.onchange('date_start')
+    def _onchange_date_start(self):
```
NL66278, line 90: Question, what is the reasoning for not testing on recurring_invoices?

NL66278 (approved):
I have one question, but this seems a whole set of very welcome improvements to me!

contract/models/account_analytic_account.py
```diff
@@ -75,15 +85,60 @@ def _onchange_contract_template_id(self):
...
+                    _("You can't have a next invoicing date before the start "
+                      "of the contract '%s'") % contract.name
+                )
+
+    @api.constrains('recurring_next_date', 'recurring_invoices')
+    def _check_recurring_next_date_recurring_invoices(self):
+        for contract in self.filtered('recurring_invoices'):
+            if not contract.recurring_next_date:
+                raise ValidationError(
+                    _("You must supply an invoicing next date for contract "
```
lasley: s/an invoicing next date/a next invoicing date/

contract/models/account_analytic_account.py
```diff
@@ -75,15 +85,60 @@ def _onchange_contract_template_id(self):
...
+                      "'%s'") % contract.name
+                )
+
+    @api.constrains('date_start', 'recurring_invoices')
+    def _check_date_start_recurring_invoices(self):
+        for contract in self.filtered('recurring_invoices'):
+            if not contract.date_start:
+                raise ValidationError(
+                    _("You must supply an start date for contract '%s'") %
```
lasley: s/an/a

lasley (approved):
Thanks @pedrobaeza. Few minor verbiage changes, but otherwise solid. I do have the same question as @NL66278 though.

lasley:
Oh and I guess the most important part - Travis is 🔴

contract/models/account_analytic_account.py
pedrobaeza: With this, now you don't need to care about that field: the recurring_next_invoice will always be synchronized.

pedrobaeza:
Done!

contract/models/account_analytic_account.py
lasley: It looks like we're also implicitly assuming that it's a recurring contract in the event of a `start_date` (IMO a great simplification). With that though, could we then just outright remove the `recurring_next_invoice` field?

lasley: Sorry `recurring_invoices` field, not `recurring_next_invoice`

rafaelbn (approved):
Tested functionally in runbot 👍 

Just check before merging if you want to improve codecov/patch — 89.65% of diff hit (target 94.61%)

contract/models/account_analytic_account.py
pedrobaeza: Previous code assumes the same: it took `date_start` when you mark `recurring_invoices` field. What you are talking about is to take `today` when marking `recurring_invoices`. It can be another thing to add indeed.

What I don't see correct is to remove `recurring_invoices` field, as it serves a lot for UI usability and user experience. Having that check, you clean the interface out of fields that aren't important if you don't use recurring invoicing. Note that in this PR, I have improved also the experience automarking by default the field when clicking on create when you are in the "Contracts" menu.

pedrobaeza:
@rafaelbn coverage improved

contract/models/account_analytic_account.py
```diff
@@ -267,7 +268,7 @@ def recurring_create_invoice(self):
             ref_date = contract.recurring_next_date or fields.Date.today()
             if (contract.date_start > ref_date or
                     contract.date_end and contract.date_end < ref_date):
-                if self.env.context.get('cron'):
+                if self.env.context.get('cron'):  # pragma: no cover
```
lasley: IMO this is cheating & should be removed. We should either take the coverage hit if we don't want to test it, or add the test (I'm good with either)

pedrobaeza: No, this is not cheating in this case. It's a case that doesn't modify anything and I don't want to be reported as a coverage failure.

lasley: There is still logic here, that if untested should yield a lack of coverage. Whether it modifies anything or not, the logic is still in the program and still does something different from the other things. 

I'm not arguing against the lack of testing for this, as I agree it's a minimal point and probably doesn't need it. I am arguing against the false coverage report though, because this is not covered and is a unique logic path that should be tested per TDD standards.

pedrobaeza: Dave, not for this case. There's no possibility of leading to that situation right now with current constraints, so it's only a safeguard for already installed DBs, but no sense of testing that specific case (forcing through SQL) or to report as not covered.

lasley: Untouchable code is still untested IMO. What's the harm in having it show what it is? Hiding the fact that the block of code wasn't executed during testing is just sweeping the fact it can't be touched under the rug. There's always going to be scenarios like this, which is why 100% test coverage is an unrealistic goal in the practical sense.

I won't block for it - but I really feel that `pragma: no cover` should align explicitly with the example case provided in Coverage.py docs, which is to exclude debug code. Any production code that's not actually tested should show as untested.

contract/models/account_analytic_account.py
lasley: Thanks for the elaboration Pedro. Does it make sense to actually store `recurring_invoices`, given that it is really just a functional field for triggering of specific workflows?

Just spitballing other upgrades here seeing as we have these nice improvements in similar areas. Not blocking

pedrobaeza: Well, I prefer to keep it right now, or we would need to change a lot of filters, context values...

contract/models/account_analytic_account.py
pedrobaeza: OK, I give up due to your insistence...

lasley:
Thanks for removing the pragma Pedro! Runbot red is the full disk and unrelated to this, so I merge

pedrobaeza:
Cherry-picked to v11 at 8d10f68610ee593acd7b7a16734f78e54b7060bb

PR #132, opened 2017-12-15
[11.0][IMP] contract: Improve partner contract smartbutton

carlosdauden:
@Tecnativa

### DIFF
--- a/contract/models/res_partner.py
+++ b/contract/models/res_partner.py
@@ -18,8 +18,7 @@ class ResPartner(models.Model):
         for partner in self:
             partner.contract_count = Contract.search_count([
                 ('recurring_invoices', '=', True),
-                ('partner_id', '=', partner.id),
-                ('date_start', '<=', today),
+                ('partner_id', 'child_of', partner.ids),
                 '|',
                 ('date_end', '=', False),
                 ('date_end', '>=', today),
@@ -37,9 +36,10 @@ class ResPartner(models.Model):
                 self.env.context,
                 search_default_recurring_invoices=True,
                 search_default_not_finished=True,
+                search_default_partner_id=self.id,
                 default_partner_id=self.id,
                 default_recurring_invoices=True,
+                default_pricelist_id=self.pricelist_id.id,
             ),
-            domain=[('partner_id', '=', self.id)],
         )
         return res
--- a/contract/views/account_analytic_account_view.xml
+++ b/contract/views/account_analytic_account_view.xml
@@ -116,12 +116,11 @@
         <field name="model">account.analytic.account</field>
         <field name="inherit_id" ref="analytic.view_account_analytic_account_search"/>
         <field name="arch" type="xml">
+            <field name="partner_id" position="after">
+                <field name="partner_id" filter_domain="[('partner_id', 'child_of', self)]"
+                       string="Partner and dependents"/>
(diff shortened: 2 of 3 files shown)


pedrobaeza (approved):

chienandalu (approved):
Tested on runbot :+1:

contract/models/res_partner.py
```diff
@@ -18,8 +18,7 @@ def _compute_contract_count(self):
         for partner in self:
             partner.contract_count = Contract.search_count([
                 ('recurring_invoices', '=', True),
-                ('partner_id', '=', partner.id),
-                ('date_start', '<=', today),
+                ('partner_id', 'child_of', partner.id),
```
lasley: IIRC `child_of` expects an array of IDs, so this should be `partner.ids`. I thought maybe this had changed in v11 (because I thought it threw a warning), but a [search of Odoo code backs this up](https://github.com/odoo/odoo/search?utf8=%E2%9C%93&q=child_of&type=).

contract/views/account_analytic_account_view.xml
```diff
@@ -115,12 +115,11 @@
         <field name="model">account.analytic.account</field>
         <field name="inherit_id" ref="analytic.view_account_analytic_account_search"/>
         <field name="arch" type="xml">
+            <field name="partner_id" position="after">
+                <field name="partner_id" filter_domain="[('partner_id', 'child_of', self)]"
```
lasley, line 120: Same note regarding the array of ids: `[self]`

lasley (changes requested):
Thanks @carlosdauden - minor comments

contract/views/account_analytic_account_view.xml
lasley: Actually wait. Shouldn't this be `[id]`?

pedrobaeza:
@carlosdauden please check this

rafaelbn (approved):
Functionally taking, tested in runbot 👍 @carlosdauden thanks!
@lasley made some questions regarding this PR

contract/models/res_partner.py
```diff
@@ -37,9 +36,9 @@ def act_show_contract(self):
                 self.env.context,
                 search_default_recurring_invoices=True,
                 search_default_not_finished=True,
+                search_default_partner_id=self.id,
                 default_partner_id=self.id,
                 default_recurring_invoices=True,
```
sergio-teruel, line 41: add default_pricelist_id=self.pricelist_id.id

pedrobaeza: This is for v11. You might want it for v10, isn't it?

pedrobaeza:
@carlosdauden please check latest comments and check Travis result.

contract/views/account_analytic_account_view.xml
carlosdauden: This self is the searchbar text

contract/models/res_partner.py
pedrobaeza: This is still an ID instead of a list.

pedrobaeza:
And please squash commits.

carlosdauden:
Squashed

pedrobaeza:
OK, all comments attended, so I merge.

PR #133, opened 2017-12-23
[IMP] contract: Automatic prices for lines

pedrobaeza:
Includes migration script for backwards compatibility

@Tecnativa

### DIFF
--- a/contract/models/account_analytic_invoice_line.py
+++ b/contract/models/account_analytic_invoice_line.py
@@ -1,9 +1,9 @@
 # -*- coding: utf-8 -*-
-# © 2004-2010 OpenERP SA
-# © 2014 Angel Moya <angel.moya@domatix.com>
-# © 2015 Pedro M. Baeza <pedro.baeza@tecnativa.com>
-# © 2016 Carlos Dauden <carlos.dauden@tecnativa.com>
+# Copyright 2004-2010 OpenERP SA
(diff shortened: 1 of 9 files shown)


contract/models/account_analytic_invoice_line.py
```diff
@@ -61,6 +71,35 @@ class AccountAnalyticInvoiceLine(models.Model):
...
+        """Get the specific price if no auto-price, and the price obtained
+        from the pricelist otherwise.
+        """
+        for line in self:
+            if line.automatic_price:
+                product = line.product_id.with_context(
+                    quantity=line.quantity,
+                    pricelist=line.analytic_account_id.pricelist_id.id,
+                    partner=line.analytic_account_id.partner_id.id,
+                    date=line.env.context.get('old_date', fields.Date.today()),
```
carlosdauden, line 92: All OK when create invoice, but recurring_next_date will be ignored on view.

carlosdauden (approved):
Minor comment

contract/models/account_analytic_invoice_line.py
pedrobaeza: Well, when adding lines, we will see the price for the current date, which is more than acceptable IMO.

lasley (approved):
Thanks @pedrobaeza

PR #135, opened 2017-12-30
[10.0][FIX] Fix agreement_account view by using form sheet

robyf70:
Just a small view fix adding sheet tag in form view


pedrobaeza:
This has been changed already.

PR #140, opened 2018-01-09
[IMP] contract: Add computed dates from/to period invoiced

carlosdauden:
This change improves too contract_variable_quantity filters

@Tecnativa

### DIFF
--- a/contract/__manifest__.py
+++ b/contract/__manifest__.py
@@ -1,14 +1,14 @@
 # Copyright 2004-2010 OpenERP SA
 # Copyright 2014-2017 Tecnativa - Pedro M. Baeza
 # Copyright 2015 Domatix
-# Copyright 2016-2017 Tecnativa - Carlos Dauden
+# Copyright 2016-2018 Tecnativa - Carlos Dauden
 # Copyright 2017 Tecnativa - Vicent Cubells
 # Copyright 2016-2017 LasLabs Inc.
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 {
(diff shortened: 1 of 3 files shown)


pedrobaeza:
I don't think this a good change. We already have the information on the context, and this allows other modules like line #109 to work without too much changes. Why are you making this change?

contract_variable_quantity has access to context also, so no need for doing this.

carlosdauden:
This change only abstract this implementation
https://github.com/OCA/contract/pull/140/files#diff-47b79491db78ebebcb372014ce9c7bf3L167
to computed fields.

I don't modify context values, therefore other modules can access to it.

Adding operation with relativedelta is not allowed by safe_eval in contract_variable_quantity.

pedrobaeza:
What you are altering is the method for inserting the markers, forcing to use the computed values from contract. Please check #109, which sets periodicity by line, which is broken with this.

And for what you need to use `relativedelta` in the formula for variable quantity? We already have the needed dates.

carlosdauden:
If we already have the needed dates, we don't need this:
https://github.com/OCA/contract/pull/140/files#diff-47b79491db78ebebcb372014ce9c7bf3L167
but date_from and  date_to are computed there.

I don't see what is the problem in the dirty PR https://github.com/OCA/contract/pull/109 to v10.0

carlosdauden:
@pedrobaeza changes done

pedrobaeza (approved):

pedrobaeza:
@rafaelbn please review

rafaelbn (approved):
👍

PR #142, opened 2018-01-18
[10.0][ADD] contract_cancel: Add module.

### LINKED ISSUE #64: [RFC] contract_cancel: Cancel Contract With Lapsed Invoices
This will be a fairly simple module that will automatically cancel any contract that has an invoice in `Past Due` status, instead of creating a new invoice. It will provide this option at the contract and contract template level in order to allow for customization where needed.

It will hook into the pre-existing [`account.analytic.account.recurring_create_invoice`](https://github.com/OCA/contract/blob/10.0/contract/models/account_analytic_account.py#L167), and will first evaluate whether there are any past due invoices related to the contract. If there are, the `recurring_invoices` option will be deselected thus cancelling the contract. The super will not be called in this instance.

Simple enough I think, although I'm not sure if I'm going about the whole cancelling of the contract correctly. I think deactivating it is not what we're going for either though?

Let me know if you know of a pre-existing module/strategy for this that I'm overlooking, or if I'm going about this incorrectly.

woodbrettm:
* Resolves #64 

Hi there,

This module adds functionality for the automatic canceling of contracts if a related account.invoice is past due.

The canceling of a contract would be disabling the recurring invoices option.

Thanks,
Brett

@lasley


pedrobaeza:
Is someone interested in this module? If not, I will close it, as the authors are not anymore on Odoo world.

PR #150, opened 2018-03-12
Imp/invoice creation

yk2kus:
Do not populate 'reference' in an invoice from contracts while creating invoices because this field is labeled as "Vendor Reference" targeted only for supplier invoices (not visible in customer invoices) while contracts are created for customers.
Update origin with the additional information name along with code [field used as source document]
'origin': '[' + self.code +'] ' + self.name if self.code else self.name,


contract/models/account_analytic_account.py
```diff
@@ -169,14 +169,14 @@ def _prepare_invoice(self):
             self.company_id.currency_id
         )
         invoice = self.env['account.invoice'].new({
-            'reference': self.code,
+            # 'reference': self.code,
```
pedrobaeza, line 172: I prefer to leave the reference

pedrobaeza:
Any news on this?

levkar:
why not have the reference if we can?

rafaelbn:
This is dicussed in https://github.com/OCA/contract/pull/122 but it's right whats @yk2kus says. This can be considered as a BUG

PR #152, opened 2018-03-26
[IMP] contract: Automatic prices for lines

pedrobaeza:
The same as #133, but for v11

@Tecnativa


carlosdauden (approved):
Climbing the kerb like an asturian Formula 1 driver
:wink:

PR #154, opened 2018-03-28
[IMP] contract: Add extend existing invoice possibility

carlosdauden:
@Tecnativa

### DIFF
--- a/contract/models/account_analytic_account.py
+++ b/contract/models/account_analytic_account.py
@@ -236,10 +236,30 @@ class AccountAnalyticAccount(models.Model):
         return invoice._convert_to_write(invoice._cache)
 
     @api.multi
-    def _create_invoice(self):
+    def _prepare_invoice_update(self, invoice):
+        vals = self._prepare_invoice()
+        update_vals = {
+            'contract_id': self.id,
+            'date_invoice': vals.get('date_invoice', False),
+            'reference': ' '.join(filter(None, [
+                invoice.reference, vals.get('reference')])),
+            'origin': ' '.join(filter(None, [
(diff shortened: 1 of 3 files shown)


pedrobaeza:
I will do it the contrary, taking precedence on the values of the invoice passed as argument, and it lacks to check if the invoice is already validated for example. As it's a WIP that is not going to be finished, I close.

pedrobaeza:
I reopen this as this has been reinitiated. Please take into account my comments and resolve conflicts.

carlosdauden:
Changes done

pedrobaeza:
Travis is red

carlosdauden:
Travis green

contract/models/account_analytic_account.py
```diff
@@ -236,10 +236,30 @@ def _prepare_invoice(self):
         return invoice._convert_to_write(invoice._cache)
 
     @api.multi
-    def _create_invoice(self):
+    def _prepare_invoice_update(self, invoice):
+        vals = self._prepare_invoice()
```
pedrobaeza, line 240: This is still overwriting all the previous values, where they should be more important than new ones.

carlosdauden: No, this function return update_vals.
Review the function, vals only is taked to respect any changes in _prepare_invoice asignation fields

pedrobaeza: OK, got it

pedrobaeza (approved):

rafaelbn (approved):
Tested in runbot :+1:

PR #156, opened 2018-03-28
[FIX] contract: Onchange contract template raise error

carlosdauden:
@Tecnativa


pedrobaeza (approved):
Fast-tracking this as important error.

PR #158, opened 2018-03-30
[FIX] contract: Invalid pricelist field name

carlosdauden:
@Tecnativa


pedrobaeza (approved):
Fast-tracking when Travis finishes for solving the problem

PR #163, opened 2018-04-24
[IMP] contract_variable_quantity: Recompute price when changing qty

pedrobaeza:
If you have contract lines with automatic price and your pricelist contains different prices per quantity, the price is not changed when computing quantity.

This PR fixes this.

@Tecnativa


carlosdauden (approved):

rafaelbn (approved):
Tested in production enviroment :+1:

PR #168, opened 2018-06-13
[FIX] contract: Custom values overwrited by xml, missed noupdate

carlosdauden:
@Tecnativa

### DIFF
--- a/contract/data/contract_cron.xml
+++ b/contract/data/contract_cron.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding='UTF-8'?>
-<odoo>
+<odoo noupdate="1">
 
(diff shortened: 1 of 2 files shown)


contract/data/contract_cron.xml
```diff
@@ -1,5 +1,5 @@
-<?xml version="1.0" encoding='UTF-8'?>
-<odoo>
+<?xml version="1.0"?>
```
yajo: You must keep the encoding for XML files. It's Python 3 where that's no more needed, but XML is not Python

yajo (changes requested):

astirpe (approved):

yajo (approved):

rafaelbn (approved):

yajo:
:shipit: :dancer:

PR #169, opened 2018-06-26
[FIX][11.0][contract] analytic invoice/contract lines inheritance

katyukha:
Bug description
---------------

`account.analytic.contract.line` inherits
`account.analytic.invoice.line`

`account.analytic.invoice.line` defines field `analytic_account_id`:
   - comodel='account.analytic.account'

`account.analytic.contract.line` redefines field `analytic_account_id`:
   - comodel='account.analytic.contract'

On attempt to extend `account.analytic.invoice.line` model adding
field that depends on `analytic_account_id.date_start`
Odoo fails to update, because it adds this field to
`account.analytic.contract.line` through inheritance,
and `account.analytic.contract` model have no this field.

What is done
------------

Change inheritance order:
- `account.analytic.invoice.line` inherits `account.analytic.contract.line`
- all logic from `account.analytic.invoice.line` moved to `account.analytic.contract.line`


contract_variable_quantity/models/contract.py
```diff
@@ -46,7 +46,7 @@ def _prepare_invoice_line(self, line, invoice_id):
 
 
 class AccountAnalyticInvoiceLine(models.Model):
```
MiquelRForgeFlow: AccountAnalyticContractLine

MiquelRForgeFlow (approved):
Interesting. I think you are in the good way.

MiquelRForgeFlow (changes requested):
I think you could the same, but without moving that many lines. Prove to do the same but just only with renaming the classes, changing inherits and moving the <code>analytic_account_id</code> field. Please, try to minimize the changes.

katyukha:
Hello @mreficent 

Sorry for long response.

If i just rename classes and change inherits, but not rename files, then file name will not reference to model it contains. Is it Ok?

Diff will look like:

```diff
diff --git a/contract/models/account_analytic_contract_line.py b/contract/models/account_analytic_contract_line.py
index a18d5db..9328178 100644
--- a/contract/models/account_analytic_contract_line.py
+++ b/contract/models/account_analytic_contract_line.py
@@ -4,14 +4,13 @@
 from odoo import fields, models
 
 
-class AccountAnalyticContractLine(models.Model):
-    _name = 'account.analytic.contract.line'
-    _description = 'Contract Lines'
-    _inherit = 'account.analytic.invoice.line'
+class AccountAnalyticInvoiceLine(models.Model):
+    _name = 'account.analytic.invoice.line'
+    _inherit = 'account.analytic.contract.line'
 
     analytic_account_id = fields.Many2one(
-        string='Contract',
-        comodel_name='account.analytic.contract',
+        'account.analytic.account',
+        string='Analytic Account',
         required=True,
         ondelete='cascade',
     )
diff --git a/contract/models/account_analytic_invoice_line.py b/contract/models/account_analytic_invoice_line.py
index 8f547aa..5eff76a 100644
--- a/contract/models/account_analytic_invoice_line.py
+++ b/contract/models/account_analytic_invoice_line.py
@@ -14,8 +14,9 @@ from odoo.exceptions import ValidationError
 from odoo.tools.translate import _
 
 
-class AccountAnalyticInvoiceLine(models.Model):
-    _name = 'account.analytic.invoice.line'
+class AccountAnalyticContractLine(models.Model):
+    _name = 'account.analytic.contract.line'
+    _description = 'Contract Lines'
     _order = "sequence,id"
 
     product_id = fields.Many2one(
@@ -24,8 +25,8 @@ class AccountAnalyticInvoiceLine(models.Model):
         required=True,
     )
     analytic_account_id = fields.Many2one(
-        'account.analytic.account',
-        string='Analytic Account',
+        string='Contract',
+        comodel_name='account.analytic.contract',
         required=True,
         ondelete='cascade',
     )
diff --git a/contract_variable_quantity/models/contract.py b/contract_variable_quantity/models/contract.py
index 1d25df9..dff047a 100644
--- a/contract_variable_quantity/models/contract.py
+++ b/contract_variable_quantity/models/contract.py
@@ -45,8 +45,8 @@ class AccountAnalyticAccount(models.Model):
         return vals
 
 
-class AccountAnalyticInvoiceLine(models.Model):
-    _inherit = 'account.analytic.invoice.line'
+class AccountAnalyticContractLine(models.Model):
+    _inherit = 'account.analytic.contract.line'
 
     qty_type = fields.Selection(
         selection=[
```

NL66278:
For long term maintainance i would rather have a lot of changes in one PR, then some completely misnamed files in perpetuity. But maybe the changes can be minimized by having one commit renaming the files and then another with the minimal changes to give the files their proper role.

I would definitely be interested in backporting the solution adopted to 10.0, where I hit the same problem.

NL66278 (approved):
:+1: LGTM

MiquelRForgeFlow:
@katyukha As @NL66278 said, the best solution is make a commit with the minimum changes and another commit with the rename of the files

katyukha:
Reimplemented it as two commits: first - minimal changed and second - only file renames.

And during this work, i see there is some hacks in module code to handle incorrect inheritance. Something like:

```python
if self._name != 'account.analytic.invoice.line':
    return
```

This code is used in computation methods on line's dates fields. May be it have sense to move these fields to `account.analytic.invoice.line` and set proper depends on compute methods?

MiquelRForgeFlow:
Something you missed, as travis ci failed.

Also, can you please divide the renaming commit in two commits step?

Like:
* first rename commit:
    account_analytic_contract_line.py -> account_analytic_contract_line_bis.py
    account_analytic_invoice_line.py -> account_analytic_invoice_line_bis.py
* second rename commit:
    account_analytic_contract_line_bis.py -> account_analytic_invoice_line.py
    account_analytic_invoice_line_bis.py -> account_analytic_contract_line.py

Then you may squash the first rename commit with the fix inheritance just to have 2 commits in all instead of 3 if you want, but I think that it's already fine with 3 commits.

Seems a bit stupid, but it would be very nice!! 😊

katyukha:
All done! :)

MiquelRForgeFlow (approved):

MiquelRForgeFlow:
remove the [wip] in the title 😉

katyukha:
removed :)

JordiBForgeFlow:
@katyukha I understand that this change impacts on the data model, does it? in that case, changes to the data model should include a migration script.

katyukha:
@jbeficent This PR changes only inheritance order. Data structure (database tables and fields) is not changed, so it *seems* that no migration needed.

Side effects may happen in other addons that inherit `account.analytic.invoice.line` and expects that changes will be forwarded to `account.analytic.contract.line`. This happened in `contract_variable_quantity_addon`, and in this case that addons have to be modified to inherit from `account.analytic.contract.line` instead.

MiquelRForgeFlow:
@katyukha could you add in this PR a fix for `contract_variable_quantity` module? A PR should not break any already installable module of the same repo.

katyukha:
@mreficent This fix already here. See [diff](https://github.com/OCA/contract/pull/169/files#diff-cc7906b4de0478d114a3286e4782b662)

MiquelRForgeFlow:
Ah, thanks!

katyukha:
One thing i've missed is to update versions of addons...
Is it ok to do it as separate commit, or it would be better to ammend last one?

MiquelRForgeFlow:
As you prefer, I think there isn't preference about this. I would do a new commit.

katyukha:
Versions have been updated!
But there is strange error on travis: build failed to start.

MiquelRForgeFlow:
@katyukha It's full green again 😉

I just have run an analysis with openupgrade_records just to be sure there is not need of migration scripts :+1:  @jbeficent @pedrobaeza

PR #170, opened 2018-07-04
[11.0][IMP] contract: adapted to purchase contracts

MiquelRForgeFlow:
Based on https://github.com/OCA/contract/pull/97.

### DIFF
--- a/contract/README.rst
+++ b/contract/README.rst
@@ -25,10 +25,10 @@ Contracts Management - Recurring
 
 |badge1| |badge2| |badge3| |badge4| |badge5| 
 
-This module brings back the contracts management with recurring invoicing
-features. Also you can print and send by email contract report.
+This module enables contracts management with recurring
+invoicing functions. Also you can print and send by email contract report.
 
-In upstream Odoo, this functionality was moved into the Enterprise edition.
+It works for customer contract and supplier contracts.
 
 **Table of contents**
 
@@ -106,6 +106,7 @@ Contributors
 * Angel Moya <angel.moya@domatix.com>
 * Dave Lasley <dave@laslabs.com>
 * Vicent Cubells <vicent.cubells@tecnativa.com>
+* Miquel Raïch <miquel.raich@eficent.com>
 
 Maintainers
 ~~~~~~~~~~~
--- a/contract/__manifest__.py
+++ b/contract/__manifest__.py
@@ -8,7 +8,7 @@
 
 {
     'name': 'Contracts Management - Recurring',
-    'version': '11.0.3.0.0',
+    'version': '11.0.4.0.0',
     'category': 'Contract Management',
     'license': 'AGPL-3',
     'author': "OpenERP SA, "
--- a/contract/models/account_analytic_account.py
+++ b/contract/models/account_analytic_account.py
@@ -199,28 +199,33 @@ class AccountAnalyticAccount(models.Model):
         return invoice_line_vals
 
     @api.multi
-    def _prepare_invoice(self):
+    def _prepare_invoice(self, journal=None):
         self.ensure_one()
         if not self.partner_id:
             raise ValidationError(
                 _("You must first select a Customer for Contract %s!") %
                 self.name)
-        journal = self.journal_id or self.env['account.journal'].search(
-            [('type', '=', 'sale'),
-             ('company_id', '=', self.company_id.id)],
-            limit=1)
+        if not journal:
+            journal = self.journal_id or self.env['account.journal'].search([
+                    ('type', '=', self.contract_type),
+                    ('company_id', '=', self.company_id.id)
+                ], limit=1)
         if not journal:
             raise ValidationError(
-                _("Please define a sale journal for the company '%s'.") %
-                (self.company_id.name or '',))
+                _("Please define a %s journal for the company '%s'.") %
+                (self.contract_type, self.company_id.name or '')
+            )
         currency = (
             self.pricelist_id.currency_id or
             self.partner_id.property_product_pricelist.currency_id or
             self.company_id.currency_id
         )
+        invoice_type = 'out_invoice'
+        if self.contract_type == 'purchase':
+            invoice_type = 'in_invoice'
         invoice = self.env['account.invoice'].new({
             'reference': self.code,
-            'type': 'out_invoice',
+            'type': invoice_type,
             'partner_id': self.partner_id.address_get(
                 ['invoice'])['invoice'],
             'currency_id': currency.id,
--- a/contract/models/account_analytic_account.py
+++ b/contract/models/account_analytic_account.py
@@ -202,14 +202,19 @@ class AccountAnalyticAccount(models.Model):
     def _prepare_invoice(self, journal=None):
         self.ensure_one()
         if not self.partner_id:
-            raise ValidationError(
-                _("You must first select a Customer for Contract %s!") %
-                self.name)
+            if self.contract_type == 'purchase':
+                raise ValidationError(
+                    _("You must first select a Supplier for Contract %s!") %
+                    self.name)
+            else:
+                raise ValidationError(
+                    _("You must first select a Customer for Contract %s!") %
+                    self.name)
         if not journal:
             journal = self.journal_id or self.env['account.journal'].search([
-                    ('type', '=', self.contract_type),
-                    ('company_id', '=', self.company_id.id)
-                ], limit=1)
+                ('type', '=', self.contract_type),
+                ('company_id', '=', self.company_id.id)
+            ], limit=1)
         if not journal:
             raise ValidationError(
                 _("Please define a %s journal for the company '%s'.") %
--- a/contract/models/account_analytic_contract.py
+++ b/contract/models/account_analytic_contract.py
@@ -6,7 +6,7 @@
 # Copyright 2015-2017 Tecnativa - Pedro M. Baeza
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo import api, fields, models
+from odoo import api, fields, models, _
 
 
 class AccountAnalyticContract(models.Model):
@@ -25,6 +25,12 @@ class AccountAnalyticContract(models.Model):
     partner_id = fields.Many2one(
         comodel_name="res.partner",
         string="Partner (always False)",
+    )
+    contract_type = fields.Selection(
+        selection=[
+            ('sale', _('Sale')),
+            ('purchase', _('Purchase')),
+        ], default='sale'
     )
     pricelist_id = fields.Many2one(
         comodel_name='product.pricelist',
@@ -64,7 +70,7 @@ class AccountAnalyticContract(models.Model):
         'account.journal',
         string='Journal',
         default=lambda s: s._default_journal(),
-        domain="[('type', '=', 'sale'),('company_id', '=', company_id)]",
+        domain="[('company_id', '=', company_id)]",
     )
     company_id = fields.Many2one(
         'res.company',
@@ -72,6 +78,13 @@ class AccountAnalyticContract(models.Model):
         required=True,
         default=lambda self: self.env.user.company_id,
     )
+
+    @api.onchange('contract_type')
+    def _onchange_contract_type(self):
+        self.journal_id = self.env['account.journal'].search([
+            ('type', '=', self.contract_type),
+            ('company_id', '=', self.company_id.id)
+        ], limit=1)
 
     @api.model
     def _default_journal(self):
--- a/contract/models/account_analytic_contract.py
+++ b/contract/models/account_analytic_contract.py
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 # Copyright 2004-2010 OpenERP SA
 # Copyright 2014 Angel Moya <angel.moya@domatix.com>
 # Copyright 2016 Carlos Dauden <carlos.dauden@tecnativa.com>
@@ -6,7 +5,7 @@
 # Copyright 2015-2017 Tecnativa - Pedro M. Baeza
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo import api, fields, models, _
+from odoo import api, fields, models
 
 
 class AccountAnalyticContract(models.Model):
@@ -28,9 +27,9 @@ class AccountAnalyticContract(models.Model):
     )
     contract_type = fields.Selection(
         selection=[
-            ('sale', _('Sale')),
-            ('purchase', _('Purchase')),
-        ], default='sale'
+            ('sale', 'Customer'),
+            ('purchase', 'Supplier'),
+        ], default='sale',
     )
     pricelist_id = fields.Many2one(
         comodel_name='product.pricelist',
@@ -70,7 +69,8 @@ class AccountAnalyticContract(models.Model):
         'account.journal',
         string='Journal',
         default=lambda s: s._default_journal(),
-        domain="[('company_id', '=', company_id)]",
+        domain="[('type', '=', contract_type),"
+               "('company_id', '=', company_id)]",
     )
     company_id = fields.Many2one(
         'res.company',
@@ -81,6 +81,9 @@ class AccountAnalyticContract(models.Model):
 
     @api.onchange('contract_type')
     def _onchange_contract_type(self):
+        if self.contract_type == 'purchase':
+            self.recurring_invoice_line_ids.filtered('automatic_price').update(
+                {'automatic_price': False})
         self.journal_id = self.env['account.journal'].search([
             ('type', '=', self.contract_type),
             ('company_id', '=', self.company_id.id)
@@ -91,6 +94,6 @@ class AccountAnalyticContract(models.Model):
         company_id = self.env.context.get(
             'company_id', self.env.user.company_id.id)
         domain = [
-            ('type', '=', 'sale'),
+            ('type', '=', self.contract_type),
             ('company_id', '=', company_id)]
         return self.env['account.journal'].search(domain, limit=1)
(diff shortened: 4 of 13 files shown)


mreficent:
@jbeficent can you review?

mreficent:
@jbeficent can you review?

MiquelRForgeFlow:
@jbeficent , can yo review now?

JordiBForgeFlow (approved):
:+1: Tested functionally, including migration script

MiquelRForgeFlow:
@pedrobaeza maybe you want to review this?

contract/views/account_analytic_account_view.xml
```diff
@@ -39,7 +42,7 @@
                             />
                 </div>
                 <group col="4" attrs="{'invisible': [('recurring_invoices','!=',True)]}">
-                    <field name="contract_template_id" colspan="4"/>
+                    <field name="contract_template_id" colspan="4" domain="['|', ('contract_type', '=', contract_type), ('contract_type', '=', False)]"/>
```
RoelAdriaans: When you create a new contract template from the purchase contract, the type is a sale contract. Please provide an default_contract_type=contract_type context

contract/tests/test_contract.py
```diff
@@ -269,11 +275,11 @@ def test_contract_onchange_product_id_name(self):
...
 
     def test_contract_count(self):
-        """It should return contract count."""
-        count = self.partner.contract_count + 2
+        """It should return sale contract count."""
+        count = self.partner.sale_contract_count + 2
         self.contract.copy()
         self.contract.copy()
-        self.assertEqual(self.partner.contract_count, count)
+        self.assertEqual(self.partner.sale_contract_count, count)
```
RoelAdriaans, line 291: Please add an test for purchase_contract_count , ideally on the same partner to have different counts

RoelAdriaans (changes requested):
For the contract template, can you add the contract type to the tree view, and maybe group-by filter?

https://github.com/Eficent/contract/blob/bad0c8309450f6fd211640f5fba000ddf13735b2/contract/views/account_analytic_contract_view.xml#L65

MiquelRForgeFlow:
@RoelAdriaans comments attended

RoelAdriaans (approved):

RoelAdriaans:
Code and functional :+1: 
Build is red because of missing tests..
Lot of the changes are from res.partner where there are new lines for just creating a view. Creating a test for that looks pointless? This file is now at 62% coverage.. 

Other item is the `_onchange_contract_type` in `account_analytic_contract.py` , adding a test for that could be helpful.

contract/__manifest__.py
```diff
@@ -8,15 +8,15 @@
...
+    'version': '11.0.4.0.0',
     'category': 'Contract Management',
     'license': 'AGPL-3',
     'author': "OpenERP SA, "
               "Tecnativa, "
               "LasLabs, "
               "Odoo Community Association (OCA)",
     'website': 'https://github.com/oca/contract',
-    'depends': ['base', 'account', 'analytic'],
+    'depends': ['base', 'account_invoicing', 'analytic'],
```
pedrobaeza: Why this change?

contract/migrations/11.0.4.0.0/post-migration.py
```diff
@@ -0,0 +1,23 @@
+# -*- coding: utf-8 -*-
+# Copyright 2018 Eficent - Miquel Raïch
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+
+def migrate(cr, version):
+    """All previous contract where of 'sale' type."""
```
pedrobaeza: s/where/were

contract/migrations/11.0.4.0.0/post-migration.py
```diff
@@ -0,0 +1,23 @@
...
+    if not version:
+        return
+    cr.execute(
+        """
+        UPDATE account_analytic_account
+        SET contract_type = 'sale'
```
pedrobaeza: I'm thinking you don't need this putting as default value `sale` in the field `contract_type`

contract/models/account_analytic_contract.py
```diff
@@ -26,6 +25,12 @@ class AccountAnalyticContract(models.Model):
         comodel_name="res.partner",
         string="Partner (always False)",
     )
+    contract_type = fields.Selection(
+        selection=[
+            ('sale', _('Sale')),
```
pedrobaeza: Selection fields don't require to translate the label this way (they are extracted automatically), so remove `_` calls

contract/models/account_analytic_contract.py
```diff
@@ -73,11 +79,22 @@ class AccountAnalyticContract(models.Model):
         default=lambda self: self.env.user.company_id,
     )
 
+    @api.onchange('contract_type')
+    def _onchange_contract_type(self):
+        if self.contract_type == 'purchase':
+            for line in self.recurring_invoice_line_ids.filtered(
```
pedrobaeza: `self.recurring_invoice_line_ids.filtered('automatic_price').update({'automatic_price': False})`

contract/models/res_partner.py
```diff
@@ -7,30 +7,63 @@
...
         compute='_compute_contract_count',
     )
 
     def _compute_contract_count(self):
-        Contract = self.env['account.analytic.account']
+        contract_model = self.env['account.analytic.account']
         today = fields.Date.today()
         for partner in self:
-            partner.contract_count = Contract.search_count([
+            partner.sale_contract_count = contract_model.search_count([
```
pedrobaeza: This is not very performance wise. Please implement it through a `read_group` outside the loop with lazy=False, the domain without the contract_type, and grouping by partner_id and contract_type, and you'll get the same results but faster.

contract/models/res_partner.py
```diff
@@ -7,30 +7,63 @@
...
+            partner.purchase_contract_count = contract_model.search_count([
+                ('recurring_invoices', '=', True),
+                ('partner_id', 'child_of', partner.ids),
+                '|',
+                ('date_end', '=', False),
+                ('date_end', '>=', today),
+                ('contract_type', '=', 'purchase'),
+            ])
+
+    def act_show_sale_contract(self):
```
pedrobaeza: Please use a common method called from both methods that uses an argument with the XML-ID, but don't repeat the very similar code.

contract/views/account_analytic_account_view.xml
```diff
@@ -95,6 +98,57 @@
         </field>
     </record>
 
+    <record id="account_analytic_account_sale_form" model="ir.ui.view">
+        <field name="name">account.analytic.account.sale.form</field>
+        <field name="model">account.analytic.account</field>
+        <field name="inherit_id" ref="contract.account_analytic_account_recurring_form_form"/>
```
pedrobaeza: No need of putting `contract.`

contract/views/account_analytic_account_view.xml
```diff
@@ -95,6 +98,57 @@
...
+        <field name="mode">primary</field>
+        <field name="priority" eval="20"/>
+        <field name="arch" type="xml">
+            <field name="partner_id" position="attributes">
+                <attribute name="string">Customer</attribute>
+                <attribute name="domain">[('customer', '=', True)]</attribute>
+                <attribute name="context">{
```
pedrobaeza: Don't put values in multi-line. They are read very ugly on UI.

contract/views/account_analytic_account_view.xml
```diff
@@ -95,6 +98,57 @@
...
+            <field name="product_id" position="attributes">
+                <attribute name="domain">[('sale_ok', '=', True)]</attribute>
+            </field>
+        </field>
+    </record>
+
+    <record id="account_analytic_account_purchase_form" model="ir.ui.view">
+        <field name="name">account.analytic.account.purchase.form</field>
+        <field name="model">account.analytic.account</field>
+        <field name="inherit_id" ref="contract.account_analytic_account_recurring_form_form"/>
```
pedrobaeza: No need of putting `contract.`

contract/views/account_analytic_account_view.xml
```diff
@@ -95,6 +98,57 @@
...
+        <field name="mode">primary</field>
+        <field name="priority" eval="20"/>
+        <field name="arch" type="xml">
+            <field name="partner_id" position="attributes">
+                <attribute name="string">Supplier</attribute>
+                <attribute name="domain">[('supplier', '=', True)]</attribute>
+                <attribute name="context">{
```
pedrobaeza: Don't put values in multi-line. They are read very ugly on UI.

contract/models/account_analytic_contract.py
```diff
@@ -73,11 +79,22 @@ class AccountAnalyticContract(models.Model):
         default=lambda self: self.env.user.company_id,
     )
 
+    @api.onchange('contract_type')
+    def _onchange_contract_type(self):
```
pedrobaeza, line 83: Is this a real case? You can't modify the contract type through UI, so no onchange is going to be called.

pedrobaeza (changes requested):

contract/__manifest__.py
MiquelRForgeFlow: I can revert it if you like. But `account_invoicing` is the module that makes visible the invoicing menu. And without that menu visible, the `contract` module doesn't make sense.

contract/models/account_analytic_contract.py
MiquelRForgeFlow: In the contract template form view, you can change the contract_type.

contract/__manifest__.py
pedrobaeza: Well, the problem is to change dependencies on the middle of the lifecycle. You can put a comment for changing it on v12

contract/models/account_analytic_contract.py
pedrobaeza: I think the field should be always invisible and not changeable by UI, as you have 2 separate menus already, and imagine who weird would be to have a sale contract (Maybe a better name should be "Customer Contract"?), generate a customer invoice, and then change to purchase the contract...

pedrobaeza:
The more I think about it, the more I'm sure you should change the nomenclature from `sale` to `customer` and from `purchase` to `vendor` or `supplier`. Let's do it correctly from the beginning.

contract/__manifest__.py
MiquelRForgeFlow: Oh, right. But v12 already merged <code>account_invoicing</code> to <code>account</code> 😂

contract/models/account_analytic_contract.py
MiquelRForgeFlow: There are three views:

- the sale contract, where contract_type is invisible.
- the purchase contract, where contract_type is invisible.
- the contract template, where contract_type is visible and changeable. You don't generate invoices from contract templates.

contract/__manifest__.py
pedrobaeza: Hehe, there's nothing to do :rofl:

contract/models/res_partner.py
MiquelRForgeFlow: If I put outside the loop, what I do about <code>('partner_id', 'child_of', partner.ids),</code>?

pedrobaeza: Uhm, interesting. You can add `commercial_partner_id` field and group by it (or even better, search directly by it).

MiquelRForgeFlow: But it's not the same. That will give different result if the partner is different from its commercial_partner_id

pedrobaeza: Why? You shouldn't have "grandsons" (Odoo doesn't allow it naturally). Any way, the initial intention of that was to cover all the hierarchy.

MiquelRForgeFlow: That's not what I mean. I mean that now the search is counting the partner contracts and their child contracts. But not their parent contracts. Thus, if I use the commercial_partner_id, for child partners they will have in the count their parent also.

contract/migrations/11.0.4.0.0/post-migration.py
MiquelRForgeFlow: But this way they are already filled by the migration to avoid any rare case.

contract/models/account_analytic_contract.py
pedrobaeza: OK, I see. Thanks for clarifying me this.

contract/migrations/11.0.4.0.0/post-migration.py
pedrobaeza: There's no rare case. This will simply be filled by ORM default. Which case can be?

contract/README.rst
```diff
@@ -25,10 +25,10 @@ Contracts Management - Recurring
 
 |badge1| |badge2| |badge3| |badge4| |badge5| 
 
-This module brings back the contracts management with recurring invoicing
-features. Also you can print and send by email contract report.
+This module enables contracts management with recurring
+invoicing functions. Also you can print and send by email contract report.
 
-In upstream Odoo, this functionality was moved into the Enterprise edition.
+It works for sale contract and purchase contracts.
```
pedrobaeza: s/sale/customer
s/purchase/vendor

contract/models/account_analytic_account.py
```diff
@@ -199,28 +199,33 @@ def _prepare_invoice_line(self, line, invoice_id):
         return invoice_line_vals
 
     @api.multi
-    def _prepare_invoice(self):
+    def _prepare_invoice(self, journal=None):
         self.ensure_one()
         if not self.partner_id:
             raise ValidationError(
                 _("You must first select a Customer for Contract %s!") %
```
pedrobaeza: This text should change according the type

contract/models/account_analytic_contract.py
```diff
@@ -26,6 +25,12 @@ class AccountAnalyticContract(models.Model):
         comodel_name="res.partner",
         string="Partner (always False)",
     )
+    contract_type = fields.Selection(
+        selection=[
+            ('sale', 'Sale'),
```
pedrobaeza: customer/vendor?

pedrobaeza: on label part only (for not wasting the rest)

contract/readme/DESCRIPTION.rst
```diff
@@ -1,4 +1,4 @@
-This module brings back the contracts management with recurring invoicing
-features. Also you can print and send by email contract report.
+This module enables contracts management with recurring
+invoicing functions. Also you can print and send by email contract report.
 
-In upstream Odoo, this functionality was moved into the Enterprise edition.
+It works for sale contract and purchase contracts.
```
pedrobaeza: s/sale/customer
s/purchase/vendor

contract/views/account_analytic_account_view.xml
```diff
@@ -95,6 +98,51 @@
...
+
+    <record id="account_analytic_account_purchase_form" model="ir.ui.view">
+        <field name="name">account.analytic.account.purchase.form</field>
+        <field name="model">account.analytic.account</field>
+        <field name="inherit_id" ref="account_analytic_account_recurring_form_form"/>
+        <field name="mode">primary</field>
+        <field name="priority" eval="20"/>
+        <field name="arch" type="xml">
+            <field name="partner_id" position="attributes">
+                <attribute name="string">Supplier</attribute>
```
pedrobaeza, line 130: Vendor for being aligned with new Accounting nomenclature?

contract/models/res_partner.py
```diff
@@ -7,30 +7,44 @@
...
-                ('date_end', '=', False),
-                ('date_end', '>=', today),
-            ])
+            partner_child_ids = partner.child_ids.ids + partner.ids
+            partner.sale_contract_count = sum([
+                r[2] for r in result
+                if r[0] in partner_child_ids and r[1] == 'sale'])
+            partner.purchase_contract_count = sum([
+                r[2] for r in result
+                if r[0] in partner_child_ids and r[1] == 'purchase'])
```
pedrobaeza, line 39: Very good solution! :clap:

contract/views/account_analytic_account_view.xml
MiquelRForgeFlow: But in all the module I used the Supplier nomenclature...

MiquelRForgeFlow:
Any more comments @pedrobaeza ? 😎

pedrobaeza:
It would be good to go for full coverage :wink:

contract/migrations/11.0.4.0.0/post-migration.py
pedrobaeza: This is really not needed.

contract/models/res_partner.py
```diff
@@ -7,30 +7,44 @@
...
+        string='Purchase Contracts',
         compute='_compute_contract_count',
     )
 
     def _compute_contract_count(self):
-        Contract = self.env['account.analytic.account']
+        contract_model = self.env['account.analytic.account']
         today = fields.Date.today()
+        fetch_data = contract_model.read_group([
+            ('recurring_invoices', '=', True),
```
pedrobaeza, line 23: I'm seeing that this doesn't filter by current partners, on the selection, so it's going to be low performance also. You need to get before all the partners (parent and children) for filtering the `read_group` statement.

MiquelRForgeFlow: done

MiquelRForgeFlow:
@pedrobaeza codecov green :green_apple: 😎

pedrobaeza:
@mreficent codecov is complaining about CI failed. Maybe a rebase? I insist that the migration script is not required.

MiquelRForgeFlow:
@pedrobaeza No, it's just an error of the runbot. the same as https://github.com/OCA/runbot-addons/issues/173

pedrobaeza:
OK, but please remove the migration script commit.

MiquelRForgeFlow:
@pedrobaeza now codecov it's green and the migration commit removed

pedrobaeza (approved):

PR #171, opened 2018-07-09
[IMP] contract: possible error in migration if the column does not exist

jonerikceberio:


pedrobaeza:
Why it wouldn't exist?

jonerikceberio:
Because you have already have launched the migration before and has ended in a raise

pedrobaeza:
Uhm curious case, but this is not very problematic any way. Fast-tracking this.

pedrobaeza (approved):

PR #174, opened 2018-07-21
[WIP] [ADD] Make the time markers configurable on contact line

RoelAdriaans:
This module enables you do make the time on the contract lines configurable. 
For example, you can use #START(MMMM yyyy)# for times then you want to show only the starting month in the invoice line.

Feedback welcome, still not sure about the best way to document the options in the usage and view. This looks a but too much..


pedrobaeza:
Please don't add plurals on module names. Maybe you can put this in the main module, as it can be a generic feature interesting for all users of contracts that use markers.

RoelAdriaans:
@pedrobaeza I'll remove the plurals.

Main reason I've put this into a separate module is because it's a quite technical module, with  scary documentation. 
An option is to merge it into the main module, but put the extended documentation for the technical-group/debug mode?

pedrobaeza:
> An option is to merge it into the main module, but put the extended documentation for the technical-group/debug mode?

Yeah, I like this option!

RoelAdriaans:
Since I'm no longer working with Odoo, I'm abandoning this PR;

PR #175, opened 2018-07-23
[10.0] [IMP] New module contract_line_extended.

NL66278:
Adds related fields partner_id, date_start and date_end to contract line.

### DIFF
--- a/contract/tests/test_contract.py
+++ b/contract/tests/test_contract.py
@@ -40,6 +40,7 @@ class TestContractBase(common.SavepointCase):
             'recurring_invoices': True,
             'date_start': '2016-02-15',
             'recurring_next_date': '2016-02-29',
+            'recurring_rule_type': 'monthly'
         })
         cls.line_vals = {
             'analytic_account_id': cls.contract.id,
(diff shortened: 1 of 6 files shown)


RoelAdriaans (approved):

pedrobaeza:
Isn't more interesting to have a BI SQL view of lines and pivot/graph views for querying it, integrated on the main module? FYI, on v11 we have already added start/end dates at contract line level, but with other technique (being computed fields).

NL66278:
@pedrobaeza I looked at the 11.0 code, and I am a bit amazed on how the computed date in the contract lines for start and end are tied to the next invoice date. Also on other places on the contract module I see a mixup of the validity of a contract, which would be between date_start and date_end, and when the contract is invoiced. In the code for the customer where we use contract_line_extended we use the date_start and date_end to show the currently valid contract lines on a tab on the partner form. This seems to be the most logical use case.
I would have no objection in principle to integrate my code in the main module. But would such a PR be accepted, seeing the way things have developed in 11.0?

pedrobaeza:
The reason for that is for having the time frame period where the lines are going to be invoiced, and thus the dynamic computation. It's a larger concept that only relying on contract start and end dates, but also a bit different information given.

What I'm proposing you is to build an SQL view (`contract.report` for example), that joins header and lines, being available for a BI pivot/graph view, and then you can add a one2many field in partene over that SQL view on your custom module easily. This improvement can be on the main module.

PR #177, opened 2018-07-29
[ADD] Contract section module

RoelAdriaans:
Create a contract with sections. When the invoice is created, the sections are transferred to the invoice.
Cannot be merged into the contract module, because of the dependency on the sale module. This module is absent in the original contract module, and contract can be independent from sale module.


contract_section/tests/test_contract.py
```diff
@@ -0,0 +1,58 @@
+# Copyright 2016 Tecnativa - Carlos Dauden
+# Copyright 2017 Tecnativa - Pedro M. Baeza
+# Copyright 2018 Roel Adriaans - roel@road-support.nl
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.tests import common
+
+
+class TestContractBase(common.SavepointCase):
```
pedrobaeza: Maybe you can refactor the base tests for making them inheritable and avoid to rewrite each time the same base stuff. See for example https://github.com/OCA/l10n-spain/commit/b9b4c9d83fc33ba10febe37cedafc4ddf7598599 how I made that tests inheritable, and I use it in https://github.com/OCA/l10n-spain/commit/16db2446d53ed00dda08a94abb0b69f187a3f934#diff-09b5031fb79d45f262679b74a15ff361R31

RoelAdriaans:
@pedrobaeza Thanks for the pointer, changes applied.

contract_section/tests/test_contract.py
```diff
@@ -0,0 +1,31 @@
...
+from odoo.addons.contract.tests.test_contract import TestContractBase
+
+
+class TestContract(TestContractBase):
+    @classmethod
+    def setUpClass(cls):
+        # Reuse the tests from the contract module, but replace the
+        # contract line
+        super(TestContract, cls).setUpClass()
+        cls.acct_line.unlink()
```
pedrobaeza, line 15: Isn't easier to just do:

```python3
cls.acct_line.layout_category_id = cls.section.id
cls.line_vals['layout_category_id'] = cls.section.id
```
instead of unlinking and creating again?

RoelAdriaans: I rather test a clean line..
Test result should be the same, and isn't explicit is better than implicit? :)

pedrobaeza: Well, I think it's also explicit enough as you have the line created in super, but I won't oppose on this as you have already done a good job.

pedrobaeza (approved):

MiquelRForgeFlow:
@pedrobaeza Is this an indirect question about if sale layout categories makes sense in vendor contract lines?

pedrobaeza:
Well, it's more to request your review for collaborating in repository PSC, but yes, you have risen a good issue, and it should be taken into account now that vendor bill management has been included. Are sections allowed in purchase invoices?

MiquelRForgeFlow:
Well, the invoice report is provided in `sale` module, not in the `purchase`. So the purchase invoices use the same template as the sale invoices (it just happens that `sale` module is required to enable sections). Thus, the answer is yes, the sections are also allowed in purchase invoices.

In fact, Odoo should remove the sale layout categories and invoice report from `sale` and put them in `account`. The people that have `purchase` installed but not `sale` don't have any way to use the invoice report with sections, and that is sad for them 😢

😂

MiquelRForgeFlow (approved):

pedrobaeza:
Thanks for the investigation.

MiquelRForgeFlow:
@pedrobaeza In fact, because I was unaware up until now, Odoo did in v12 what I said before, see https://github.com/odoo/odoo/commit/762adedece1c6c2b97b282348659c1b39b988d24 😎
They refactored the sections putting them in `account` with a more subtile way.

pedrobaeza:
Great to know. Noted for OpenUpgrade.

PR #178, opened 2018-08-03
[10.0] [FIX] contract: Make sure tests don't fail on other languages

NL66278:
Make sure tests do not fail because of missing sale journal.

### DIFF
--- a/contract/__manifest__.py
+++ b/contract/__manifest__.py
@@ -9,7 +9,7 @@
 
 {
     'name': 'Contracts Management - Recurring',
-    'version': '10.0.4.3.0',
+    'version': '10.0.4.3.1',
     'category': 'Contract Management',
     'license': 'AGPL-3',
     'author': "OpenERP SA, "
--- a/contract/tests/test_contract.py
+++ b/contract/tests/test_contract.py
@@ -9,9 +9,30 @@ from odoo.tests import common
 
 
 class TestContractBase(common.SavepointCase):
+
+    post_install = True
+    at_install = False
+
     @classmethod
     def setUpClass(cls):
         super(TestContractBase, cls).setUpClass()
+        # Make sure a sale journal is present for tests
+        sequence_model = cls.env['ir.sequence']
+        contract_sequence = sequence_model.create({
+            'company_id': cls.env.user.company_id.id,
+            'code': 'contract',
+            'name': 'contract sequence',
+            'number_next': 1,
+            'implementation': 'standard',
+            'padding': 3,
+            'number_increment': 1})
+        journal_model = cls.env['account.journal']
+        journal_model.create({
+            'company_id': cls.env.user.company_id.id,
+            'code': 'contract',
+            'name': 'contract journal',
+            'sequence_id': contract_sequence.id,
+            'type': 'sale'})
         cls.partner = cls.env.ref('base.res_partner_2')
         cls.product = cls.env.ref('product.product_product_2')
         cls.product.taxes_id += cls.env['account.tax'].search(
@@ -286,10 +307,14 @@ class TestContract(TestContractBase):
         line = self._add_template_line()
         line.product_id.description_sale = 'Test'
         line._onchange_product_id()
-        self.assertEqual(line.name,
-                         '\n'.join([line.product_id.name,
-                                    line.product_id.description_sale,
-                                    ]))
+        # Make sure we use same language settings as in tested code:
+        product = line.product_id.with_context(
+            lang=self.partner.lang,
+            partner=self.partner.id)
+        product_name = product.name_get()[0][1]
+        self.assertEqual(
+            line.name,
+            '\n'.join([product_name, line.product_id.description_sale]))
 
     def test_contract_count(self):
         """It should return contract count."""
--- a/contract/tests/test_contract.py
+++ b/contract/tests/test_contract.py
@@ -307,11 +307,7 @@ class TestContract(TestContractBase):
         line = self._add_template_line()
         line.product_id.description_sale = 'Test'
         line._onchange_product_id()
-        # Make sure we use same language settings as in tested code:
-        product = line.product_id.with_context(
-            lang=self.partner.lang,
-            partner=self.partner.id)
-        product_name = product.name_get()[0][1]
+        product_name = line.product_id.name_get()[0][1]
         self.assertEqual(
             line.name,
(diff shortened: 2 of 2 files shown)


pedrobaeza:
What is the use case where you don't have journal? CoA not installed?

NL66278:
We have encountered this in our own runbot instances. Also the tests would fail for buildouts based on another language than en_US. So these changes decrease the amount of assumptions the test makes about the environment in which the tests are run.

pedrobaeza:
But you will have the same problem with accounts, so adding this bunch of code is a lot. You should fix your test process instead, setting for example an initial country in your DB.

NL66278:
We set nl_NL in our database. That is why the test based on a journal description (which assumes English always) was not working. Also the tests for the contract module are run before our custom chart of account is installed, which is logical because our custom modules are based on OCA or Odoo modules, not the other way around obviously.

pedrobaeza:
I don't see where is the "test based on a journal description". Can you point me to it? I see better to make the test more resilient to that than to create the journal disconnected from the rest.

NL66278:
Sorry it is not a journal description but the line name compared with the product name in line 267 of the existing code.

pedrobaeza:
OK, that patch is correct, but the conflicting one is the journal manual creation.

pedrobaeza:
I have change the title to the most suitable one. Are you going to fix this?

NL66278:
@pedrobaeza Maybe we still have a different view on the need or not of having the tests independent of Chart of Account installation.

The travis errors are due to a timeout in another module. I will look into that next monday.

pedrobaeza:
It's not a different vision, but simply that with your approach, integration tests doesn't work correctly, as no chart account is possible to be installed having your extra stuff

NL66278:
@pedrobaeza Maybe I misunderstand you. I do not quite see how my extra stuff (the creation of a journal) would prevent installation of a chart of accounts. Especially not because the data / records added by the test will be automatically rolled back when the test is finished.

pedrobaeza:
You are right about the rollback, but I remember one collision doing something similar one time. OK, let's greenify this and merge, and will see if there's a problem in the future.

rafaelbn:
Hello, could we update this PR? conflicts + travis :red_circle:  please! thanks a lot!

NL66278:
Travis errors seem to be unrelated. Problem with mailing invoice from another module.

yajo:
Hi there @NL66278. I fought in the past with your exact same problem. :disappointed:  When you get to fix the accounting in foreign language problem, your'e gonna start seeing lots of other tests breaking. Some of them just can't be written in other form, like i.e. the website tours (which need for instance the product names in English). The same can happen when moving tests to post mode to make them run fine on your custom addons; you'll find some folks here at OCA rolling back your changes and you can't do anything about it because, after all, they're right.

So, basically, run tests in English or you'll enter an endless rabbit hole that will lead you to... :rabbit: run your tests in English :smile:. I've been there too.

Alternatives that will make you sleep better:

- Enable `l10n_generic_coa` in your CI envs, although it's not present in production ones.
- Disable, or defer to a later step, testing country-specific addons. Usually the dependency chain handles that.
- You can `from odoo.addons.contract.tests.test_contract import TestContractBase` and extend it in your own CoA to adapt it to your custom needs, so you actually test your CoA too.

I have a feeling this patch is going to break more things that fix. Would any of these strategies (or another one you can think of) work for you?

Also, **Travis' :x: comes from this PR** (main branch is :heavy_check_mark:). Possibly an unclosed cursor from the `setUpClass` of a `SavepointCase`. It's a nasty bug that I've hit many times.

NL66278:
@Yajo we are running our tests in English. Although the default language of most of our buildouts is Dutch, we always switch the administrator to English. A few test will still fail then, but we have patched Odoo to run these correctly as well (see: https://github.com/nl66278/odoo/tree/10.0-delete-some-tests).

However the main point of this PR is not the language, but to run the tests after all modules have installed, and to run them independently of wether or not there already is a journal created.

Travis is red because the test for one other module reach out to the internet with a http request that timed out. I really think tests that are dependent on external services should not be part of a global integration test like runbot.

pedrobaeza:
> Travis is red because the test for one other module reach out to the internet with a http request that timed out. I really think tests that are dependent on external services should not be part of a global integration test like runbot.

On that, I don't agree, as if a module integrates an external service, you have to test that it works, and that can only be done testing with the real external service. Mocking and so on is cheating. For example, `base_location_geonames_import` must connect to Geonames.org to download a file and test the proper import, or if the format of the file has changed, then the module becomes unuseful as it won't work when using it in real life.

NL66278:
@pedrobaeza I am not saying those modules should not be tested, just that they should not be tested as part of an integral tests of all the modules in a repository (in addition to all the Odoo modules) that can together depend on lots of external services, the non availability of any of them causing a travis failure that is unrelated to the actual changes in the PR.

yajo:
> to run the tests after all modules have installed, and to run them independently of wether or not there already is a journal created

If the test works no matter if there is or not a journal installed, then why do you need to move it to post install mode? :thinking:

pedrobaeza:
Again I don't agree. The combination of several modules in an specific installation might make them to fail, and you need to check that. If not, your conditions won't be reproducible. Scientific method: always recreate all the variables of the experiment.

NL66278:
@Yajo The post install is to test it in combination with all the other modules to be installed. On that I fully agree with the last commit by @pedrobaeza .

NL66278:
@pedrobaeza I have no specific proposal for now on how to actually run the tests with the externally dependent tests left out. So pushed my branch again to force a new Travis run. I agree on scientist reproducing tests while keeping all variables under control. We just don't control the external variables. So it is like doing research into the growth of bacteria with the windows open, instead of in a clean room.

yajo:
> The post install is to test it in combination with all the other modules to be installed

But the test is stateless now, isn't it? Other modules shouldn't affect it.

Relying in random modules installed or not is somehow like relying in external network access: not reproducible.

I think that the main problem is that Odoo lacks a proper integration testing framework, so we tend to use unit tests for integration tests. OCA should provide a solution on _that_ matter. :thinking:

rafaelbn:
Do we agree something here? Thanks. Travis is :red_circle:

yajo:
I guess the problem is the PR title, which is misleading because the fix has nothing to do with languages.

The fix is OK, but it should be in at-install mode and travis should be green.

PR #190, opened 2018-10-03
[FIX] Contract Spanish Translation

j-zaballa:
- Module: contract
    - Summary: several terms where not translated to Spanish.


pedrobaeza:
This should be done through Weblate, but I'm going to merge it as this is the first time.

PR #195, opened 2018-10-14
[IMP] contract,contract_payment_auto: do not rollback the whole trans…

### LINKED ISSUE #166: [contract_payment_auto] Lack of robustness - missing commits after each payment (?)
Hi folks,

I recently faced an important issue (see https://github.com/OCA/contract/issues/165) which rolled back the creation of a bunch of payment transactions in odoo, while their acquirer counterpart operations were sent to the bank. At next cron execution time, all these operations were then re-executed, resulting in a double bank payment.

The consequences would have been mitigated if a commit had been executed in the recurring task loop. Do you agree? Where in the code would you put that commit? In `AccountAnalyticAccount.recurring_create_invoice` in contract/models/account_analytic_account.py? Any advice before I give it a try and submit a pull request?

fcayre:
…action on recurring invoice creation and auto pay retry errors

Fixes #166.

Rather save the database state before each invoice creation (or auto pay
retry) so that only those in failure are rollbacked.

Before the change, the whole transaction was rollbacked on error
although the recurring invoice creations (resp. auto pay retries) are
independant operations.

This was leading to specially bad bugs when using an HTTP-driven service
for auto payments, as the service could register payments that odoo would
rollback on a later invoice creation crash in the same cron execution,
thus auto paying them again at next cron execution although they were
already registered as paid in the distant service.

### DIFF
--- a/contract/models/account_analytic_account.py
+++ b/contract/models/account_analytic_account.py
@@ -6,11 +6,18 @@
 # Copyright 2016-2017 LasLabs Inc.
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
+import logging
+import traceback
+from contextlib import contextmanager
+
 from dateutil.relativedelta import relativedelta
 
 from odoo import api, fields, models
 from odoo.exceptions import ValidationError
 from odoo.tools.translate import _
+
+
+_logger = logging.getLogger(__name__)
 
 
 class AccountAnalyticAccount(models.Model):
@@ -259,6 +266,17 @@ class AccountAnalyticAccount(models.Model):
         invoice.compute_taxes()
         return invoice
 
+    @contextmanager
+    @api.multi
+    def logged_exception_savepoint(self, *args):
+        try:
+            with self.env.cr.savepoint() as ctx:
+                yield ctx
+        except:
+            _logger.exception(traceback.format_exc())
+            if args:
+                _logger.exception(*args)
+
     @api.multi
     def recurring_create_invoice(self):
         """Create invoices from contracts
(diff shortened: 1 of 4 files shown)


contract/models/account_analytic_account.py
```diff
@@ -259,38 +266,52 @@ def _create_invoice(self):
...
-                'next_date': new_date,
-                # Force company for correct evaluation of domain access rules
-                'force_company': contract.company_id.id,
-            })
-            # Re-read contract with correct company
-            invoices |= contract.with_context(ctx)._create_invoice()
-            contract.write({
-                'recurring_next_date': fields.Date.to_string(new_date)
-            })
+            with self.logged_exception_savepoint(error_msg_tmpl, contract.id):
```
pedrobaeza, line 292: Is it needed to enclose this in a new method? When any errors happen, you will already have logging entries, so with this, you will duplicate the logging activity.

fcayre: Hello Pedro,

you're probably right for the traceback part, I am going to remove it, thanks for the remark. In the opposite, the contextual part (the error sentence with the contract id in this case) is pretty useful in my experience, as the other logging entries you may get in such circumstances don't help much: they often refer to rollback-removed model instances in the best case, or worse, a high-level http error (to the payment acquirer) without any clue to identify the e.g. faulty db data.

Also a dedicated method looks useful to me to avoid kind of critical code duplication (there are already 2 use cases here: invoice creation and auto payment retries), as well as later programming errors (like catching the exception inside the savepoint context, preventing the db rollback).

fcayre: After digging a bit more into the logging issue @pedrobaeza , I realized that logging the traceback is required here, otherwise the exception will never get logged as it is blocked in the logged_exception_savepoint and will not make it to the catch-all except clause of the kernel.

This is illustrated by the contract module tests that use the new assertExceptionLogged test method: there is no raised exception anymore, so the previous assertRaises call has to be replaced by an error logging check.

pedrobaeza: OK, but please enclose on that only the strict called to `create_invoice` and let the rest out of it, so the diff is lower and the rest of possible issues are not masked.

fcayre: Hello again @pedrobaeza and thanks for reviewing this again and again.

I first intended to do just that (even before you suggested), but changed my mind because:
* if it were only a matter of diff viewing, it would be a bad thing; git diff has a "-w" option that serves exactly this purpose, and it has a counterpart in the github UI ("Diff settings > Hide whitespace changes"); this really makes the change easy to read
* the code is clearer and of better quality as is IMHO, because the whole idea is to isolate contract handling: a crash wrt. a given contract should have no influence on the others; the present implementation does exactly that, with the context manager enclosing the whole loop content; if we do not do that, not matter the error before the context manager, it would rollback the whole transaction and destroy the desired effect.

pedrobaeza: It's not a question of viewing, but commit diff size. Any way, if you want to do that, then extract code to an isolated method, so that we can better read it, hook into it, etc

contract/models/account_analytic_account.py
```diff
@@ -259,38 +266,52 @@ def _create_invoice(self):
...
         invoices = self.env['account.invoice']
+
+        error_msg_tmpl = ('Invoice creation failed for contract id %d'
+                          ' (traceback above).'
+                          ' Going on with other contracts, if any.')
         for contract in self:
-            ref_date = contract.recurring_next_date or fields.Date.today()
-            if (contract.date_start > ref_date or
-                    contract.date_end and contract.date_end < ref_date):
-                if self.env.context.get('cron'):
```
pedrobaeza, line 273: Why are you removing this part of code? It should be here as you shouldn't raise any error (even masking it) when on cron mode, as this is a condition to check for not generating invoices, but to alert when you click interactively on the create invoice button on the contract.

pedrobaeza:
It seems there's no activity since long and conflicts, so closing. Please comment back if you want to finalize it.

PR #202, opened 2018-10-25
[12.0][MIG] - Migration module contract to 12.0

sbejaoui:

PR #203, opened 2018-10-25
[12.0][MIG] - Migration module contract_sale to 12.0

sbejaoui:
Depends on:

- [x] contract - #202


pedrobaeza:
You can rebase now

sbejaoui:
@gva-acsone can you review this please.

gva-acsone (approved):
👍 Functional test

contract_sale/__manifest__.py
```diff
@@ -0,0 +1,24 @@
...
+
+{
+    'name': 'Contract from Sale',
+    'version': '12.0.1.0.0',
+    'category': 'Sales',
+    'author': 'Tecnativa, '
+              'Odoo Community Association (OCA)',
+    'website': 'https://github.com/OCA/contract',
+    'depends': [
+        'sale',
```
rafaelbn, line 12: IMHO I think we should depends on sale_management because if not you have in any case to install sale_management always as without this module you don't have any menu sales.

rafaelbn (approved):
You have to install sale_management to test this module

Tested functionally :+1:

contract_sale/__manifest__.py
pedrobaeza: I don't think so. These are technical dependencies that prevents the module to work, but not real ones. Maybe a comment can be dropped in the installation section for testing the module.

sbidoul:
@pedrobaeza can you merge this one?

pedrobaeza:
Today I'm on holidays. Let me check on Monday.

sbidoul:
No problem, enjoy :)

pedrobaeza (approved):

PR #206, opened 2018-10-26
[11.0][ADD] Service Profile to Agreement

bodedra:


bodedra:
CC @wolfhall @max3903 @osi-scampbell

agreement/__manifest__.py
```diff
@@ -17,6 +17,8 @@
         'mail',
         'product',
         'sale_management',
+        'maintenance',
```
pedrobaeza: Please don't add this hard dependencies. You should create extension modules like `agreement_maintenance` for handling this.

pedrobaeza (dismissed):

agreement/__manifest__.py
bodedra: Make sense. I will be on vacation so will do it when I back. Thank you.

max3903 (approved):

Issue #200, opened 2018-10-25
Implementation of a 'Contract group' notion


gva-acsone:
Hello,

### Problem with periodical contract changes

I find it currently really difficult to deal with **periodical** upsells / downsells / suspensions / contract product changes, i.e when a contract is modified, but only for a fixed duration of time. 
Currently, if I want to apply a periodical upsell for instance, I need to add a new contract line and once the 'periodical upsell period' is reached, a manual action is required by a user to put the quantity on this contract line back to 0. The same is true when I want to apply a periodical contract product modification (example: a customer concludes a contract for a year for product A. But after four months, he asks to try out product B during two months, before switching back to product A for the last six months of the contract. To deal with this situation in the current module, the quantity for product A has to be set to 0 and a new contract line has to be added for product B once the modification period starts. Again, once the modification period ends, a manual intervention is necessary to change the quantity for product A back to its original value and set the quantity for product B back to zero. 

This mechanism leads to **three problems**:
1) Manual actions are required by users at certain moments during the contract life cycle, which will possibly lead to mistakes (product B invoiced for one more month etc...). 
2) I find it difficult to understand the history of the contract, i.e what happened with the contract (upsells, downsells, product changes... between the start date and now. The user has to open all the invoices to understand the history of the contract, which is annoying for multi-year contracts. 
3) While opening a contract form, users ignore for instance at what time the periodical upsell will end. So, they lack a view on the future of the contract.

I know planned activities and the tracking of certain fields (ex: product, quantity...) on contract lines could be used to avoid a part of these three problems, but we believe an easier way to deal with this use case would be to work with contract groups.

### Contract groups

Our plan would be to change the way 'upsells / suspensions / contract modifications' are recorded in the OCA module, without touching / changing the invoicing mechanism, which is crucial.

Our idea is that every 'important' change to a contract leads to the creation of a new contract, linked to the original contract by this new notion of 'Contract group' (explained below) . New buttons on a contract will enable users to trigger three different actions: Suspension, Upsell, Product modification. One contract would correspond to one contract product only. 

1° In case of an **upsell**, the user has to complete four fields in the wizard: Upsell start date ; New contract product related to the upsell; Quantity; Upsell end date (optional).
--> The original contract is not modified. However, a new contract is created by Odoo with a start and (eventual) end date and a contract product that correspond to the informations mentioned by the user in the wizard. The contract group mentioned on the original contract will be copied to the new contract, to indicate that those two contracts are related to each other. 

[...]

PR #207, opened 2018-10-30
[12.0][IMP] - Make recurrence mechanism on contract line

sbejaoui:
Make recurrence mechanism on contract line and some other refactoring
Related to #200


sbejaoui:
@gva-acsone, it's ready for a shot of test.

gva-acsone:
@sbejaoui , I've an error message when trying to add an invoice line on my contract template.
Can you check please? 
...
  File "src/lxml/parser.pxi", line 706, in lxml.etree._handleParseResult (src/lxml/lxml.etree.c:106810)
  File "src/lxml/parser.pxi", line 646, in lxml.etree._raiseParseError (src/lxml/lxml.etree.c:105956)
  File "<string>", line 0
lxml.etree.XMLSyntaxError: <no detail available>

gva-acsone:
1. On the contract lines, I would switch the end date and the next invoice date so that the end date is located just next to the start date.

2. I managed to create my first invoice, but afterwards, Odoo refuses to create my next invoices.

sbejaoui:
@gva-acsone 
> 1. On the contract lines, I would switch the end date and the next invoice date so that the end date is located just next to the start date.

I will make the change for this

> 2. I managed to create my first invoice, but afterwards, Odoo refuses to create my next invoices.

The next date invoice > the today date, that's why, odoo would not create a new invoice. Nothing to invoice until the next date invoice.

gva-acsone:
Can we modify this so that we can create invoices in advance?

gva-acsone (approved):
👍 Functional test

contract/data/contract_cron.xml
```diff
@@ -3,9 +3,9 @@
 
     <record model="ir.cron" id="account_analytic_cron_for_invoice">
         <field name="name">Generate Recurring Invoices from Contracts</field>
-        <field name="model_id" ref="analytic.model_account_analytic_account"/>
+        <field name="model_id" ref="contract.model_account_analytic_invoice_line"/>
```
pedrobaeza: This doesn't serve for existing installations as this won't be overwritten being noupdate=1

pedrobaeza: Any way, why moving this cron to lines? We can keep the same and call from that one whatever its' needed.

rafaelbn (commented):
:scream: :scream: :scream: 

Are you going to make main contract module line by line?

rafaelbn (changes requested):
Please let me know some arguments. I disagree.

- Customers don't one to define dates or invoicing types line by line
- If you create a invoice by error and after you deleted then you have to change dates for every line and more more cases

pedrobaeza:
I think recurrence by line is a good move, but with some UX refinements (I have not tried yet the PR):

* A security group for hiding recurrence by line by default.
* Recurrence definition in header, propagated to lines when set (silently if the security group is not present, and hidden on contrary).
* The same for next invoice date.

The rest can continue to happen at line level.

rafaelbn:
I guess that if @sbejaoui and @acesone in general were using this module like us from several years and their customersss were using it then the approach would be different, because you cannot change the functionality and UX so hard. Take thousands of  contract with several lines each to define all line by line.

In this PR we are changing the module for a **1 single customer case** but we have fit the standard contract module to dozens of customer. And I suppose that more Odoo integrators will be damaged by this if we don't take care of the situation.

pedrobaeza:
@rafaelbn please check my comment for the suggestion for having everything the same way but also with the possibility of defining recurrence by lines. I'm now thinking that maybe instead the security group, we can put a check at contract level for deciding if doing recurrence at line level.

sbidoul:
@rafaelbn I understand your concerns.

It's not only for one customer: the approach here is much more generic and enables many interesting features for contracts with many lines (such as having a subscriptions with some options that start and stop during the life of the contract). Also possibilities to add lines to existing contracts from the sale order, etc.

Migrating the data from previous version will work seamlessly.

Now indeed, from a UI perspective, if you have use cases with contracts with many lines where it is important to keep the recurrence information at the contract level, we can add something at the UI level to facilitate that. A checkbox on the contract saying "same recurrence on all lines" should be doable, with the recurrence field on contract being compute with inverse.

rafaelbn:
OK, thank you. I cannot test this again this week but next week.

rafaelbn (commented):
I will try my best review soon. Thank you

rafaelbn:
Hello,

OK I agree @pedrobaeza 

Thank you @sbidoul for your comments.

Review

- [ ] I still don't see any way to define for all the lines:
  - Repeat every (number an interval)
  - Invoicing type
  - Date start
  - Date of next invoice
  - Date end is nor editable
- [x] The bif Title "Recurring invoices" is missing
- [x] The checkbox "Generate recurring invoices automatically" is missing 
- [x] The link "create invoices" is missing
- [x] The button Cost/Revenue is missing
- [ ] How to disable discounts? In v11 where disabled by default (so no need to use them)
- [ ] Where are the Tags?
- [ ] IMHO the Tab "other information" is not needed there aren't so many fields

sbidoul:
@rafaelbn a couple of comments of my side. (@sbejaoui will say more)

We will add the UX feature to manage recurrence globally at the contract level a bit later, don't worry.

About the Tags, and Cost/revenue buttons: it follows from a discussion with @pedrobaeza where we agreed to make separate views for contract that are independent from analytic account views. This is a first step towards splitting contracts from analytic accounts (we will not do the full split now, but we prepare for it). So the tags and cost/revenue buttons can be seen on the native analytic account view only. I think this is indeed better so people managing contracts (usually sales persons) do not see nor need to care about analytic accounting concepts. What would be nice though, is add a button to navigate to the analytic account from the contract form.

contract/data/contract_line_constraints.py
```diff
@@ -0,0 +1,281 @@
...
+        'IS_AUTO_RENEW',
+        'HAS_SUCCESSOR',
+        'PREDECESSOR_HAS_SUCCESSOR',
+        'CANCELED',
+    ],
+)
+ALLOWED = namedtuple(
+    'ALLOWED',
+    ['PLAN_SUCCESSOR', 'STOP_PLAN_SUCCESSOR', 'STOP', 'CANCEL', 'UN_CANCEL'],
+)
```
sbidoul: This is a detail, but the python convention for named tuples, is CamelCase for the named tuple, and lower_case for the fields (since this is more or less a class with it's fields).

contract/data/contract_line_constraints.py
```diff
@@ -0,0 +1,281 @@
...
+                CANCELED=True,
+            )
+    when = compute_when(date_start, date_end)
+    has_date_end = date_end if not date_end else True
+    is_auto_renew = is_auto_renew
+    has_successor = True if successor_contract_line_id else False
+    canceled = is_canceled
+    return CRITERIA(
+        WHEN=when,
+        HAS_DATE_END=has_date_end,
```
sbidoul:
```suggestion
        HAS_DATE_END=bool(date_end),
```

contract/data/contract_line_constraints.py
```diff
@@ -0,0 +1,281 @@
...
+        IS_AUTO_RENEW=is_auto_renew,
+        HAS_SUCCESSOR=has_successor,
```
sbidoul:
```suggestion
        HAS_SUCCESSOR=bool(successor_contract_line_id),
```

sbejaoui:
Hello @rafaelbn,

Thank you for the feedback.
> * [ ]  I still don't see any way to define for all the lines:

Work in progress
> * [ ]  The bif Title "Recurring invoices" is missing
> * [ ]  The checkbox "Generate recurring invoices automatically" is missing

This checkbox was used in V8 as analytic accounts and contracts were in the same menu and have the same views. 
But, as the two concepts were separated since OCA migrated the module to V9, a contract has no use if this checkbox is unchecked.

> * [ ]  The link "create invoices" is missing
I replaced it by a native button on the form view visible in debug mode.
> * [ ]  How to disable discounts? In v11 where disabled by default (so no need to use them)
This was fixed at V12 migration, discount field is visible for users of `sale.group_discount_per_so_line`

contract/data/contract_line_constraints.py
```diff
@@ -0,0 +1,281 @@
...
+        HAS_SUCCESSOR=None,
+        PREDECESSOR_HAS_SUCCESSOR=True,
+        CANCELED=True,
+    ): ALLOWED(
+        PLAN_SUCCESSOR=False,
+        STOP_PLAN_SUCCESSOR=False,
+        STOP=False,
+        CANCEL=False,
+        UN_CANCEL=False,
+    ),
```
sbidoul: I suggest removing these criteria with `None` values: this makes the logic a bit confusing.

contract/data/contract_line_constraints.py
```diff
@@ -0,0 +1,281 @@
...
+            )
+        else:
+            return CRITERIA(
+                WHEN=None,
+                HAS_DATE_END=None,
+                IS_AUTO_RENEW=None,
+                HAS_SUCCESSOR=None,
+                PREDECESSOR_HAS_SUCCESSOR=True,
+                CANCELED=True,
+            )
```
sbidoul: As mentioned above I think it's simpler to simply get the canceled case out of the matrix.
So something like `if is_canceled: return Allowed(False, False, False, ..., uncancel=some condition based on successor and predecssor)`?

contract/data/contract_line_constraints.py
```diff
@@ -0,0 +1,281 @@
...
+from odoo.fields import Date
+
+CRITERIA = namedtuple(
+    'CRITERIA',
+    [
+        'WHEN',
+        'HAS_DATE_END',
+        'IS_AUTO_RENEW',
+        'HAS_SUCCESSOR',
+        'PREDECESSOR_HAS_SUCCESSOR',
```
sbidoul: Why is this named PREDECESSOR_HAS_SUCCESSOR and not HAS_SUCCESSOR?

contract/models/abstract_contract.py
```diff
@@ -0,0 +1,69 @@
...
+# Copyright 2016-2017 LasLabs Inc.
+# Copyright 2018 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, models, fields
+
+
+class AbstractAccountAnalyticContract(models.AbstractModel):
+    _name = 'account.abstract.analytic.contract'
+    _description = 'Abstract Account Analytic Contract'
```
sbidoul: Let's use the descriptions to provide better names (while we keep the old model names for compatibility and ease of migration).

```suggestion
    _description = 'Abstract Recurring Contract'
```

contract/models/abstract_contract.py
```diff
@@ -0,0 +1,69 @@
...
+        string='Journal',
+        default=lambda s: s._default_journal(),
+        domain="[('type', '=', contract_type),"
+        "('company_id', '=', company_id)]",
+    )
+    company_id = fields.Many2one(
+        'res.company',
+        string='Company',
+        required=True,
+        default=lambda self: self.env.user.company_id,
```
sbidoul: `self.env['res.company']._company_default_get(self._name)`

contract/models/abstract_contract_line.py
```diff
@@ -0,0 +1,204 @@
...
+
+from odoo import api, models, fields
+from odoo.addons import decimal_precision as dp
+from odoo.exceptions import ValidationError
+from odoo.tools.translate import _
+
+
+class AccountAbstractAnalyticContractLine(models.AbstractModel):
+    _name = 'account.abstract.analytic.contract.line'
+    _description = 'Account Abstract Analytic Contract Line'
```
sbidoul:
```suggestion
    _description = 'Abstract Recurring Contract Line'
```

contract/data/contract_line_constraints.py
sbejaoui: At cancel action we cut relation between the contract line and his predecessor, but in one way.
Only the successor of the predecessor is set to false. 
Like that, we permit again to perform plan_successor on the predecessor line.
We allow uncancel, only when the predecessor still without successor and lines are re-linked again.

That's why, we need `PREDECESSOR_HAS_SUCCESSOR` criteria for uncancel case.

contract/data/contract_line_constraints.py
```diff
@@ -0,0 +1,281 @@
+# Copyright 2018 ACSONE SA/NV.
```
sbidoul: I would not put this python file in the data dir, this is surprising. Pls move it to models.

contract/data/contract_line_constraints.py
sbidoul: Ok, a short comment explaining that would help.

contract/models/account_invoice.py
```diff
@@ -8,5 +8,5 @@ class AccountInvoice(models.Model):
     _inherit = 'account.invoice'
 
     contract_id = fields.Many2one(
-        'account.analytic.account',
-        string='Contract')
+        'account.analytic.account', string='Contract'
+    )
```
sbidoul: I still believe this should be removed.

contract/models/contract_line.py
```diff
@@ -0,0 +1,819 @@
...
+    _inherit = 'account.abstract.analytic.contract.line'
+
+    contract_id = fields.Many2one(
+        comodel_name='account.analytic.account',
+        string='Analytic Account',
+        required=True,
+        ondelete='cascade',
+        oldname='analytic_account_id',
+    )
+    date_start = fields.Date(string='Date Start', default=fields.Date.today())
```
sbidoul: should date_start be a required field?

contract/models/contract_line.py
```diff
@@ -0,0 +1,819 @@
...
+    create_invoice_visibility = fields.Boolean(
+        compute='_compute_create_invoice_visibility'
+    )
+    successor_contract_line_id = fields.Many2one(
+        comodel_name='account.analytic.invoice.line',
+        string="Successor Contract Line",
+        required=False,
+        readonly=True,
+        copy=False,
+        help="Contract Line created by this one.",
```
sbidoul: Elaborate a little bit the help text to explain in which circumstances a successor is created (suspension and restart, renewal, ...)

contract/models/contract.py
```diff
@@ -0,0 +1,223 @@
...
+    date_end = fields.Date(
+        compute='_compute_date_end', string='Date End', store=True
+    )
+
+    @api.depends('recurring_invoice_line_ids.date_end')
+    def _compute_date_end(self):
+        for contract in self:
+            contract.date_end = False
+            date_end = contract.recurring_invoice_line_ids.mapped('date_end')
+            if date_end and all(date_end):
```
sbidoul, line 154:
```suggestion
            if all(date_end):
```

contract/models/contract.py
```diff
@@ -0,0 +1,223 @@
...
+                contract.date_end = max(date_end)
+
+    @api.depends('recurring_invoice_line_ids.recurring_next_date')
+    def _compute_recurring_next_date(self):
+        for contract in self:
+            recurring_next_date = contract.recurring_invoice_line_ids.filtered(
+                'create_invoice_visibility'
+            ).mapped('recurring_next_date')
+            if recurring_next_date:
+                contract.recurring_next_date = min(recurring_next_date)
```
sbidoul, line 167: else contract.recurring_next_date = False?

contract/models/contract.py
```diff
@@ -0,0 +1,223 @@
...
+        compute='_compute_date_end', string='Date End', store=True
+    )
+
+    @api.depends('recurring_invoice_line_ids.date_end')
+    def _compute_date_end(self):
+        for contract in self:
+            contract.date_end = False
+            date_end = contract.recurring_invoice_line_ids.mapped('date_end')
+            if date_end and all(date_end):
+                contract.date_end = max(date_end)
```
sbidoul, line 155: else contract.date_end = False?

contract/models/contract_line.py
```diff
@@ -0,0 +1,819 @@
...
+class AccountAnalyticInvoiceLine(models.Model):
+    _name = 'account.analytic.invoice.line'
+    _inherit = 'account.abstract.analytic.contract.line'
+
+    contract_id = fields.Many2one(
+        comodel_name='account.analytic.account',
+        string='Analytic Account',
+        required=True,
+        ondelete='cascade',
+        oldname='analytic_account_id',
```
sbidoul: index = True

contract/models/contract_line.py
```diff
@@ -0,0 +1,819 @@
...
+                if rec.successor_contract_line_id:
+                    raise ValidationError(
+                        _(
+                            "A contract line with a successor "
+                            "can't be set to auto-renew"
+                        )
+                    )
+                if not rec.date_end:
+                    raise ValidationError(
+                        _("An auto-renew line should have a " "date end ")
```
sbidoul:
```suggestion
                        _("An auto-renew line must have a end date")
```

contract/models/contract_line.py
```diff
@@ -0,0 +1,819 @@
...
+                    )
+            else:
+                if not rec.date_end and rec.successor_contract_line_id:
+                    raise ValidationError(
+                        _(
+                            "A contract line with a successor "
+                            "should have date end"
```
sbidoul:
```suggestion
                            "must have a end date"
```

contract/models/contract_line.py
```diff
@@ -0,0 +1,819 @@
...
+                        )
+                    )
+
+    @api.constrains('successor_contract_line_id', 'date_end')
+    def _check_overlap_successor(self):
+        for rec in self:
+            if rec.date_end and rec.successor_contract_line_id:
+                if rec.date_end > rec.successor_contract_line_id.date_start:
```
sbidoul:
```suggestion
                if rec.date_end >= rec.successor_contract_line_id.date_start:
```

contract/models/contract_line.py
```diff
@@ -0,0 +1,819 @@
...
+                    raise ValidationError(
+                        _("Contract line and its successor overlapped")
+                    )
+
+    @api.constrains('predecessor_contract_line_id', 'date_start')
+    def _check_overlap_predecessor(self):
+        for rec in self:
+            if rec.predecessor_contract_line_id:
+                if rec.date_start < rec.predecessor_contract_line_id.date_end:
```
sbidoul:
```suggestion
                if rec.date_start <= rec.predecessor_contract_line_id.date_end:
```

contract/models/contract_line.py
```diff
@@ -0,0 +1,819 @@
...
+                recurring_rule_type, recurring_interval - 1
+            )
+        if recurring_invoicing_type == 'pre-paid':
+            return date_start
+        return date_start + self.get_relative_delta(
+            recurring_rule_type, recurring_interval
+        )
+
+    @api.onchange(
+        'is_auto_renew', 'auto_renew_rule_type', 'auto_renew_interval'
```
sbidoul: add `date_start` too?

sbidoul (commented):
Next batch. Looks quite good overall.

contract/models/contract_line.py
```diff
@@ -0,0 +1,819 @@
...
+                        % line.contract_id.name
+                    )
+
+    @api.constrains('recurring_next_date')
+    def _check_recurring_next_date_recurring_invoices(self):
+        for line in self.filtered('contract_id.recurring_invoices'):
+            if not line.recurring_next_date:
+                raise ValidationError(
+                    _(
+                        "You must supply a next invoicing date for contract "
```
sbidoul:
```suggestion
                        "You must supply a date of next invoice for contract "
```

(for consistency with the field name)

contract/models/contract_line.py
```diff
@@ -0,0 +1,819 @@
...
+            )
+
+    @api.constrains('recurring_next_date', 'date_start')
+    def _check_recurring_next_date_start_date(self):
+        for line in self.filtered('recurring_next_date'):
+            if line.date_start and line.recurring_next_date:
+                if line.date_start > line.recurring_next_date:
+                    raise ValidationError(
+                        _(
+                            "You can't have a next invoicing date before the "
```
sbidoul:
```suggestion
                            "You can't have a date of next invoice anterior to the "
```

contract/models/contract_line.py
sbidoul: Nope, I forgot a contract is still an analytic account too, so not necessarily recurring.

contract/models/contract_line.py
```diff
@@ -0,0 +1,819 @@
...
+                            "start of the contract '%s'"
+                        )
+                        % line.contract_id.name
+                    )
+
+    @api.constrains('recurring_next_date')
+    def _check_recurring_next_date_recurring_invoices(self):
+        for line in self.filtered('contract_id.recurring_invoices'):
+            if not line.recurring_next_date:
```
sbidoul: Don't we set recurring_next_date to False where there is nothing to invoice anymore?

contract/models/contract_line.py
```diff
@@ -0,0 +1,819 @@
...
+                            line.recurring_next_date <= line.date_end
+                        )
+                    else:
+                        line.create_invoice_visibility = (
+                            line.recurring_next_date
+                            - line.get_relative_delta(
+                                line.recurring_rule_type,
+                                line.recurring_interval,
+                            )
+                        ) <= line.date_end
```
sbidoul: line.create_invoice_visibility = False ?

contract/models/contract_line.py
```diff
@@ -0,0 +1,819 @@
...
+                            "end date"
+                        )
+                        % line.contract_id.name
+                    )
+
+    @api.depends('recurring_next_date', 'date_start', 'date_end')
+    def _compute_create_invoice_visibility(self):
+        today = fields.Date.today()
+        for line in self:
+            if line.date_start:
```
sbidoul: date_start is required for recurring contract lines? so this if might be redundant

contract/models/contract_line.py
```diff
@@ -0,0 +1,819 @@
...
+                            - line.get_relative_delta(
+                                line.recurring_rule_type,
+                                line.recurring_interval,
+                            )
+                        ) <= line.date_end
+
+    @api.model
+    def recurring_create_invoice(self, contract=False):
+        domain = []
+        date_ref = fields.Date.today()
```
sbidoul:
```suggestion
        date_ref = fields.Date.context_today()
```

contract/models/contract_line.py
```diff
@@ -0,0 +1,819 @@
...
+                # ('date_end', '=', False),
+                # ('date_end', '>=', date_ref),
+                # with this leaf, it's impossible to invoice the last period
+                # in post-paid case.
+                # i.e: date_end = 15/03 recurring_next_date = 31/03
+                # A solution for this, is to only check recurring_next_date
+                # and filter with create_invoice_visibility
+            ]
+        )
+        lines = self.search(domain).filtered('create_invoice_visibility')
```
sbidoul: If I understand well, this is going to return all contract lines with recurring_next_date in the past, and then filter on create_invoice_visibility, so this will get slower and slower as past lines accumulate.

Would it not be much simpler to set recurring_next_date to False when there is nothing more to invoice?

contract/models/contract_line.py
```diff
@@ -0,0 +1,819 @@
...
+        )
+        return invoice_line_vals
+
+    @api.multi
+    def _insert_markers(self, date_format):
+        self.ensure_one()
+        date_from = fields.Date.from_string(self.recurring_next_date)
+        date_to = date_from + self.get_relative_delta(
+            self.recurring_rule_type, self.recurring_interval
+        )
```
sbidoul, line 570: This is not a good place to compute date_from and date_to, they should be passed as parameter to this function, since they must have been computed elsewhere when deciding what we are invoicing.

The logic that computes the period being invoiced deserves it's how method, with unit tests.

contract/models/contract_line.py
```diff
@@ -0,0 +1,819 @@
...
+        """Create invoices from contracts
+
+        :return: invoices created
+        """
+        invoices = self.env['account.invoice']
+        for contract in self.mapped('contract_id'):
+            lines = self.filtered(lambda l: l.contract_id == contract)
+            invoices |= lines._create_invoice()
+            lines._update_recurring_next_date()
+        return invoices
```
sbidoul: I'm not sure this method will be efficient if we are invoicing many lines.

contract/models/contract_line.py
```diff
@@ -0,0 +1,819 @@
...
+
+    @api.multi
+    def _create_invoice(self):
+        """
+        :param invoice: If not False add lines to this invoice
```
sbidoul: the param does not exist

contract/models/contract.py
sbejaoui: I think it's not necessary,

contract/models/contract_line.py
```diff
@@ -0,0 +1,819 @@
...
+        )
+        name = self.name
+        name = name.replace('#START#', date_from.strftime(date_format))
+        name = name.replace('#END#', date_to.strftime(date_format))
+        return name
+
+    @api.multi
+    def _update_recurring_next_date(self):
+        for line in self:
+            ref_date = line.recurring_next_date or fields.Date.today()
```
sbidoul: why today?

contract/models/contract.py
sbejaoui: computed field are set to false by the model before even the call of the compute method.

contract/models/contract_line.py
```diff
@@ -0,0 +1,819 @@
...
+            old_date = fields.Date.from_string(ref_date)
```
sbidoul: from_string not needed in v12?

contract/models/contract_line.py
```diff
@@ -0,0 +1,819 @@
...
+                new_date_start,
+                rec.recurring_invoicing_type,
+                rec.recurring_rule_type,
+                rec.recurring_interval,
+            )
+            rec.date_end = (
+                rec.date_end
+                if not rec.date_end
+                else rec.date_end + delay_delta
+            )
```
sbidoul: This is harder to read than a plain old:

```python
if rec.date_end:
    rec.date_end = rec.date_end + delay_delta
```

contract/models/contract_line.py
sbejaoui: Right

contract/models/contract_line.py
```diff
@@ -0,0 +1,819 @@
...
+        """
+        Delay a contract line
+        :param delay_delta: delay relative delta
+        :return: delayed contract line
+        """
+        for rec in self:
+            old_date_start = rec.date_start
+            old_date_end = rec.date_end
+            new_date_start = rec.date_start + delay_delta
+            rec.recurring_next_date = self._compute_first_recurring_next_date(
```
sbidoul, line 653: Since this calls _compute_first_recurring_next_date, is it correct that we are not supposed to call delay if the first invoice has been sent?

contract/models/contract_line.py
```diff
@@ -0,0 +1,819 @@
...
+            return relativedelta(years=interval)
+
+    @api.multi
+    def delay(self, delay_delta):
+        """
+        Delay a contract line
+        :param delay_delta: delay relative delta
+        :return: delayed contract line
+        """
+        for rec in self:
```
sbidoul, line 644: First check that conditions to call delay are satisfied?

contract/models/contract_line.py
```diff
@@ -0,0 +1,819 @@
...
+        Put date_end on contract line
+        We don't consider contract lines that end's before the new end date
+        :param date_end: new date end for contract line
+        :return: True
+        """
+        if not all(self.mapped('is_stop_allowed')):
+            raise ValidationError(_('Stop not allowed for this line'))
+        for rec in self:
+            if date_end < rec.date_start:
+                rec.cancel()
```
sbidoul, line 675: This cancel() side effect is suprising, why is it here?

contract/models/contract_line.py
```diff
@@ -0,0 +1,819 @@
...
+            else:
+                old_date_end = rec.date_end
+                date_end = (
+                    rec.date_end
+                    if rec.date_end and rec.date_end < date_end
+                    else date_end
+                )
```
sbidoul: Maybe a question of taste but I think important business logic like this deserves an explicit if/then/else. Especially since we probably want to avoid write and message_post if the end date does not change.

contract/models/contract.py
sbidoul: Are you 100% sure of that?

contract/models/contract.py
sbejaoui: `all([])` return True, so we should check if date_end is not empty and don't contain False

contract/models/contract_line.py
```diff
@@ -0,0 +1,842 @@
...
+                        for name in lines.mapped('name')
+                    ]
+                )
+            )
+            contract.message_post(body=msg)
+        for rec in self:
+            if rec.predecessor_contract_line_id:
+                rec.predecessor_contract_line_id.successor_contract_line_id = (
+                    rec
+                )
```
sbidoul: add `assert not  rec.predecessor_contract_line_id.successor_contract_line_id`

contract/models/contract.py
sbejaoui: I know that by the use, but not 100% sure

rafaelbn:
Hi, thank you a lot both.

> So the tags and cost/revenue buttons can be seen on the native analytic account view only

OK. But Pedro know we have a customer which this case is a requsite so we will add it in other module in the fuure

> people managing contracts (usually sales persons)

Well Odoo don't think so has contact are in sales :smile: 

Thank again :+1:

contract/models/contract.py
sbejaoui: https://github.com/odoo/odoo/blob/12.0/odoo/fields.py#L1030

contract/models/contract_line.py
sbejaoui: User can empty this field and compute will be executed with false as value.

contract/models/contract.py
sbidoul: ok, thanks!

contract/models/contract_line.py
sbejaoui: yes,

contract/models/contract_line.py
sbejaoui: When a contract line is stopped before its start date, I think we should cancel it.

contract/models/contract_line.py
sbejaoui: It's handled by is_un_cancel_allowed

contract/models/contract_line.py
sbejaoui: I changed it with read_group

contract/models/contract_line.py
sbidoul: Yes, but it's hidden out of view. The assert here would help as a kind of documentation + safety net to make extra sure we do not overwrite the successor of the predecessor if one has been set in the meantime

rafaelbn (approved):

rafaelbn:
Thank! I know  codecov/patch — 87% of diff hit (target 97.6%)  it is not mandatory but I you think is affordable to increase coverage it could be great

Regards

pedrobaeza (changes requested):
I want to review it as it has a lot of changes that impact in our flows. Any way, test coverage has dropped a lot, so in this case it's blocking as I was keeping this to a sane 100% coverage before.

contract/models/contract.py
NL66278: I think it is needed. What if contract.recurring_next_date had a value and user cancels all lines? recurring_next_date variable will be empty, therefore contract will not be updated.

contract/models/abstract_contract.py
```diff
@@ -0,0 +1,74 @@
...
+# Copyright 2015 Pedro M. Baeza <pedro.baeza@tecnativa.com>
+# Copyright 2016-2018 Carlos Dauden <carlos.dauden@tecnativa.com>
+# Copyright 2016-2017 LasLabs Inc.
+# Copyright 2018 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, models, fields
+
+
+class AbstractAccountAnalyticContract(models.AbstractModel):
```
NL66278: Why the incredibly long name for class and model? I would prefer just AbstractContract and 'abstract.contract' (and the same goes for line). This would also give full consistency between python filename, model name and class name.

NL66278:
With this large reorganization, that is very welcome and addresses a lot of issues that I encountered with customers, why not go the extra mile and do away with some unfortunate designs that are inherited from the past:
1. Make contracts completely independent from analytic accounts (and name the model just contract);
2. Rename account.analytic.invoice.line to contract.line. I found it very confusing that these invoice lines are not actually invoice lines...;

So we would have:
- abstract.contract (.line)
- contract.template (.line)
- contract (.line)

3. Then give some fields shorter (and less confusing) names:
recurring_invoice_line_ids ==> contract_line_ids

sbidoul:
@NL66278 thanks for the review. Glad you like it.

We totally agree with your last mile suggestions. We started the split from analytic account by disconnecting the views already. We did not go further to have a first step that is easy on migration: we were a bit fearing the PR would be rejected if we went too far.

We are currently 1 week from go live on a project that generates 10 000 invoices per month with this module so we can't really change much right now.

Our idea is to have this one merged as 12.0.2.0.0, then we can help/work on a 3.0.0 (on 12 or 13) that does the full split from analytic account and the renaming you mention.

(BTW, @sbejaoui is currently working on a contract_forecast module that uses queue_job to maintain a list of future invoice lines each time a contract line change, for reporting purpose)

pedrobaeza:
:+1: for going with total isolation, but we need the migration scripts. Ping me for any doubt in how to do it.

contract/models/abstract_contract.py
sbejaoui: Agree,

sbejaoui: @sbidoul summarized why we are in this "just between" situation

contract/models/contract_line.py
```diff
@@ -512,31 +512,57 @@ def _prepare_invoice_line(self, invoice_id=False):
...
+    def _get_invoiced_period(self):
+        self.ensure_one()
+        dates = self._get_next_invoiced_period(
+            self.date_start,
+            self.date_end,
+            self.last_date_invoiced,
+            self.recurring_next_date,
+            self.recurring_interval,
+            self.recurring_rule_type,
+            self.recurring_invoicing_type,
```
sbidoul: @sbejaoui for the purpose of reusing in the forecasting algorihtm it would be slightly easier to read if _get_next_invoiced_period remains a multi/ensure_one method, taking only last_date_invoiced and recurring_next_date as arguments.

sbejaoui: yes

contract/models/contract_line.py
```diff
@@ -513,30 +513,44 @@ def _prepare_invoice_line(self, invoice_id=False):
...
+        return first_date_invoiced, last_date_invoiced, recurring_next_date
+
+    @api.multi
+    def _get_invoiced_period(self):
+        self.ensure_one()
+        dates = self._get_next_invoiced_period(
+            self.last_date_invoiced, self.recurring_next_date
+        )
+        first_date_invoiced, last_date_invoiced, recurring_next_date = dates
         return first_date_invoiced, last_date_invoiced
```
sbidoul: @sbejaoui another question: it's looks strange that `recurring_next_date` is computed by `_get_next_invoiced_period` then ignored in the return here, and computed (differently) elsewhere in `_update_recurring_next_date`. Also this `_get_invoiced_period` now looks a little bit useless. 

Perhaps you can push the refactoring further?

sbejaoui: yes, I was in a hurry to get out `_get_next_invoiced_period` to use it for the forecast algorithm. Unify the invoiced period compute is in my to do list.

contract/models/res_partner.py
```diff
@@ -8,35 +8,47 @@ class ResPartner(models.Model):
...
         for partner in self:
             partner_child_ids = partner.child_ids.ids + partner.ids
-            partner.sale_contract_count = sum([
-                r[2] for r in result
-                if r[0] in partner_child_ids and r[1] == 'sale'])
-            partner.purchase_contract_count = sum([
-                r[2] for r in result
-                if r[0] in partner_child_ids and r[1] == 'purchase'])
+            partner.sale_contract_count = sum(
+                [
```
MiquelRForgeFlow: I don't understand why you need to do this kind of cosmetics changes. This creates more difference, adds difficulty to trace the originator of these lines and there is not really any visual improvement.

sbejaoui: Sorry about that, I delegate code formatting to [balck](https://github.com/ambv/black) and it's automatically make this "cosmetics changes". I will try to perform a revert on this kind of changes to facilitate code review.

rafaelbn (commented):
@ThomasBinsfeld this improvements it changing the functionality so the Usage section of the readme should be updated. If not new reviewer don't know how to use it.

Also travis is :red_circle: , please could you check it?

Thanks a lot!

ThomasBinsfeld:
@rafaelbn Indeed. I'll make Travis green and coordinate with @sbejaoui to update the USAGE section.

AaronHForgeFlow:
IMO  the end date for the contract line should be computed when the line es created after onchange_template_id is called. Now the user has to manually add the date in the line even if the renewal period is set in the template:

image

@sbejaoui if you agree I can propose a PR to your branch

ThomasBinsfeld:
@rafaelbn I updated the USAGE section and Travis should be green now. I'll start the split from analytic account.

sbejaoui:
> IMO the end date for the contract line should be computed when the line es created after onchange_template_id is called. Now the user has to manually add the date in the line even if the renewal period is set in the template:
> 
> image
> 
> @sbejaoui if you agree I can propose a PR to your branch

I am not sure. As there is the use case of contract without predefined end date. auto_renew_rule_type is considered only if auto_renew is checked.

AaronHForgeFlow:
> I am not sure. As there is the use case of contract without predefined end date. auto_renew_rule_type is considered only if auto_renew is checked.

Yes I noticed that. What I can do is only to fill the end date when the auto_renew is set in the template. Is that fine?

sbejaoui:
> > I am not sure. As there is the use case of contract without predefined end date. auto_renew_rule_type is considered only if auto_renew is checked.
> 
> Yes I noticed that. What I can do is only to fill the end date when the auto_renew is set in the template. Is that fine?

I think it's the case with this onchange 

https://github.com/OCA/contract/blob/16c2cd75cfdc2e53902658bb904fa9ff62a05e32/contract/models/contract_line.py#L380-L396

AaronHForgeFlow:
> I think it's the case with this onchange

Thanks. https://github.com/sbejaoui/contract/pull/11

AaronHForgeFlow:
It seems there are conflicts in files

 ```
contract/i18n/de.po
contract/i18n/gl.po
```
@sbejaoui if you don't mind I can resolve them here.

**Comment by sbejaoui** [comment]
> It seems there are conflicts in files
> 
> ```
> contract/i18n/de.po
> contract/i18n/gl.po
> ```
> 
> @sbejaoui if you don't mind I can resolve them here.

yes sure

**Comment by rafaelbn** [comment]
@pedrobaeza please review to unblock this PR please, It's quite near to be merged

**Comment by MiquelRForgeFlow** [comment]
Could you wait until https://github.com/OCA/contract/pull/337 gets merged?

Also, please do a general rebase and squash, to reduce number of commits.

File: contract/models/res_partner.py [hunk 46]
```diff
@@ -8,35 +8,43 @@ class ResPartner(models.Model):
     _inherit = 'res.partner'
 
     sale_contract_count = fields.Integer(
-        string='Sale Contracts',
-        compute='_compute_contract_count',
+        string='Sale Contracts', compute='_compute_contract_count'
```
carlosdauden: Why this changes?

contract/views/contract.xml
```diff
@@ -0,0 +1,316 @@
...
+                <field name="pricelist_id"/>
+                <separator/>
+                <filter name="recurring_invoices"
+                        string="Recurring Invoices"
+                        domain="[('recurring_invoices','=',True)]"
+                />
+                <separator/>
+                <filter name="not_finished"
+                        string="In progress"
+                        domain="['|', ('date_end', '&gt;=', time.strftime('%Y-%m-%d')), ('date_end', '=', False)]"
```
carlosdauden:
```suggestion
                        domain="['|', ('date_end', '&gt;=', context_today().strftime('%Y-%m-%d')), ('date_end', '=', False)]"
```

contract/views/contract.xml
```diff
@@ -0,0 +1,316 @@
...
+                />
+                <filter name="finished"
+                        string="Finished"
+                        domain="[('date_end', '&lt;', time.strftime('%Y-%m-%d')), ('recurring_next_date', '=', False)]"
```
carlosdauden:
```suggestion
                        domain="[('date_end', '&lt;', context_today().strftime('%Y-%m-%d')), ('recurring_next_date', '=', False)]"
```

carlosdauden (approved):

AaronHForgeFlow (approved):
:+1: Functional tests

AaronHForgeFlow:
There is gonna be a minor conflict on the manifest version when merging #337  but it will be easy to solve, this module version number should be taken in this case

AaronHForgeFlow:
The payment terms are not passed from the partner to the contract. I created a PR for that https://github.com/sbejaoui/contract/pull/13

contract/models/contract.py
```diff
@@ -188,6 +188,12 @@ def _onchange_contract_template_id(self):
     def _onchange_partner_id(self):
         self.pricelist_id = self.partner_id.property_product_pricelist.id
         self.fiscal_position_id = self.partner_id.property_account_position_id
+        if self.contract_type == 'purchase':
```
nikul-serpentcs, line 210: @sbejaoui No Need `Else`
``
self.payment_term_id = \
                self.partner_id.property_payment_term_id
 if self.contract_type == 'purchase':
self.payment_term_id = \
                self.partner_id.property_supplier_payment_term_id
```

**Review by ThomasBinsfeld — APPROVED** [review]

**Comment by rafaelbn** [comment]
@ThomasBinsfeld could you please check the code suggestion of @carlosdauden please?

**Comment by rafaelbn** [comment]
@ThomasBinsfeld @sbidoul about 

> "End date on sale order line should recompute when start date changes etc?"

In the google doc. What do you think?

**Comment by ThomasBinsfeld** [comment]
> @ThomasBinsfeld could you please check the code suggestion of @carlosdauden please?

Done

**Comment by ThomasBinsfeld** [comment]
@pedrobaeza @rafaelbn @carlosdauden @aheficent @NL66278 @nikul-serpentcs @mreficent @gva-acsone I pushed the split from analytic account with migration scripts. The PR is now ready for a final review.

**Review by sbejaoui — COMMENTED** [review]
Thank you @ThomasBinsfeld , good work.
Code review for split commits, LGTM.

File: contract/models/contract.py [hunk 50]
```diff
@@ -0,0 +1,469 @@
...
+    contract_template_id = fields.Many2one(
+        string='Contract Template', comodel_name='contract.template'
+    )
+    recurring_invoice_line_ids = fields.One2many(
+        string='Invoice Lines',
+        comodel_name='contract.line',
+        inverse_name='contract_id',
+        copy=True,
+    )
+    recurring_invoices = fields.Boolean(
```
sbejaoui: @ThomasBinsfeld , this field is not needed anymore

contract/views/contract.xml
```diff
@@ -0,0 +1,308 @@
...
+        <field name="name">contract.contract search view (in contract)</field>
+        <field name="model">contract.contract</field>
+        <field name="arch" type="xml">
+            <search>
+                <field name="name"
+                       filter_domain="['|', ('name','ilike',self), ('code','ilike',self)]"/>
+                <field name="journal_id"/>
+                <field name="pricelist_id"/>
+                <separator/>
+                <filter name="recurring_invoices"
```
sbejaoui: This filter is not needed anymore

contract/models/contract_line.py
```diff
@@ -511,12 +514,10 @@ def _prepare_invoice_line(self, invoice_id=False):
         invoice_line._onchange_product_id()
         invoice_line_vals = invoice_line._convert_to_write(invoice_line._cache)
         # Insert markers
-        contract = self.contract_id
         name = self._insert_markers(dates[0], dates[1])
         invoice_line_vals.update(
             {
                 'name': name,
-                'account_analytic_id': contract.id,
```
sbejaoui: `'account_analytic_id': contract.analytic_account_id.id` ?

contract/models/contract.py
```diff
@@ -11,19 +11,36 @@
...
+    currency_id = fields.Many2one(
+        related="company_id.currency_id",
+        string="Currency",
+        readonly=True,
+    )
     contract_template_id = fields.Many2one(
-        string='Contract Template', comodel_name='account.analytic.contract'
+        string='Contract Template', comodel_name='contract.template'
     )
     recurring_invoice_line_ids = fields.One2many(
```
sbejaoui: what do you think if we rename this field to contract_line_ids?

ThomasBinsfeld: Good idea, this name is confusing.

contract/models/contract_line.py
ThomasBinsfeld: Analytic accounts on contract lines copied on the invoice lines is a future feature.

contract/models/contract_template.py
```diff
@@ -9,13 +9,13 @@
...
+    _name = 'contract.template'
+    _inherit = 'contract.abstract.contract'
+    _description = "Contract Template"
 
     recurring_invoice_line_ids = fields.One2many(
-        comodel_name='account.analytic.contract.line',
+        comodel_name='contract.template.line',
         inverse_name='contract_id',
         copy=True,
         string='Invoice Lines',
```
sbejaoui: `string='Contract Template lines'` ?

contract/views/contract_template_line.xml
```diff
@@ -1,17 +0,0 @@
-<?xml version="1.0" encoding="utf-8"?>
```
sbejaoui: I think you removed this file unintentionally.

contract/models/contract.py
```diff
@@ -11,19 +11,36 @@
...
         string='Invoice Lines',
```
sbejaoui: `string='Contract Lines',` ?

contract/models/contract_line.py
sbejaoui: so we will have analytic account in contract line level?

contract/views/contract_template_line.xml
ThomasBinsfeld: No, in my understanding, it was useless.

sbejaoui: Without this inheritance, contract template lines have no default form view.
If you take a look at runbot, when I add a line to a contract template, I get this view with all model fields.
Impossibility to save as contract template is required.

xavierjimenez (approved):
Have checked it functionally and everything seems to be working fine, great work @ThomasBinsfeld !

contract/views/contract_template_line.xml
ThomasBinsfeld: Ok! Thanks.

contract/migrations/12.0.2.0.0/pre-migration.py
```diff
@@ -0,0 +1,24 @@
...
+
+_logger = logging.getLogger(__name__)
+
+
+def migrate(cr, version):
+    """
+    set recurring_next_date to false for finished contract
+    """
+    _logger.info("order all contract line")
+    with api.Environment(cr, SUPERUSER_ID, {}) as env:
```
AaronHForgeFlow: I am getting an error when running the migration here: 

AttributeError: __exit__

But I am not sure what is wrong

contract/__manifest__.py
```diff
@@ -4,29 +4,36 @@
...
-    'version': '12.0.1.0.0',
+    'name': 'Recurring - Contracts Management',
+    'version': '12.0.4.0.0',
     'category': 'Contract Management',
     'license': 'AGPL-3',
     'author': "OpenERP SA, "
-              "Tecnativa, "
-              "LasLabs, "
-              "Odoo Community Association (OCA)",
+    "Tecnativa, "
```
pedrobaeza: Please don't de-indent this

contract/migrations/12.0.4.0.0/post-migration.py
```diff
@@ -0,0 +1,145 @@
...
+               create_date,
+               write_uid,
+               write_date
+        FROM account_analytic_account
+        WHERE recurring_invoices = TRUE
+        """
+    )
+    cr.execute(
+        """
+        INSERT INTO contract_line (
```
pedrobaeza: No need to do this, but only renaming the table and the link field column (in pre).

contract/migrations/12.0.2.0.0/pre-migration.py
```diff
@@ -0,0 +1,24 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+import logging
+
+from odoo import SUPERUSER_ID, api
+
+_logger = logging.getLogger(__name__)
+
+
+def migrate(cr, version):
```
pedrobaeza: Please use openupgradelib and decorate it with `@openupgrade.migrate()`. This way, you get a `env` proper populated.

You should also use `openupgrade.logged_query` method for registering the query and the result in the log, which is valuable.

contract/migrations/12.0.4.0.0/post-migration.py
```diff
@@ -0,0 +1,145 @@
...
+               manual_renew_needed,
+               active,
+               create_uid,
+               create_date,
+               write_uid,
+               write_date
+        FROM account_analytic_invoice_line
+        """
+    )
+    openupgrade.rename_models(cr, [('account.analytic.invoice.line',
```
pedrobaeza: Call also `rename_tables` and you don't need the insert SQL.

contract/migrations/12.0.4.0.0/post-migration.py
```diff
@@ -0,0 +1,145 @@
...
+                                    'contract.line')])
+    cr.execute(
+        """
+        DROP TABLE account_analytic_invoice_line
```
pedrobaeza: No need to do that with the renaming

contract/migrations/12.0.4.0.0/post-migration.py
```diff
@@ -0,0 +1,145 @@
...
+               payment_term_id,
+               fiscal_position_id,
+               invoice_partner_id,
+               message_main_attachment_id,
+               create_uid,
+               create_date,
+               write_uid,
+               write_date
+        FROM account_analytic_account
+        WHERE recurring_invoices = TRUE
```
pedrobaeza: You can remove `= TRUE` part

contract/migrations/12.0.2.0.0/pre-migration.py
```diff
@@ -0,0 +1,24 @@
...
+    """
+    set recurring_next_date to false for finished contract
+    """
+    _logger.info("order all contract line")
+    with api.Environment(cr, SUPERUSER_ID, {}) as env:
+        contracts = env["account.analytic.account"].search([])
+        finished_contract = contracts.filtered(
```
pedrobaeza: You can't do this on pre. See my alternative solution

contract/migrations/12.0.2.0.0/pre-migration.py
```diff
@@ -0,0 +1,24 @@
...
+            lambda c: not c.create_invoice_visibility
+        )
+        cr.execute(
+            "UPDATE account_analytic_account set recurring_next_date=null "
+            "where id in (%)" % ','.join(finished_contract.ids)
```
pedrobaeza: Replace this by `WHERE date_end IS NULL OR recurring_next_date <= date_end`

contract/models/contract.py
```diff
@@ -0,0 +1,470 @@
...
+        string='Fiscal Position',
+        ondelete='restrict',
+    )
+    invoice_partner_id = fields.Many2one(
+        string="Invoicing contact",
+        comodel_name='res.partner',
+        ondelete='restrict',
+    )
+    partner_id = fields.Many2one(
+        comodel_name='res.partner', inverse='_inverse_partner_id'
```
pedrobaeza: This is not a computed field. Why do you use an inverse?

contract/migrations/12.0.4.0.0/pre-migration.py
```diff
@@ -0,0 +1,72 @@
...
+         'contract.contract_line_user'),
+        ('contract.account_analytic_contract_line_manager',
+         'contract.contract_template_line_manager'),
+        ('contract.account_analytic_contract_line_user',
+         'contract.contract_template_line_user'),
+    ]
+    openupgrade.rename_models(cr, models_to_rename)
+    openupgrade.rename_tables(cr, tables_to_rename)
+    openupgrade.rename_xmlids(cr, xmlids_to_rename)
+    # A temporary column is needed to avoid breaking the foreign key constraint
```
pedrobaeza: This shouldn't be needed if you only rename the column

contract/migrations/12.0.2.0.0/post-migration.py
```diff
@@ -0,0 +1,32 @@
...
+        recurring_next_date=contract.recurring_next_date,
+        date_start=contract.date_start,
+        date_end=contract.date_end
+        FROM account_analytic_account AS contract
+        WHERE contract.id=contract_line.contract_id"""
+    )
+
+    _logger.info("order all contract line")
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    contract_lines = env["account.analytic.invoice.line"].search(
```
pedrobaeza: This will raise an error on current code state, as the model `account.analytic.invoice.line` is not existing anymore.

contract/migrations/12.0.2.0.0/post-migration.py
```diff
@@ -0,0 +1,32 @@
...
+
+    cr.execute(
+        """UPDATE account_analytic_invoice_line AS contract_line
+        SET recurring_rule_type=contract.recurring_rule_type,
+        recurring_invoicing_type=contract.recurring_invoicing_type,
+        recurring_interval=contract.recurring_interval,
+        recurring_next_date=contract.recurring_next_date,
+        date_start=contract.date_start,
+        date_end=contract.date_end
+        FROM account_analytic_account AS contract
```
pedrobaeza: As renaming is done in pre, you have to put here new table names

pedrobaeza (changes requested):
First review on migration scripts

contract_sale_invoicing/models/contract.py
```diff
@@ -28,5 +28,5 @@ def _create_invoice(self, invoice=False):
         if sales:
             invoice_ids = sales.action_invoice_create()
             invoice = self.env['account.invoice'].browse(invoice_ids)[:1]
-        return super(AccountAnalyticAccount, self)._create_invoice(
+        return super(ContractContract, self)._create_invoice(
```
pedrobaeza: You can take the occasion to update this to P3 syntax and avoid the class name.

contract_sale_invoicing/models/contract.py
```diff
@@ -16,9 +16,9 @@ class AccountAnalyticAccount(models.Model):
     @api.multi
     def _create_invoice(self, invoice=False):
         if not self.invoicing_sales:
-            return super(AccountAnalyticAccount, self)._create_invoice()
+            return super(ContractContract, self)._create_invoice()
```
pedrobaeza: You can take the occasion to update this to P3 syntax and avoid the class name.

contract/tests/test_contract.py
```diff
@@ -1,83 +1,129 @@
...
 
+from datetime import timedelta
+from dateutil.relativedelta import relativedelta
 from odoo import fields
 from odoo.exceptions import ValidationError
 from odoo.tests import common
 
 
+def to_date(date):
+    return fields.Date.to_date(date)
```
pedrobaeza, line 13: For not needing to define this method:
```python
from fields.Date import to_date
```

contract/models/res_partner.py
```diff
@@ -50,19 +58,20 @@ def act_show_contract(self):
             context=dict(
                 self.env.context,
                 search_default_recurring_invoices=True,
```
pedrobaeza: Remove this key when removing the filter

contract/migrations/12.0.4.0.0/post-migration.py
ThomasBinsfeld: I couldn't do a renaming of the contract lines table in the pre-migration script because of the foreign key on contracts (which can't be renamed since analytic accounts must be preserved). So I have to migrate contracts and their lines the same way.

contract/models/contract.py
ThomasBinsfeld: A field doesn't have to be computed to use an inverse on it. ping @sbejaoui

pedrobaeza (commented):
Some more comments.

You need also to manually updated some noupdate fields like cron record. I have detected these 3 at least:

- contract_cron_for_invoice
- email_contract_template
- rule_contract_template_multi_company

you can use `openupgrade.load_data` method for that.

contract/models/contract.py
pedrobaeza: Isn't this weird? For what is this intended?

pedrobaeza:
Can you rebase and solve conflicts after merging the minor change and what are the expected date for the changes we talked in OCA Spanish days?

rafaelbn:
@ThomasBinsfeld @sbidoul This branch has conflicts that must be resolved. After that could we merge? or you plan something else? thanks!

pedrobaeza:
Any news on this?

ThomasBinsfeld:
@pedrobaeza I should be able to finalize it this week.

pedrobaeza:
OK, thanks. Let me know when finished.

contract/tests/test_contract.py
ThomasBinsfeld: I'm afraid we can't to that.

`ModuleNotFoundError: No module named 'odoo.fields.Date'; 'odoo.fields' is not a package`

contract/models/contract.py
sbejaoui: IMO, we can remove the inverse method as `_inverse_partner_id` code is redundant with `_onchange_partner_id`

pedrobaeza: Better to keep inverse method, and on onchange, call to the inverse method. This way, you cover both programatical and UI update.

pedrobaeza:
Any news on this?

contract/migrations/12.0.2.0.0/pre-migration.py
sbejaoui: Hi @aheficent ,
I pushed a new version of the migration scripts. I think it resolve this issue

contract/migrations/12.0.4.0.0/post-migration.py
sbejaoui: @pedrobaeza,
As @ThomasBinsfeld mentioned, we can not rename the contract line table before we prepare the contract table.
How can we solve this problem?
I was looking for a way to create the contract table in the pre-migration script and insert the data there. Thus, the change of contract line name became possible.
I did not find an easy way to do it. In the end, it seems to me that the way things are done now is easy to understand if there is no decent alternative.

sbejaoui:
Hi,
I pushed a new version of the migration script:

- Move recurrence info from contract template to contract template lines
- Keep link between the contract lines and original analytic account
- Get rid of recurring_invoices filed on the contract
- Update no_update data after migration
- Move chatter and attachments from analytic account to contract

I did a lot of database migration tests from 12.0.1.0.0 to 12.0.4.0.0 and everything seems working fine.

@pedrobaeza , can review please.

pedrobaeza:
The main problem I see now is that you have migration scripts for 2 version (12.0.2.0.0 and 12.0.4.0.0). Are they compatible for being used in a v11 database?

sbejaoui:
> The main problem I see now is that you have migration scripts for 2 version (12.0.2.0.0 and 12.0.4.0.0). Are they compatible for being used in a v11 database?

Yea, I didn't totally think about that.
I made some checks on current version as we are using this PR in production and our current version is 12.0.3.0.0

But I believe that we can move the 12.0.2.0.0 into the 12.0.4.0.0. 
I will see for our installation if I keep the check on version number or made a patch for it.

sbejaoui:
> The main problem I see now is that you have migration scripts for 2 version (12.0.2.0.0 and 12.0.4.0.0). Are they compatible for being used in a v11 database?

I improved the way I check the current version number.
https://github.com/OCA/contract/blob/526860425e46ae18aae65f88e9d8b5532da42041/contract/migrations/12.0.4.0.0/post-migration.py#L213-L216

The 12.0.2.0.0 no longer contains post-migration script and the pre-migration uses sql query only.
So, it can be used in a v11 database

sbidoul:
@pedrobaeza I think all comments have been handled here.

rafaelbn:
This PRs is a historic one 😄

rafaelbn:
Please @sbejaoui could you rebsase please?

This branch has conflicts that must be resolved

Thanks!

sbidoul:
rebased

rafaelbn:
Please @sbidoul could you rebase again ? The are PR arround v12 branch that must take care of this PR and the could want to test this. 

By the way we should merge this :smile: deadline next OCA code sprint!

sbejaoui:
> Please @sbidoul could you rebase again ? The are PR arround v12 branch that must take care of this PR and the could want to test this.
> 
> By the way we should merge this  deadline next OCA code sprint!

Hi @rafaelbn ,
Rebase done. 
Appreciate your efforts.
We must take care of this one ASAP to resolve other pending PRs depending on it.

pedrobaeza:
Yes, I have been able to almost migrate our own database to v12 for trying this whole PR, as we use intensively contracts and thus we can test everything. I expect to be able to do it next week.

contract/migrations/12.0.4.0.0/post-migration.py
```diff
@@ -0,0 +1,229 @@
...
+            """
+            UPDATE account_invoice
+            SET old_contract_id = old_contract_id_tmp
+            """,
+        )
+        openupgrade.logged_query(
+            cr,
+            """
+            ALTER TABLE account_invoice
+            DROP COLUMN old_contract_id_tmp
```
pedrobaeza: Never drop any data

contract/migrations/12.0.4.0.0/post-migration.py
pedrobaeza: Never drop any data, and it fails due to constraints and foreign keys.

contract/migrations/12.0.4.0.0/post-migration.py
```diff
@@ -0,0 +1,229 @@
...
+        """
+        UPDATE account_invoice_line
+        SET contract_line_id = contract_line_id_tmp
+        """,
+    )
+    openupgrade.logged_query(
+        cr,
+        """
+        ALTER TABLE account_invoice_line
+        DROP COLUMN contract_line_id_tmp
```
pedrobaeza: Never drop any data. No matter is a temporary column you have created. This data can be needed for other dependent modules that needs the link for complementing some data.

contract/migrations/12.0.4.0.0/pre-migration.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    _logger.info(">> Pre-Migration 12.0.4.0.0")
+    cr = env.cr
+    openupgrade.logged_query(
+        cr,
+        """
+        DROP TABLE IF EXISTS account_analytic_invoice_line_wizard
```
pedrobaeza: Never drop any data

pedrobaeza (changes requested):
@sbejaoui I'm trying a migration and due to the droppings, there are errors. I can't push to your branch. Have you removed this ability? And you have a lot of commits that doesn't add too much value. Can you squash a bit commit history?

pedrobaeza:
I'm finding more problems, now:
```
  File "/opt/odoo/auto/addons/contract/migrations/12.0.4.0.0/post-migration.py", line 187, in _init_last_date_invoiced_on_contract_lines
    contract_lines._init_last_date_invoiced()
  File "/opt/odoo/auto/addons/contract/models/contract_line.py", line 619, in _init_last_date_invoiced
    if last_date_invoiced > rec.date_start:
UnboundLocalError: local variable 'last_date_invoiced' referenced before assignment
```
I have created a temp branch in my fork until I adjust everything and tell you back.

sbejaoui:
@pedrobaeza , you can push to my branch now

pedrobaeza:
OK, thanks! Can I squash commits by logic set/author?

Any way, I'm going to continue pushing to my local branch until the migration finishes for not hammering this PR.

sbejaoui:
> I'm finding more problems, now:
> 
> ```
>   File "/opt/odoo/auto/addons/contract/migrations/12.0.4.0.0/post-migration.py", line 187, in _init_last_date_invoiced_on_contract_lines
>     contract_lines._init_last_date_invoiced()
>   File "/opt/odoo/auto/addons/contract/models/contract_line.py", line 619, in _init_last_date_invoiced
>     if last_date_invoiced > rec.date_start:
> UnboundLocalError: local variable 'last_date_invoiced' referenced before assignment
> ```
> 
> I have created a temp branch in my fork until I adjust everything and tell you back.

I am pushing a fix for this.

contract/models/contract_line.py
```diff
@@ -0,0 +1,1105 @@
...
+                    )
+                )
+            elif rec.recurring_invoicing_type == 'post-paid':
+                last_date_invoiced = (
+                    rec.recurring_next_date
+                    - self.get_relative_delta(
+                        rec.recurring_rule_type, rec.recurring_interval
+                    )
+                ) - relativedelta(days=1)
+            if last_date_invoiced > rec.date_start:
```
pedrobaeza, line 621: This fails as there can be the option of `rec.recurring_invoicing_type = 'pre-paid'` and `recurring_rule_type != 'monthlylastday'`, but not sure about the strategy to fix it:

- Pre-filter before the records that comply with this.
- Set another `last_date_invoiced` in that cases.

@sbejaoui can you help me?

pedrobaeza:
> I am pushing a fix for this.

Great, because I have just asked you in a comment about how to fix it

contract/models/contract_line.py
sbejaoui: I just pushed a fix for it. 
for pre-paid contract, the last date invoiced is the day before the next invoice date.
for the post-paid, it's is the last day of the invoicing period before the next invoice date.

A special treatment for the `monthlylastday` as it's a special case of the post-paid method.

sbejaoui:
> OK, thanks! Can I squash commits by logic set/author?
> 
> Any way, I'm going to continue pushing to my local branch until the migration finishes for not hammering this PR.

Feel free please to push directly to the PR please. But let squash commits when you finish your review.

contract/readme/CONTRIBUTORS.rst
```diff
@@ -5,3 +5,5 @@
 * Vicent Cubells <vicent.cubells@tecnativa.com>
 * Miquel Raïch <miquel.raich@eficent.com>
 * Souheil Bejaoui <souheil.bejaoui@acsone.eu>
+* Thomas Binsfeld <thomas.binsfeld@acsone.eu>
```
rafaelbn, line 8: Hello, I should be as a contributor IMHO from long time ago (as least from v9). If you agree, please add me.

rafaelbn (dismissed):
I saw 2 things:

1) In the contract the button "Create invoice" is only show when debug mode is activated. I think i should be visible all the time.



2) The field reference (`code`) in contract should be populated to the field reference (`name`) in the invoice and not in the field "Payment ref." (`reference`), which it was used in v11 and before only for Vendors and now in v12 for a payment "Sequence" usualy really similar to the invoice number

contract/readme/CONTRIBUTORS.rst
sbejaoui: yes sure.

sbejaoui:
> I saw 2 things:
> 
> 1. In the contract the button "Create invoice" is only show when debug mode is activated. I think i should be visible all the time.
> 

It was already the case in all the previous versions and I think it's correct from an UX point of view.
IMO, the manual creation of invoices is not the normal case scenario and invoices should be created by the planned action. That action is meant to be used for debugging reasons.

But we can imagine a new security group "Can manually create invoices from contracts" to restrict the visibility of the button instead of using the debug mode. 

But still for me, the debug mode make sense to hide/show the create invoice button.

> 
> 
> 1. The field reference (`code`) in contract should be populated to the field reference (`name`) in the invoice and not in the field "Payment ref." (`reference`), which it was used in v11 and before only for Vendors and now in v12 for a payment "Sequence" usualy really similar to the invoice number
> 
> 

I get your point and it make sense for me. But it's the case in [12.0](https://github.com/OCA/contract/blob/12.0/contract/models/account_analytic_account.py#L232). Can you create an issue for this.

contract/migrations/12.0.4.0.0/post-migration.py
```diff
@@ -0,0 +1,210 @@
...
+    contract_line_fields.append('contract_id')
+    account_analytic_invoice_line_fields.append(contract_field_name)
+    contract_line_fields.append('analytic_account_id')
+    account_analytic_invoice_line_fields.append(contract_field_name)
+    contract_line_fields.append('active')
+    account_analytic_invoice_line_fields.append('true')
+
+    openupgrade.logged_query(
+        cr,
+        "INSERT INTO contract_line ("
```
pedrobaeza: @sbejaoui  what is the reason for inserting records instead of renaming the table? I'm just experimenting now with the renaming to see problems, but I don't see in theory anyone.

sbejaoui: As we are not renaming the `account_analytic_account` table, renaming the `account_analytic_invoice_line` table in pre-migration trigger foreign-key constraint to `contract_contract` table as it's not created yet.

With @ThomasBinsfeld , we think we had two solutions to resolve this.
1- Copy the create the `contract_contract` and copy in it the `account_analytic_account` table in the pre-migration script. Rename `account_analytic_invoice_line` table to contract_line.
2- Let odoo create the `contract_contract` and the `contract_line` tables and use insert to fill the data.

Do you think we can resolve the foreign-key constraint issue differently?

pedrobaeza: Yes, I think so. Usually the best option is to directly remove that FK constraint in pre and let Odoo to do rebuild it, but I'm trying right now. What are the benefits of not needed to move data?:

1. Speed: if you have a lot of records, insert operation takes time. As this is by SQL, it's not too much, but it's something.
2. Additional modules that added in v11 fields to this table, will need to create extra migration scripts for moving these fields values to the new table with this approach. If we simply rename the table, SQL columns will be there and then no migration script will be needed.
3. Avoid FK constraints.

What I'm going to try is to create table copying columns as is. Having extra columns there (coming from account_analytic_account table) is not problem, as they can be cleaned afterwards.

pedrobaeza:
I have finished the check and everything seems OK. A fine work, congratulations!

I have added the following changes, following the previous conversation about not to copy lines, but rename the table (a treacherous oldname was provoking an error without a hint of what was happening):

* Remove incorrect oldname attributes.
* Add filter on partners for running contracts (+ a support o2m field for that).
* Cover more tables in model renaming + cleaner code using a loop.
* Don't copy contract lines, but rename table + copy contract records on pre.
* Contract code is now populated to "Reference/Description" field in invoice.
* Order on new contract model has been restored to the same as old analytic accounts.
* Added Acsone as co-author (you deserve it!).

If you agree, I finally merge this PR.

pedrobaeza:
Ah, please squash as commented!

pedrobaeza (approved):

sbejaoui:
> Ah, please squash as commented!

Great job, thank you @pedrobaeza for the efforts.
I am squashing now.

sbejaoui:
@pedrobaeza , squash is done. PR is ready to merge.

pedrobaeza:
/ocabot merge

Hope to see you all on Odoo XP.

rafaelbn:
Hello @sbidoul , @sbejaoui !

Did you made this promise changes?

https://github.com/OCA/contract/pull/207#issuecomment-442759393

> We will add the UX feature to manage recurrence globally at the contract level a bit later, don't worry.

I see this is not available in v12 but it is in v13 and v14. And I cannot find where is that improvement in 13 or 14. I only know that it's not in 12 and today that contract module have FIX FIX FIX FIX and stills doesn't work propertly. Futhermore there are already BUGs in 13 and 14. https://www.loom.com/share/86fc343c704c42b6b5ab3c7b86828d29

PR #208, opened 2018-10-31
[12.0][IMP] - Create contract on sale order order confirmation

sbejaoui:
This PR meant to add this functionalities: 
- [x] Sale order lines contain invoicing recurrence details
- [x] Product has a new field is_contract applied only for type service
- [x] Contract products are linked to contract templates
- [x] On Sale Order confirmation, a contract is created for each contract template used on sale order lines
- [x] Contract product are ignored on invoicing process and pass to nothing to invoice
- [x] A not finished contract can be mentioned on sale order line
- [x] A sale order line linked to a contract will update it and don't create a new one if it had the same template

FIXME:
- [ ] remove change made in .travis.yml to get the #207 on build


product_contract/models/sale_order_line.py
```diff
@@ -0,0 +1,115 @@
...
+        default=1,
+        string='Repeat Every',
+        help="Repeat every (Days/Week/Month/Year)",
+        copy=False,
+    )
+    date_start = fields.Date(string='Date Start', default=fields.Date.today())
+    date_end = fields.Date(string='Date End', index=True)
+    recurring_next_date = fields.Date(
+        default=fields.Date.today(), copy=False, string='Date of Next Invoice'
+    )
```
sbidoul: What's the use case for having recurring_next_date on the so line?

product_contract/models/sale_order.py
```diff
@@ -0,0 +1,70 @@
...
+            rec.contract_count = len(rec.order_line.mapped('contract_id'))
+
+    @api.multi
+    def action_show_contracts(self):
+        self.ensure_one()
+        action = self.env.ref(
+            "contract.action_account_analytic_sale_overdue_all"
+        ).read()[0]
+        action["domain"] = [
+            ("id", "in", self.order_line.mapped('contract_id').ids)
```
sbidoul: I think it's better to search contracts that have a line linked to a sale order line of ours.

env['account.analytic.account'].search([contract_line_ids.recurring_invoice_line_ids.sale_order_line_id.order_id = self.id])

Such a link should be more robust.

sbidoul:
@sbejaoui we need have the reference to the so line on the contract line, and then propagating the so line to the invoice line.

product_contract/models/sale_order_line.py
sbejaoui: It is more user friendly if we can fill all needed information from the sale order.

But thinking about it now, recurring_next_date should have the same value as start_date in prepaid case and start_date + invoicing_interval for postpaid.

sbidoul: Yes, that's my understanding too: this field is not entered by the user but computed on the contract line, so it should not be on the so line.

gva-acsone:
Hi @pedrobaeza and @rafaelbn ,

This is the pull request that regroups most of the functionalities that we would like to implement in the 'OCA modules' related to contracts. We have a meeting with our customer end of this week to validate these changes. If you have some time this week, it would be really interesting to have your feedback on these features, as we would like to implement them in the OCA modules in the upcoming week(s).

As a reminder, the major changes are:

- Put the start/end dates and the recurrence mechanism (prepaid/post paid and recurrence interval) on the contract lines instead of the contract form.
- It is now possible to create a contract from a Sales Order, with contract products coming from both the SO and contract template (and not only from the contract template as it was the case before).
- Users can add SO lines that contain contract products in an existing (active) contract.
- The recurrence type (prepaid/postpaid) and the recurrence interval can be filled on the product form of contract products. These informations are copied on the SO line when the product is selected.
- It is now possible to select start/end dates on SO lines. These informations, as well as the recurrence type and the recurrence interval are copied from the SO line on the contract line once the SO is confirmed.
- Invoices for contracts are also available on the related Sales Order form.
- The date of next invoice is correctly updated once the recurrence interval 'Month last day' is selected.
- The invoicing algorithm has been adapted to keep up with all these changes

There are still 2 tasks we are working on:
- invoicing the last month for contract lines that are post-paid (actually not invoiced as the end date is prior to the last invoice date)
- adding the invoices related to contrats on the customers contact form.


Thanks for your feedback,

Guillaume

product_contract/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+{
+    'name': 'Recurring - Product Contract',
+    'version': '12.0.1.0.0',
+    'category': 'Contract Management',
+    'license': 'AGPL-3',
+    'author': "LasLabs, "
+              "ACSONE SA/NV, "
+              "Odoo Community Association (OCA)",
+    'website': 'https://github.com/oca/contract',
+    'depends': ['product', 'contract_sale'],
```
rafaelbn, line 14: New dependency with contract_sale

product_contract/models/product_template.py
```diff
@@ -0,0 +1,75 @@
...
+        """ Clear the relation to contract_template_id when downgrading
+        product from contract
+        """
+        if not self.is_contract:
+            self.contract_template_id = False
+
+    @api.constrains('is_contract', 'type')
+    def _check_contract_product_type(self):
+        """
+        Contract product should be service type
```
rafaelbn, line 56: Review, this is not working  

> Product has a new field is_contract applied only for type service

rafaelbn:
Hello, 

Thank you for this nice, great and huge work! :smile: 

Some comments from my side.

Let's say firstable that I've never understand why this product_contract should depends on contract templates. For me is an aditional complexity not needed by default. :cry: 

> Put the start/end dates and the recurrence mechanism (prepaid/post paid and recurrence interval) on the contract lines instead of the contract form

Please don't force to every one to define the recurrent mechanism line by line. This should be an option not the rule. This is not a new development, what happen with the customer using this module that only need to define one single recurring mechanism per contract and not by line? I think you should consider both options (maintain the existing + adding the improvements)




> The recurrence type (prepaid/postpaid) and the recurrence interval can be filled on the product form of contract products. These informations are copied on the SO line when the product is selected.

With this you are forcing to make 2 different products for pre-paid and postpaid. This in an error because same product can be selled to diferent customer with diferent invoicing type. Please don't make this filed mandatory.




> Put the start/end dates and the recurrence mechanism (prepaid/post paid and recurrence interval) on the contract lines instead of the contract form

Again the same. And what happens with the one who was using this option and the new way is not a option because is not needed and is and a extra job? IMHO I think this should be an option and we should work with product_contract without the details of line by line by default.


Take in care about this. If the contract template don't have defined a Journal (which is OK in my side) then when creating the contract from the sale order you should stablish the first sale journal and not leave empty

rafaelbn (commented):
I beg to consider a more generic and general approach for product_contract and after add more modules like:

- product_contract_invoice_type (allow to set the invoice type to the product for the customer who need this
- product_contract_dates_by_line
- etc....

If not this module is ALL-IN (all or nothing)

Thank you and sorry for the noise.

I will test more deeper as I've found use cases to consider but next time I will record a video

rafaelbn:
Please @sbejaoui could you give me the sector and product examples in your side of your customer? Just to put my self in you situation. I my side we have generally Service Consulting Companies, Software companies that sell product and services. Maintenance companies...

sbejaoui:
Hello @rafaelbn ,
thank you for your review, I will take your remark one by one.
To understand our use case, this is an example of contract that need recurrence mechanism on lines level:

- [01/01 - 12/31] - An annual subscription plan for website hosting
- [06/01 - 07/31] - An advertising campaign
- [01/01 - 31/03] - A discount of 30$ 
- [01/09 - 30/09] - A mass mailing campaign

I join @pedrobaeza in his idea, when he says, we can leave the recurring fields on both sides, contract and lines.
Visible by default at the contract level and used to fill lines level.

rafaelbn:
OK, thank you @sbejaoui . But again remember (1) you are not creating something from scratch you are improving something that already exists and is in use in production (2) if you don't get a more modular and generic approach you will have another customer with different requirement that this modules will not fit. Just trying to do something generic.

> I join @pedrobaeza in his idea, when he says, we can leave the recurring fields on both sides, contract and lines. Visible by default at the contract level and used to fill lines level.

OK, I agree

Thank you

rafaelbn:
Hello @sbejaoui 



- [ ] When a Product Type in Consumable the Check box "Is a contract" is available. Should be fixed? :thinking: 
- [ ] As I said we have here the same problem if we want to define one contract template with
  - Invoicing type
  - Repeat every
  - Etc...
We cannot because we have to define all that data product by product

Have do you planned all this staff in the migration from v10? Deleting all the fileds in the contract template and adding new fields in the product

Sorry, really this big changes are being hard, imagine for the customers. I wish we sill not have problems with them.

- [ ] Again if you don't delete the fields from the contract and the contract template and you don't make this new fields mandatory in the product than you can cover you uses cases and customer that don't need recurring by line maintain its functionality from v8/9/10/11

> I join @pedrobaeza in his idea, when he says, we can leave the recurring fields on both sides, contract and lines. Visible by default at the contract level and used to fill lines level.

This is important

Thank you!

sbejaoui:
@rafaelbn ,

> * [ ]  When a Product Type in Consumable the Check box "Is a contract" is available. Should be fixed? thinking

Yes, we can do that
> * [ ]  As I said we have here the same problem if we want to define one contract template with
>   
>   * Invoicing type
>   * Repeat every
>   * Etc...
>     We cannot because we have to define all that data product by product

Fields on contract template are default value used to automatically fill on product change. A kind of helper.
You are right, this filelds should not be required.
 
> Have do you planned all this staff in the migration from v10? Deleting all the fileds in the contract template and adding new fields in the product

Fields still exist on contract template, but at lines level and we don't need to bring data to product template.
As discussed, at the end of this work, we will bring recurring fields back to contract and contract template level, and we should have the same functional flows as the actual version but with the possibility to chose to manage contracts globally or line by line.

> Sorry, really this big changes are being hard, imagine for the customers. I wish we sill not have problems with them.

I understand your concern, and we will work it together to have a more generic module that fit customers needs. 

> * [ ]  Again if you don't delete the fields from the contract and the contract template and you don't make this new fields mandatory in the product than you can cover you uses cases and customer that don't need recurring by line maintain its functionality from v8/9/10/11

I agree,

Thank you!

rafaelbn:
Thank you to you! :heart:

AaronHForgeFlow (commented):
It works good, great feature!

Just a minor comment. If the product in the sale order line is a contract. Then the contract_id field should be readonly to prevent users to fill it. contract_id will be filled on SO confirmation so it is fine.

sbejaoui:
> It works good, great feature!
> 
> Just a minor comment. If the product in the sale order line is a contract. Then the contract_id field should be readonly to prevent users to fill it. contract_id will be filled on SO confirmation so it is fine.

User can set contract_id if he want to add a sale line to an existing contract. Contract are filtered by customer and contract template.

A user can also set a contract line in that contract so the system will stop the target at the start of the new one (cases of upsell/downsell)

sbejaoui:
In version 11.0, product_contract only adds the creation of a contract from a sales order based on a contract template linked to a product without considering the sale order line it self.

With this PR we are improving and adding many features to the sales order/contracts relationship, such as upsell and downsell, converting orders into contracts and handling invoiced quantity for converted lines.

In the other hand, contract_sale module, add the contract menu to the sales application. 

When I was working on this PR, I always wondered why we keep two separate modules with the same purpose (introduce contract management into the sales application).

So isn't the time now to merge them in one module (contract_sale), or can you see a reason why we should keep them separate?

AaronHForgeFlow:
> User can set contract_id if he want to add a sale line to an existing contract. Contract are filtered by customer and contract template.
> 
> A user can also set a contract line in that contract so the system will stop the target at the start of the new one (cases of upsell/downsell)

Yes I agree. Please dismiss my comment

AaronHForgeFlow (approved):

hparfr:
I did some functional tests, it looks goods.

Filling date start and date end are painful to enter for users and error prone if they are the same for all the lines.

Last remark, is it possible to rename ['sale.order.line'].date_start  and ['sale.order.line'].date_end to something more obvious like contract_date_start and contract_date_end ?

rafaelbn:
i set WIP because depends on https://github.com/OCA/contract/pull/207 and bots are :red_circle:

sbejaoui:
Hello @rafaelbn ,
I made the rebase. This PR is ready to be reviewed.

ThomasBinsfeld (approved):

ThomasBinsfeld:
Conflicts resolved (on .travis.yml)

gva-acsone (approved):
Functional test + 1

hparfr:
A little late on this point, but @sbejaoui, have you considered to depend of this module https://github.com/OCA/sale-workflow/blob/12.0/sale_start_end_dates ?

sbejaoui:
> A little late on this point, but @sbejaoui, have you considered to depend of this module https://github.com/OCA/sale-workflow/blob/12.0/sale_start_end_dates ?

Hi @hparfr ,

The same question was raised by @sbidoul [here](https://github.com/OCA/sale-workflow/pull/943#issuecomment-539883414).

I think we should do it. 

This improvement worth a git issue: #395 395.

hparfr:
Hi. My bad,  I didn't see it :)
Thanks

PR #210, opened 2018-11-06
Update pt_PT.po

mybengui:
Complete translation of module


pedrobaeza:
Please use Weblate for doing translations: https://odoo-community.org/page/translate

PR #212, opened 2018-11-07
[12.0][MIG] Contract show invoice

sbejaoui:


sbejaoui:
FYI @gva-acsone

gva-acsone (approved):
Functional test 👍

pedrobaeza:
Sorry for being late on this, but this module is a bad understanding / migration of a feature that shouldn't be in this repo.

Let's see a bit the history: on v8, we have the module `contract_show_recurring_invoice` (https://github.com/OCA/contract/tree/8.0/contract_show_recurring_invoice) and the module `contract_show_invoice` (https://github.com/OCA/contract/tree/8.0/contract_show_invoice). The first one shows the invoices created from contracts with recurring invoices, and the second one show the invoices that have the analytic account of the contract. And they were 2 independent modules as in Odoo the recurring invoicing feature was part of the core, but the second one had the wrong name and repo.

When we migrate to v9, we merge this feature of `contract_show_recurring_invoice` in the main `contract` module, as Odoo removes the feature from the core and we adopt it under that new name, but the other module was kept with the same name and in the same repo.

And this has been until today due to laziness / people keeping migrating that module. We have already discussed in previous migration PRs to move this module out of the repo, but not done. The idea is to rename to `account_analytic_show_invoice` and put it on OCA/account-analytic, but the main thing is TO NOT REMOVE "show invoices" feature from contract.

sbejaoui:
Thank you @pedrobaeza for your feedback.
IMHO, I think we should change the way the contract form view is defined by creating a new one not inherited from account.analytic one and add this button in the contract base module.

sbidoul:
@pedrobaeza do you see objections to creating dedicated views for contracts, instead of inheriting from the analytic account views (of which we use next to nothing, except the name)

pedrobaeza:
@sbidoul do you mean for `contract` module of for this one? As stated, I think this module shouldn't be here, but on OCA/account-analytic, and I still think there should be smart-button or link to recurrent invoices on the contract.

If you are talking about using dedicated views for contracts, primary views are more or less that, but the idea was to have added any extra option on analytic account also on contracts, but it's not a great deal. What I'm more and more convinced is to detach contracts out of analytic accounts and to be an independent entity. We can link analytic accounts to entire contract/contract lines of course and maybe provide facilities for creating that analytic account, but not being the same thing.

sbejaoui:
I will start by detaching the contract form view from the account analytic one on PR #207.
It make more sens to separate them as they are used in very different functional processes.

rafaelbn:
Hello, @ThomasBinsfeld @sbidoul  I'm testing https://github.com/OCA/contract/pull/207 and Contract already show with a smartbutton the invocices, so why is this module needed? :thinking:

sbejaoui:
> Hello, @ThomasBinsfeld @sbidoul I'm testing #207 and Contract already show with a smartbutton the invocices, so why is this module needed?

Hi @rafaelbn ,
It's not needed any more. We have a dedicated form view for contract in #207 with a smart button to invoices and there is no inheritance between contract analytic account  form views.

ThomasBinsfeld:
@sbejaoui This PR can be closed once #207 is merged then?

sbejaoui:
> @sbejaoui This PR can be closed once #207 is merged then?

according to @pedrobaeza , we shouldn't migrate this module, so I think we can close it even before. 

> And this has been until today due to laziness / people keeping migrating that module. We have already discussed in previous migration PRs to move this module out of the repo, but not done. The idea is to rename to `account_analytic_show_invoice` and put it on OCA/account-analytic, but the main thing is TO NOT REMOVE "show invoices" feature from contract.

PR #215, opened 2018-11-20
[10.0][ADD]contract_variable_discount

AaronHForgeFlow:
With this module, you will be able to define in recurring contracts some lines with variable discount according to a provided formula.
The most typical case is the partner meets some condition to have a discount for a period of time: (unemployment, student, retired).
If the partner loses that condition then the discount is not applied or a different discount is applied for the next invoices.


contract_variable_discount/tests/test_contract_variable_discount.py
```diff
@@ -0,0 +1,77 @@
...
+            'pricelist_id': self.partner.property_product_pricelist.id,
+            'recurring_invoices': True,
+            'date_start': '2016-02-15',
+            'recurring_next_date': '2016-02-29',
+            'contract_template_id': self.template.id,
+        })
+        self.formula = self.env['contract.line.discount.formula'].create({
+            'name': 'Test formula',
+            # For testing each of the possible variables
+            'code': 'result = 15',
```
sbejaoui: We should test all  variables 
```python
self.formula = self.env['contract.line.discount.formula'].create({
            'name': 'Test formula',
            'code': 'env["res.users"]\n'
                    'context.get("lang")\n'
                    'user.id\n'
                    'contract.id\n'
                    'partner.id\n'
                    'line.discount_type\n'
                    'result = 15',
        })
```

contract_variable_discount/__manifest__.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
+# © 2016 Pedro M. Baeza <pedro.baeza@tecnativa.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    'name': 'Variable discount in contract recurrent invoicing',
+    'version': '10.0.1.2.0',
```
sbejaoui: 'version': '10.0.1.0.0',

sergio-teruel: @sbejaoui Please review oca documentation about version numbers here https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#version-numbers
I think that the version is 10.0.1.1.0

contract_variable_discount/models/contract.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+class AccountAnalyticContractLine(models.Model):
+    _inherit = 'account.analytic.contract.line'
+
+    discount_type = fields.Selection(
+        selection=[
+            ('fixed', 'Fixed discount'),
+            ('variable', 'Variable discount'),
+        ], required=True, default='fixed', string="discount. type")
```
sergio-teruel: remove dot after discount string

contract_variable_discount/__manifest__.py
sbejaoui: Hi @sergio-teruel ,

Thank you for the hint,

As this is the first release for the module, I suggested  `10.0.1.0.0` according to OCA guideline

PR #218, opened 2018-11-28
[WIP][FIX] Prefetch some fields into cache to speed up operations. 

daramousk:
This should reduce the runtime to 1/3 of the original.
cc @NL66278

### DIFF
--- a/contract/models/account_analytic_account.py
+++ b/contract/models/account_analytic_account.py
@@ -164,6 +164,11 @@ class AccountAnalyticAccount(models.Model):
 
     @api.model
     def _insert_markers(self, line, date_start, next_date, date_format):
+        self._read_from_database([
+            'recurring_rule_type',
+            'recurring_interval',
+            'recurring_invoicing_type',
+            'partner_id'])
         contract = line.analytic_account_id
         if contract.recurring_invoicing_type == 'pre-paid':
             date_from = date_start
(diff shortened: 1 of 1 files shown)


rafaelbn:
Related? https://github.com/OCA/contract/pull/240

yajo:
Hi there.

AFAIK, prefetching is an automatic action performed by the ORM itself, so prefetching by hand doesn't seem like adding much benefit, except in the case where the model has a lot of fields and we only want to fetch some of them. Even in that case, usually what takes most time is applying the filters in the database or computing computed fields, not actually fetching the raw fields.

So, it is good to optimize, but I think we need some explanation about _why and how_ is this reducing that amount of time. If we don't know that, then a future patch or review could remove or inhibit these benefits.

daramousk:
Hello Jairo and thanks for taking time to reply,

When we loop over records and ask for fields of them there seems to be always a cache miss which turns to a read from database *for every field of every record*.

While the computed fields will be in cache, will the rest of the fields be?

Maybe I have misunderstood how `_cache` works or is there a decorator I am missing?
@Yajo

yajo:
> While the computed fields will be in cache, will the rest of the fields be?

Yes, all of them will be in cache. That's what cache is for, after all.

To understand fine the prefetching mechanism, I think the best you can do is go see section 3 of this talk: https://youtu.be/dQYGau81RYg?t=1638

Basically, the batching of operations serves the purpose of reducing calls to database backend. However, when a model is too bloated, indeed it's possible that too much prefetching kills performance.

The problem with your approach is that:

- What happens if some other module overrides one of those fields you're reading from database and converts it into a computed, non-stored field? It would make it probably fail, given that it cannot be read _from database_ anymore.
- What happens if another module overrides one of these methods and it uses another field? That'd be triggering again the prefetch of everythin, and this performance improvement would be lost.

So, the common pattern when prefetching shoots you in the toe is one of:

- Reducing batches (slice).
- Disabling prefetch (`.with_context(prefetch_fields=False)`)

Go see the video linked above, and let's see if there's a chance to increase performance with a "more standard" system. :wink:

daramousk:
Ah I see. In any case since the other PR was merged I will close this.

Thanks for taking time.

Issue #216, opened 2018-11-23
agreement 11.0.0.1.0


max3903:
# agreement.recital
- [x] Add a agreement.recital object
- [x] Add a required char field "Name" (name)
- [x] Add a char field "Title" (title)
- [x] Add a integer field "Sequence" (sequence)
- [x] Add a html field "Content" (content)
- [x] Allow the content to include values from the agreement
- [x] Add a many2one field "Agreement" (agreement_id)
- [x] Add a menu item "Recitals" in Master Data > Content

# agreement.appendix
- [x] Add a agreement.appendix object
- [x] Add a required char field "Name" (name)
- [x] Add a char field "Title" (title)
- [x] Add a integer field "Sequence" (sequence)
- [x] Add a html field "Content" (content)
- [x] Allow the content to include values from the agreement
- [x] Add a many2one field "Agreement" (agreement_id)
- [x] Add a menu item "Appendices" in Master Data > Content

# agreement.section
- [x] Allow the content to include values from the agreement

# agreement.clause
- [x] Allow the content to include values from the agreement

# agreement.serviceprofile
- [x] Add a agreement.serviceprofile object with a required name field
- [x] Add a many2one field "Agreement" (agreement_id) to agreement

# Agreement
- [x] Add a one2many field "Service Profiles" (serviceprofile_ids) to agreement.serviceprofile
- [x] Add a one2many field "Recitals" (recital_ids) to agreement.recital and display the sequence, name and title
- [x] Add a one2many field "Appendices" (appendix_ids) to agreement.appendix and display the sequence, name and title
* Modify the pdf report to have:
  - [x] the dynamic content replaced by the values of the agreement
  - [x] the recitals (title and content) displayed before the sections
  - [x] the appendices (title and content) after the signature with break pages before each appendix

PR #219, opened 2018-11-28
[12.0][MIG] agreement

ygol:


ygol:
Hi @max3903.
1. Regarding Issue #216: Wouldn't be better to first migrate and then add those new fields?
2. I created 1 PR per module. Here's the first module.
3. There's an error in the log:

> Invalid view Agreement Kanban definition in agreement/views/agreement.xml

but the kanban seem to function alright. There are also few warnings about same labels used twice.
kindly advise

max3903:
Hello @ygol,

1. I have no date yet as to when we will add those fields in v11, so yes we can proceed that way.
2. :+1: 
3. Please look at https://travis-ci.org/OCA/contract/jobs/460800412#L1211. Regarding the labels, I don't have a better option. The description warnings can be easily fixed.

Thank you for your patience.

max3903:
@ygol We need the original history from v11.

Please have a look at the documentation on how to do that:
https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-12.0#howto

Thanks!

ygol:
> 
> 
> @ygol We need the original history from v11.
> 
> Please have a look at the documentation on how to do that:
> https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-12.0#howto
> 
> Thanks!

so I close this PR and start over?

max3903:
@ygol You can keep this PR: while you go through the steps, you can push to that same branch.

Issue #192, opened 2018-10-11
Agreement


max3903:
We would like to publish a module to manage content for agreements and contracts. The module is meant to be used by the legal team of a company and to allow them to define sections, clauses and template with their respective content that can be dynamic. Based on the template, an agreement can be created and the pdf document generated. The agreement would go through a workflow to finally become a contract with the customer signature.

Let us know if this repository is not the right place.

pedrobaeza:
I think this one is correct. For the signing, we already have the module contract_digitized_signature.

bealdav:
Hi @max3903,

I've just seen this issue now, and also I see a module named agreement in v11.

I'm very surprised, that you haven't contacted Akretion (@alexis-via or me) about agreement_account and agreement_sale module done in v10.

These modules are really small and easy to override. It could have been a good way to complete them instead of define other implementation or at least ping us to see how to work together on these concepts.

Maybe we can talk to define how to implement the best modularity, on this matter with an RFC ?

Thanks

max3903:
Hi @bealdav,

I understand your frustration. The module we shared was done by our customer as a prototype. The project is not over yet so we still have the opportunity to converge.

bealdav:
Happy to hear that. Maybe we can open an RFC to define changes in v12 branch ? V11 can stay as it.

Thanks

max3903:
Sure!

PR #220, opened 2018-12-04
[MIG] agreement -> agreement_legal

ygol:


ygol:
@max3903 what is next step?

max3903 (approved):
LGTM. No Test.

bealdav:
@ygol Thanks for your contribution.

The first implementation of agreement is in v10 here

https://github.com/OCA/contract/blob/10.0/agreement_account/models/agreement.py

The feature here is really, really small and serves many options to deal with agreement (on large context, not only big ones).

I ok to remove `account` dependency of agreement object. That make sense (thanks)

Since v11 new needs are addressed by @max3903 for complex workflows on this topic.

But to make it modular, I think we should keep agreement module as small as `agreement_account` in v10 and add new concepts in other modules.

I think, you should port agreement_account (becomes agreement) with its commit and authoring from v10 to v12 (it may be fair play or I can do it, if you want) and propose another PR to add other modules to deal with all your concepts.

What do you think ? 

cc @alexis-via

bealdav (changes requested):
Require less features in the first module agreement as explained

bealdav:
FYI `agreement_account` is used in 

https://github.com/OCA/l10n-france/blob/10.0/l10n_fr_chorus_account/__manifest__.py#L17

ygol:
> 
> 
> @ygol Thanks for your contribution.
> 
> The first implementation of agreement is in v10 here
> 
> https://github.com/OCA/contract/blob/10.0/agreement_account/models/agreement.py
> 
> The feature here is really, really small and serves many options to deal with agreement (on large context, not only big ones).
> 
> I ok to remove `account` dependency of agreement object. That make sense (thanks)
> 
> Since v11 new needs are addressed by @max3903 for complex workflows on this topic.
> 
> But to make it modular, I think we should keep agreement module as small as `agreement_account` in v10 and add new concepts in other modules.
> 
> I think, you should port agreement_account (becomes agreement) with its commit and authoring from v10 to v12 (it may be fair play or I can do it, if you want) and propose another PR to add other modules to deal with all your concepts.
> 
> What do you think ?
> 
> cc @alexis-via

I think @max3903 is the one who must reply here because this refactoring will break his upgrade path of his own v11 agreement module

max3903:
@bealdav @ygol What do you think of:

agreement_base in v12 = agreement_account in v10 without the account dependency
agreement in v12 (depends on agreement_base) = agreement in v11

ygol:
> 
> 
> @bealdav @ygol What do you think of:
> 
> agreement_base in v12 = agreement_account in v10 without the account dependency
> agreement in v12 (depends on agreement_base) = agreement in v11

@bealdav I started migrating agreement_account as agreement_base without the account dependency but the README.rst should be re-written if all the account dependencies are removed. 

It does make sense to migrate it with ``` its commit and authoring from v10 to v12``` so @bealdav; would you go ahead and migrate agreement_account as agreement_base 

I will then refactor my v12 agreement to inherit your new agreement_base module.

bealdav:
@ygol Thanks. We really need @alexis-via here.
From your side you've a project v12 with these modules ?

ygol:
> 
> 
> @ygol Thanks. We really need @alexis-via here.
> From your side you've a project v12 with these modules ?

not anymore as I couldn't wait and wrote what I needed from scratch but if possible I will switch to it once it is ready.

agreement/models/agreement.py
```diff
@@ -0,0 +1,401 @@
...
+        string="Approved By",
+        track_visibility='onchange'
+    )
+    currency_id = fields.Many2one(
+        'res.currency',
+        string='Currency'
+    )
+    partner_id = fields.Many2one(
+        'res.partner',
+        string="Partmer",
```
alexis-via, line 164: Typo : m -> n

ygol: Hi @alexis-via this module "agreement" will have to be refactored after agreement_base migration.
ref. https://github.com/OCA/contract/pull/220#issuecomment-452285572
tnx

alexis-via:
Sorry for this long silence ; there is my point of view.

First, let me sum up the history. In July 2017, I sent a mail in contributors@odoo-community.org about my need for a new module "Contract without analytic", cf https://odoo-community.org/groups/contributors-15/contributors-60794
The discussion resulted in PR #95 that I created in August 2017 for v10 and introduced 2 modules:
- agreement_account, which only depends on "account" ; 76 lines of python
- agreement_sale, which only depends on "sale" ("sale_commercial_partner" in fact) and "agreement_account" ; 50 lines of Python code
These modules use the "agreement" object that has about 10 fields (name, code, agreement_type, partner_id, company_id, active, signature_date, out_invoice_ids, in_invoice_ids, sale_ids). These module have 2 particularities:
- not related to analytic
- the agreement object is linked to invoice (NOT to invoice lines) and to sale order (NOT to sale order line)
We use these small agreement modules for several different customer projects where some custom modules add many other specific fields on the agreement object) and we also use these modules in the l10n_fr_chorus_account module of the OCA (cf https://github.com/OCA/l10n-france/blob/10.0/l10n_fr_chorus_account/__manifest__.py#L17).

In October 2018, @max3903 proposed in this PR https://github.com/OCA/contract/pull/194 for v11 a module named "agreement" (name quite similar in odoo v11 that uses an object called "agreement" (same name), which is NOT based on the agreement modules of Odoo v10. This module has 807 lines of Python code and proposes several objects in additional to the "agreement" object (agreement.appendix, agreement.clause, agreement.increasetype, agreement.line, agreement.recital, agreement.renewaltype, agreement.section, agreement.serviceprofile, agreement.stage, agreement.subtype, agreement.type). This PR was merged in the v11 branch in 5 days without any notification for the authors of the agreement modules of v10. I think it's something that should not happen, but we are now there... As you may know, Akretion doesn't work on v11, but we work on v12 so we need to find a solution in v12.

I see 2 options:
1) collaborate and have a common code base with dependancies
2) have 2 different sets of modules (would imply renaming one of the sets of modules and renaming the "agreement" object in one of them to avoid model name conflicts)

Let's study the option 1 for now, and if we really don't find any clever way to collaborate, we would switch to option 2.

What I propose for option 1 for v12 :
- we port our modules agreement_account and agreement_sale to v12 with minimum modification: these modules stay simple, without analytic 
- your modules are renamed with the "legal_" prefix and depend on our modules

At the end, we would have the following modules:
- "agreement_account": only depends on "account" ; defines the "agreement" objects with some basic fields and invoices have a M2O link to it
- "agreement_sale": only depends on "sale" and "agreement_account" ; adds a M2O link from sale.order to agreement
- "legal_agreement_account" : depends on "agreement_account" and adds many fields on the "agreement" object and also adds several objects (agreement.appendix, agreement.clause, agreement.increasetype, etc...)
- "legal_agreement_sale" : depends on "legal_agreement_account" and "agreement_sale"

In this scenario, the things to discuss are:
a) in the current "agreement" module of v11, the "agreement" object has a M2O field to sale.order, whereas in the "agreement_sale" module, the M2O link is from sale.order to agreement (the other way around !)
b) the "agreement" module of v11 has a M2O "agreement_type_id" field ; in our "agreement_account" module, we have a selection field "agreement_type" with 2 possible values "sale" and "purchase". I feel it serves 2 very different purposes, so maybe we could keep both.
c) the field "approved_date" of the "agreement" v11 module and the field "signature_date" of our module "agreement_account" could probably be merged
d) the fields "start_date" and "end_date" are very generic and could be moved from to the base "agreement_account" module

pedrobaeza:
I don't think all the modules should be kept, but they should be merged. @alexis-via you have proposed your easiest way, but not the best one.

ygol:
@alexis-via Is @max3903 proposition (https://github.com/OCA/contract/pull/220#issuecomment-449972472) wouldn't work for you? 
we could have a common agreement_base (without the dependency to account which would be handled by a new agreement_account extending agreement_base and re-introducing the dependency to account)
is this understandable :) ?

pedrobaeza:
I would remove `_base` from `agreement_base` and let only `agreement`.

alexis-via:
@pedrobaeza I can understand that you consider that my proposal is the easiest way for me. But you also have to consider the development of @max3903 with the "agreement" module in v11 blew up our work (and cause big problems when we'll port the OCA module l10n_fr_chorus_account), and we never asked for that. The proposal for this new "agreement" module was made by @max3903 on October 12 in this issue https://github.com/OCA/contract/issues/192 (without any ping for me, the author of agreement_account and agreement_sale), then the PR https://github.com/OCA/contract/pull/194 was made the next day by @max3903  (without any ping for me) and he auto-merged himself without any review 5 days later. I guess that @max3903 was not aware of the existence of agreement_sale and agreement_account on v10, that's why he auto-merged himself considering that nothing was at stake and nobody cared. Otherwise, if he knew the existence of agrement_sale and agreement_account, it means he consciously super-seeded our agreement modules by-passing a peer review that would have obviously raised compatibility concerns over module name and model name... and that would be a very a bad attitude (that's why I think he was just un-aware of the existence of agreement_account and agreement_sale).

@ygol No, Maxim's proposal doesn't work for me:
- removing the dependency on the "account" module is not interesting for me. In fact, I don't imagine a project managing such agreements without the account module. And I need the M2O link from agreement to invoice.
- I need the simple "agreement_sale" module

max3903:
@alexis-via I was not aware of the modules in v10 (when I created the issue on Oct 12th) until @bealdav notified me on November 8th https://github.com/OCA/contract/issues/192#issuecomment-436974362.

I proposed the module the next day because the module was already done by our customer as a prototype. You will agree with me that this is a rare case. You can look at our other improvements, they have gone through issues and theirs pull requests came 1 or 2 weeks after (time for our customer to review the specifications and approve + time for us to assign the development to a developer + time for the developer to develop, push and create the pull request).

Finally, I auto-merged it because it's a beta module.

My module set is meant to be used by legal teams to manage content for agreement, contract, letter of intent, etc. They don't care about invoicing and accounting. They care about content revisioning, reviewing and getting a signed document when a sales order or purchase order is confirmed.

Would you agree to split your module into:
* `agreement` (depends on base) and
* `agreement_account` (depends on agreement and account)

so I can migrate my modules with a dependency on `agreement` and rename my `agreement` to `legal_agreement`?

max3903:
@alexis-via We owe a solution to @ygol who is generously contributing the migration here...

alexis-via:
Your proposal seems good to me. I'll try to write a more detailed version of your proposal tomorrow.

rvalyi:
Looks like we will have an agreement for the agreement module... Kind of stack overflow ;-)

bealdav:
Thanks for all to have shared opinions here and finally found a good solution.

alexis-via:
So here is a more detailed description of the solution proposed by @max3903:
1) "agreement" module: depends on "base" and contains only the "agreement" object as defined in the current "agreement_account" v10 module with the "base" fields (code, name, partner_id, company_id, active, signature_date, start_date, end_date)
2) "agreement_account" : depends on "agreement" and "account" : only adds the M2O field from invoice to agreement (and maybe also adds the "agreement_type selection field)
3) "agreement_sale" : depends on "agreement_account" and "sale" : only adds the M2O field from sale order to agreement and the copy to invoice upon invoice creation from SO (will be a simple port of the current agreement_sale v10 module)
4) legal_agreement : depends on "agreement" and contains the current code of the agreement module of v11

sbidoul:
Hi all,

FYI we are working on a major refactoring of the `contract` (as in subscription / recurring invoices) module in #207 and related PRs. It's almost finished and IMHO very powerful alternative to the enterprise `subscription` module.

For now, a contract/subscription is still an `account.analytic.account` (inherited from Odoo 8.0). Some preparatory work has been done to make it independent (the views do not inherit from the analytic account views anymore). A medium term evolution option could be to make subscriptions a type of `agreement` instead of an analytic account. 

Reflecting on @alexis-via 's last [comment](https://github.com/OCA/contract/pull/220#issuecomment-453423098), I'd like to mention that we came to the conclusion that the link between invoices and contracts , and SO and contracts is best done at the level of the invoice *line* or SO *line*. This mainly enables preserving the links when merging invoices, or generating one invoice for a set of contracts. So you may want to consider moving the M2O from invoice/so to agreement to the lines (I don't know if the time is right for that, I'll let you judge).

cc/ @sbejaoui

alexis-via:
@sbidoul Thanks for your valuable input and these news on the "contract" module.
In fact, if you plan to put links at SO/invoice LINE level (was already the case for invoices but it's a change for SO), then it's a good justification to keep the agreement module, because we specifically want the link at SO/invoice level.

sbidoul:
> we specifically want the link at SO/invoice level

Why? I'm curious. In my experience, it's more generic to put it on the line level, and optionally have a computed field at the invoice/so level, which is populated when all lines have the same contract.

pedrobaeza:
@sbidoul @alexis-via I already stated that both links should be valid. If you want to merge invoices, then put a m2m at invoice level instead.

pedrobaeza:
A computed m2m field also serves as suggested by @sbidoul

sbidoul:
Yes: nowadays in Odoo all links tend to be made at the line level. Plus navigation buttons on the main objects.

alexis-via:
Because, in EN16931 (EU electronic invoicing standard) the link to the contract is at invoice level, not at invoice line level => for the agreement module, I would like to stay close to the EU e-invoice datamodel. And, in the projects where we use the agreement module, it's clear that we want it at invoice/SO level, for the same reasons that payment terms/payment mode/fiscal position/... are at invoice/SO level, and not LINES.
But it's certainly a good thing to have the link at line level in the contract module, which target a different purpose and use-case.

sbidoul:
@alexis-via that's interesting. I personally don't think the electronic invoicing standard should influence the data model for that specific matter. Keeping the link at the line level would simply mean that you cannot reference the contract in the electronic invoice if you have an invoice that concern more than one contract. It can then be a deployment/configuration constraint to allow or not to mix contracts in different invoice lines, depending on the business requirements.

ygol:
@max3903 I start the migration of agreement which will be dependent on agreement_base. the model in agreement_base is named "agreement". 
> class Agreement(models.Model):
>     _name = 'agreement'
> _description = 'Agreement'

1. I would like your confirmation that your 11.0 agreement model will now be renamed to "agreement_legal" as in #189.

2. To start working on it. do you want me to start with a clean branch (12.0-mig-agreement_legal) or import the 11.0 history per https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-12.0#howto ?

3. When using https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-12.0#howto I also get in the repo the modules: "contract" and "contract_sale" should I keep them along in this branch or may I remove them? If remove, how would I do this in order to avoid the problem I got on #259 ?

4. How do I handle the dependency to #259 before it is merged to let me start migrating this module? Should I add it in this repo?

thank you

max3903:
@ygol There is no agreement_base and agreement, there is agreement and agreement_legal ;)

1. No, the agreement model from v11 will inherit the agreement model from v10.
2. Import the history per the documentation.
3. Those 2 modules are already migrated, so they should be in branch but they should not appear on the pull request diff.
4. You need a third branch with agreement and agreement_legal. When you are done with migrating agreement_legal, you cherry-pick the commit of this 3rd branch into your branch 12.0-mig-agreement_legal

Issue #231, opened 2018-12-07
Raising errors for missing modules by agreement_fsm (and it is not really installed)


etobella:
With agreement_fsm, you must have the repository  https://github.com/asaunier/geospatial.git installed in order to not raise errors. Can you remove somehow this lines of code:

https://github.com/OCA/contract/blob/11.0/agreement_fsm/models/fsm_order.py#L6

etobella:
Also, the dependency that is giving problems is not even in OCA:
https://github.com/OCA/field-service/blob/11.0/oca_dependencies.txt#L17

etobella:
@max3903 @bodedra this is related to your PR

pedrobaeza:
Please never introduce things outside OCA. @max3903 you should double check everything before merging.

etobella:
Also, the added module is specific to field-service, so I think it could be better to add it to field-service. 
Right now, we cannot refresh the modules we use from this repo because it is giving warnings because it is not using the standard inheritance system.
As I am not used to the field-service system, so I don't know the implications of the change.

max3903:
@etobella We will move it to OCA/field-service. Sorry for the inconvenience.

PR #229, opened 2018-12-05
[ADD] empty smart button place holder to easy inheritance

bodedra:
Issue reference https://github.com/OCA/contract/issues/228

PR #232, opened 2018-12-10
[11.0][RM] agreement_fsm module

bodedra:
Issue Reference: https://github.com/OCA/contract/issues/231


etobella (approved):
LGTM :+1:

pedrobaeza (approved):
There's no need of removing postgres version, but it's not a requirement any way for this version.

bodedra:
Thank you @pedrobaeza I have added that lines at https://github.com/OCA/contract/pull/229/files so I have removed it and there will be no chance to fail again. ;-)

pedrobaeza:
Well, there has been an error, as field-service is still used in agreement_maintenance, but I have re-added the dependency in oca_dependencies.txt, although I think we might move that one also to OCA/field-service.

pedrobaeza:
There are still problems with the CI, as by dependencies geospatial is trying to be installed and there's an error on that:

https://travis-ci.org/OCA/contract/jobs/466071710#L756

Possibilities:

1. Fixed .travis.yml for avoiding this problem.
2. Remove `agreement_maintenance`.

max3903:
@bodedra Let's go with option 2 for now.

PR #240, opened 2018-12-22
[10.0] [FIX] contract. Optimize insertion of dates in invoice and lines.

NL66278:
Try to prevent repetition of date and date format computation, with i/o to language table, for every contract line, while the computation is needed only once per contract.

### DIFF
--- a/contract/models/account_analytic_account.py
+++ b/contract/models/account_analytic_account.py
@@ -4,7 +4,9 @@
 # Copyright 2015 Pedro M. Baeza <pedro.baeza@tecnativa.com>
 # Copyright 2016-2017 Carlos Dauden <carlos.dauden@tecnativa.com>
 # Copyright 2016-2017 LasLabs Inc.
-# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+# Copyright 2018 Therp BV <https://therp.nl>.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+# pylint: disable=no-member
 
 from dateutil.relativedelta import relativedelta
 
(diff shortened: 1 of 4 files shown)


NL66278:
Actually refactoring the setting of the context, also allows for further optimizations by disabling mail tracking, this is not part of the PR because that would be a functional change, but in our customization we have:
```
    @api.multi
    def get_invoice_context(self):
        """Compute context for invoice creation."""
        self.ensure_one()
        ctx = super(AccountAnalyticAccount, self).get_invoice_context()
        ctx['tracking_disable'] = True
        return ctx
```

pedrobaeza:
Do you gain too much performance with this? I remember this was done on purpose for overwriting the method and alter the description got. @carlosdauden do you remember why?

NL66278:
@pedrobaeza Actually we do gain performance. To be true, it has an extra effect for us, as we also use the _insert_markers method for the invoice name, not only on the lines. But a lot of processing duplication is saved if you have multiple lines anyway.

I have done some profiling. And created an image of the results with gprof2dot:
```
gprof2dot -n 4.0 -f pstats invoice_cron.profile | dot -Tpng -o output.png
```
I have a before image and the final output. The before image was on 76 records, the final on 84. The difference includes the fact that we have also overriden get_invoice_context to disable mail tracking.
invoice_cron profile-before1

After changes:
output

pedrobaeza:
OK, it's hard to read that wide graph, but I trust if you say so. No blocking by my side.

rafaelbn:
@NL66278 could you please resolve conflicts ?

rafaelbn:
Related? https://github.com/OCA/contract/pull/218

yajo (approved):
It seems a pretty intelligent enhancement. Thanks!

pedrobaeza:
There are conflicts

NL66278:
@rafaelbn Yes this is related to #218. Actually an alternative approach to the same performance problems. So if this solution is sufficient and accepted, we might not need #218  any more.

NL66278:
@pedrobaeza @rafaelbn Conflicts have been resolved.

yajo:
IMHO, (without actually running performance tests, just by reading code), it seems like this patch covers more possible bottlenecks in a more scalable way than #218.

rafaelbn (approved):

rafaelbn:
Sorry for the noise @sbejaoui but you could have a similar situation of this

PR #260, opened 2019-01-14
[10.0] contract: allow to limit records on cron

chienandalu:
Allow to create process de recurring invoices in batches in case there are many contracts.

cc @Tecnativa

### DIFF
--- a/contract/__manifest__.py
+++ b/contract/__manifest__.py
@@ -9,7 +9,7 @@
 
 {
     'name': 'Contracts Management - Recurring',
-    'version': '10.0.4.2.1',
+    'version': '10.0.4.3.0',
     'category': 'Contract Management',
     'license': 'AGPL-3',
     'author': "OpenERP SA, "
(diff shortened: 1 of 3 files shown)


chienandalu:
ping @Yajo

pedrobaeza (changes requested):
This technique doesn't serve as some of the contracts might not be invoiceable, but are ignored by this line:

https://github.com/OCA/contract/pull/260/files#diff-47b79491db78ebebcb372014ce9c7bf3R274

so maybe you have more than these contracts than the limit set.

yajo:
Fixed.

I realized that, if somebody wanted to use this, updates would revert his change, because the cron wasn't noupdate, so I took the chance to give a sane non-blocking default and still setting it as noupdate to avoid updates to break the admin's choice.

yajo (approved):

pedrobaeza (approved):

pedrobaeza (changes requested):
Thinking again, I prefer to not change cron behavior, as it was intended for generating all invoices the same day. With this, you won't have them if you have more than limit * 24 invoices. This is patching a problem in other place, so it shouldn't be the default.

yajo:
OK, fixed.

pedrobaeza:
There are conflicts

yajo:
Fixed.

pedrobaeza (approved):

PR #261, opened 2019-01-15
[11.0] contract: allow to limit records on cron

yajo:
Forward port of https://github.com/OCA/contract/pull/260.

@Tecnativa

### DIFF
--- a/contract/models/account_analytic_account.py
+++ b/contract/models/account_analytic_account.py
@@ -278,13 +278,18 @@ class AccountAnalyticAccount(models.Model):
         return invoice
 
     @api.multi
-    def recurring_create_invoice(self):
+    def recurring_create_invoice(self, limit=None):
         """Create invoices from contracts
+
+        :param int limit:
+            Max of invoices to create.
 
         :return: invoices created
         """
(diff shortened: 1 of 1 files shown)


contract/models/account_analytic_account.py
```diff
@@ -321,7 +326,7 @@ def cron_recurring_create_invoice(self):
             ('date_end', '=', False),
             ('date_end', '>=', today),
         ])
-        return contracts.recurring_create_invoice()
+        return contracts.recurring_create_invoice(limit)
```
pedrobaeza: I didn't want to delay more the 10.0 merge, but now that you are forward-porting (but I don't think it's needed, as as said, we are fixing a problem elsewhere), please put `limit=limit` (always use keyword arguments).

yajo: That's not needed.

pedrobaeza: we have discussed about this in other occasions. Although you consider this not needed, please do it.

pedrobaeza (approved):

rafaelbn (approved):

PR #264, opened 2019-01-23
[12.0] [MIG] contract_section

Nikul-Chaudhary:
Contracts Management - Add section to invoice lines
===================================================

Enable the use of sections on the contract and the contract template.


Nikul-Chaudhary:
@pedrobaeza I have a doubt Regarding this module.

In v12 `sale.layout_category` model is deleted/Deprecated.

Here Can you please help me, What can i Do

pedrobaeza:
You have to switch to the same approach: having special contract lines for indicating a section, and generate that as the same special type invoice lines. It's a lot of work to do, I know.

Nikul-Chaudhary:
> You have to switch to the same approach: having special contract lines for indicating a section, and generate that as the same special type invoice lines. It's a lot of work to do, I know.

Ok, I create a new object like this `sale.layout_category`, Right?

pedrobaeza:
No, just add a field in the contract line for indicating the type (just like in sale.order.line and account.invoice.line), and adapt the interface for adding multiple options to add a line (again the same as in sales and invoices). Then propagate this field the same as the others to the invoice line.

Finally, you have to make a migration script for transforming old contract sections to this new one (although you can leave this one being too much work).

Nikul-Chaudhary:
@pedrobaeza `layout_category_id` field in two model.
**1 for account.invoice.line**
https://github.com/OCA/OpenUpgrade/blob/12.0/addons/sale/migrations/12.0.1.1/openupgrade_analysis.txt#L2
**2 for sale.order.line**
https://github.com/OCA/OpenUpgrade/blob/12.0/addons/sale/migrations/12.0.1.1/openupgrade_analysis.txt#L36

So, IMO I create a new field like `layout_category_id` and adding above model.

> No, just add a field in the contract line for indicating the type

Which field??

pedrobaeza:
As said, in v12 there's a new field `display_type`. Please study a bit more how it works in v12 or dismiss the migration.

Nikul-Chaudhary:
@pedrobaeza Can you Please take a look?

If not Correct then Close it.

pedrobaeza:
Can you please enable the same system as in sales order/invoices (with multiple "Add..." options in the one2many tree)?

seleccion_039

And check also if that is protected by any security group (I think not).

Nikul-Chaudhary:
@pedrobaeza Noted!

pedrobaeza:
Sections are already handled in standard `contract`module, so no need of this module anymore.

PR #270, opened 2019-01-25
[FIX] Adding arguments in the UI call of recurring_create_invoice

aitorbouzas:
Added parameter to UI call of `recurring_create_invoice` so that the arg `limit` doesn't get parsed as context.


pedrobaeza (approved):
Fast-tracking this

PR #281, opened 2019-03-04
[FIX] fixed domains in journal_id

jarroyomorales:
Added company filter to journal_id field in account.analytic.account form view.

### DIFF
--- a/contract/views/account_analytic_account_view.xml
+++ b/contract/views/account_analytic_account_view.xml
@@ -46,6 +46,7 @@
                     <field name="contract_template_id" colspan="4" domain="['|', ('contract_type', '=', contract_type), ('contract_type', '=', False)]" context="{'default_contract_type': contract_type}"/>
                     <field name="journal_id"
                            attrs="{'required': [('recurring_invoices', '=', True)]}"
+                           domain="[('type','=','sale'), ('company_id','=',company_id)]"
                     />
                     <field name="pricelist_id"/>
                     <field name="company_id" groups="base.group_multi_company"/>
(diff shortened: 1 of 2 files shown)


pedrobaeza:
All this company specific overhead is starting to be a little annoying... This is what happens when you co counter-current of main Odoo approach, and this is more error prone most of the time.

JordiBForgeFlow (approved):

etobella (approved):
You should dump version.
@pedrobaeza We know this is annoying, but the domains should be correctly defined. Most of the users use a single company, but we should think in all the options.

pedrobaeza:
The question is that this should be handled at other level for your "mc_fixes", not modifying each of the modules. We can discuss it this Thursday.

contract/views/account_analytic_account_view.xml
```diff
@@ -46,6 +46,7 @@
                     <field name="contract_template_id" colspan="4" domain="['|', ('contract_type', '=', contract_type), ('contract_type', '=', False)]" context="{'default_contract_type': contract_type}"/>
                     <field name="journal_id"
                            attrs="{'required': [('recurring_invoices', '=', True)]}"
+                           domain="[('type','=','sale'), ('company_id','=',company_id)]"
```
yajo, line 49: Does it make sense to apply the domain in the model instead?

contract/views/account_analytic_account_view.xml
```diff
@@ -112,7 +113,7 @@
                 <attribute name="context">{'default_customer': True, 'default_supplier': False}</attribute>
             </field>
             <field name="journal_id" position="attributes">
-                <attribute name="domain">[('type', '=', 'sale')]</attribute>
+                <attribute name="domain">"[('type','=','sale'), ('company_id','=',company_id)]"</attribute>
```
yajo, line 116: Same here.

contract/views/account_analytic_account_view.xml
```diff
@@ -133,7 +134,7 @@
                 <attribute name="context">{'default_customer': False, 'default_supplier': True}</attribute>
             </field>
             <field name="journal_id" position="attributes">
-                <attribute name="domain">[('type', '=', 'purchase')]</attribute>
+                <attribute name="domain">"[('type','=','purchase'), ('company_id','=',company_id)]"</attribute>
```
yajo, line 137: Same here.

pedrobaeza:
Already commented for going to a general approach, so this is WIP for now (to be closed in fact).

pedrobaeza:
This can be closed due to the new approach in the widget.

Issue #137, opened 2018-01-08
[Functional Improvement] Recurring Next Date Scoope


angelmoya:
The cron function witch create recurring invoices is:

```
    @api.model
    def cron_recurring_create_invoice(self):
        today = fields.Date.today()
        contracts = self.with_context(cron=True).search([
            ('recurring_invoices', '=', True),
            ('recurring_next_date', '<=', today),
            '|',
            ('date_end', '=', False),
            ('date_end', '>=', today),
        ])
        return contracts.recurring_create_invoice()
```

That create invoices only if ('recurring_next_date', '<=', today)

what do you think about create invoices X days before, for example:
```
scoope_day = today + timedelta(X)
...
('recurring_next_date', '<=', scoope_day)
```

pedrobaeza:
Well, it can be good if it's set as a parameter. Can be done at journal level, company level or system parameter. The second one can be the best balanced maybe.

angelmoya:
Yes, I was thinking about create X as a system parameter.

angelmoya:
Yes, I was thinking about create X as a system parameter.

pedrobaeza:
Better by company...

cesarsc79:
[contract_days_before.zip](https://github.com/OCA/contract/files/1714409/contract_days_before.zip)

Sorry:

I am development a new module (not finished yet... NO TIME )
it create 2 field

1º Days before
2º Virtual date

Virtual date is the today.date + Days before
    
So, you can custom every contract separately

 antelacion = fields.Integer(
        string='Dias de antelacion',
        default=5,
        help="Mete los dias de antelacion con los que quieres ejecutar el cron",
    )

virtual_date = fields.Date(
        string='Fecha Virtual',
        compute='_compute_date'
    )

    def _compute_date(self):
        self.virtual_date = date.today() + timedelta(days=self.antelacion)

    @api.model
    def cron_recurring_create_invoice(self):

        contracts = self.with_context(cron=True).search([
            ('recurring_invoices', '=', True),
            ('recurring_next_date', '<=', **virtual_date**),
            '|',
            ('date_end', '=', False),
            ('date_end', '>=', today),
        ])
        return contracts.recurring_create_invoice()

well .... dont work... i am not an an expert in odoo orm

someone wants finish it??

p.d.
other option is add an action in contract form list, for select all contract you want and execute the cron manually without need to go into form view...

babatoko:
This feature will be highly appreciate.

[...]

PR #282, opened 2019-03-04
[10.0][IMP] contract. Enable automatic invoice generation before invoice date.

NL66278:
Implements the functional enhancement requested here: https://github.com/OCA/contract/issues/137

Per company it will be possible to specify how many days before the next invoice date the invoices for recurring contracts will be generated.

### DIFF
--- a/contract/models/account_analytic_account.py
+++ b/contract/models/account_analytic_account.py
@@ -4,7 +4,7 @@
 # Copyright 2015 Pedro M. Baeza <pedro.baeza@tecnativa.com>
 # Copyright 2016-2017 Carlos Dauden <carlos.dauden@tecnativa.com>
 # Copyright 2016-2017 LasLabs Inc.
-# Copyright 2018 Therp BV <https://therp.nl>.
+# Copyright 2018-2019 Therp BV <https://therp.nl>.
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 # pylint: disable=no-member
 
@@ -14,6 +14,7 @@ from dateutil.relativedelta import relativedelta
 from odoo import api, fields, models
 from odoo.exceptions import ValidationError
 from odoo.tools.translate import _
+from odoo.tools import DEFAULT_SERVER_DATE_FORMAT
 
 
 class AccountAnalyticAccount(models.Model):
@@ -353,8 +354,9 @@ class AccountAnalyticAccount(models.Model):
         company_model = self.env['res.company']
         for company in company_model.search([]):
             days_before = company.contract_pregenerate_days or 0
-            cutoffdate = (today + relativedelta(days=days_before)).strftime(
-                '%Y-%m-%d')
+            cutoffdate = (
+                today + relativedelta(days=days_before)
+            ).strftime(DEFAULT_SERVER_DATE_FORMAT)
             contracts = self.with_context(cron=True).search([
                 ('company_id', '=', company.id),
                 ('recurring_invoices', '=', True),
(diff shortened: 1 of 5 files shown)


ntsirintanis (approved):
Tested, LGTM.

sbidoul:
@sbejaoui can you review this? Is there something comparable in #207 ?

@NL66278 if you are interested in 12.0, I encourage you to have a look at #207. It's big change (moving recurrence to contract lines). Nevertheless it should migrate smoothly from 10.0. And it enables many additional use cases. There is also #208 which enables contract creation from sale orders in a much more robust way.

sbejaoui:
LGTM, 

IMO, it is better if it is in a separate module.

NL66278:
@sbejaoui Why a separate module? If the user does not need/want the functionality, nothing changes.

sbejaoui:
My idea is to keep `cron_recurring_create_invoice` as simple as possible and make separate module for each use case attending to influence the way it works. Otherwise, with all use cases we can imagine, we can quickly bring a big method hard to maintain, hard to debug.  

This why we did some refactoring in #207 to make this logic overridable and easy to evolve.

babatoko:
@NL66278,
It looks good to me. Which version of odoo is it built for. If not for 10 would it be possible to backport to it?

Cheers.

NL66278:
@babatoko It is for 10.0

babatoko:
That's a very good news thanks @NL66278. I'm going to test further.

Cheers

flotho (approved):
LGTM

mamojdick:
@NL66278 nice. I tried and appreciated this feature, but the "date_invoice" field is filled with "recurring_next_date". would it be possible to leave it empty?

NL66278:
@mamojdick Thanks for your appreciation! But with or without my PR, the date_invoice field is filled from recurring_next_date. My PR just enables creating the invoices some configurable time in advance, to allow checking, sending them out already, etc. What would the use-case be for not filling date_invoice? Actually that date would then be set to the current date, defeating the purpose of my PR. Maybe I did not understand your question well, can you elaborate?

mamojdick:
>  But with or without my PR, the date_invoice field is filled from recurring_next_date.


Yes i know this.

> Actually that date would then be set to the current date

Yes that is my idea: i want to create hundreds invoices for an ISP like this:
#START# 1° day of the month
#END# last dey of the month
X day before (with your PR)
check it and validate it in  X+Y (where y is days spent to check)
so date_invoice is current day.

I am in Italy and we introduce the e-invoice, the date_invoice must be the same date we send it, not in the future.

I hope I was clear.
TNX

contract/models/account_analytic_account.py
```diff
@@ -341,14 +342,27 @@ def recurring_create_invoice(self, limit=None):
...
+        invoice date, to allow the user to check the invoices before
+        confirmation, or to send out the invoices to the customers before
+        the invoice date.
+        """
+        today = datetime.today()
+        company_model = self.env['res.company']
+        for company in company_model.search([]):
+            days_before = company.contract_pregenerate_days or 0
+            cutoffdate = (today + relativedelta(days=days_before)).strftime(
+                '%Y-%m-%d')
```
angelmoya: better use:

from odoo.tools import DEFAULT_SERVER_DATE_FORMAT

.strftime(DEFAULT_SERVER_DATE_FORMAT)

angelmoya (approved):
Small change, but LGTM

contract/models/account_analytic_account.py
NL66278: @angelmoya Thanks! Done!

rafaelbn:
@NL66278 could you squash commits please before merging? Thanks!

rafaelbn:
Thanks a lot for this improvement @NL66278 . Just remember if you want to forward port it to 11.0 and 12.0 as you commented in https://github.com/OCA/contract/issues/137#issuecomment-469604221

rafaelbn (approved):

PR #286, opened 2019-03-18
Decimal precision in Quantity of Contract Lines

danypr92:
We need to define the decimal precision in `quantity` of Contract Lines as in the `price` attribute.

We expect the same behavior as in the invoice line `quantity`.

-----------------

I am not sure that this repository was the correct place to put this data, I found [the other same data in the Odoo code / OCB repository](
https://github.com/OCA/OCB/blob/11.0/addons/product/data/product_data.xml#L32), but I think that has more sense that the data was here if it is used in this module. If this reflexion is wrong, please tell me. Thanks!

### DIFF
--- a/contract/models/account_analytic_contract_line.py
+++ b/contract/models/account_analytic_contract_line.py
@@ -35,6 +35,7 @@ class AccountAnalyticContractLine(models.Model):
     )
     quantity = fields.Float(
         default=1.0,
+        digits=dp.get_precision('Product Unit of Measure'),
         required=True,
     )
     uom_id = fields.Many2one(


pedrobaeza:
You must use the existing decimal precision. Don't create a new one.

danypr92:
Thanks for your answer @pedrobaeza.

I can't use the existing decimal precisions because they are used in other parts of the code.
I want to modify the decimal precision only in the quantity field of contract lines.

which is the correct way to do this if I can't create a new decimal precision?

pedrobaeza:
You must use the exact same precision used in invoice lines: https://github.com/odoo/odoo/blob/84df9948ed07310e9343876935c0050c996b73bd/addons/account/models/account_invoice.py#L1518

danypr92:
ok!

I changed the code to use the precision used in invoice lines.

Thanks, @pedrobaeza!

pedrobaeza (approved):

luismontalba (approved):
Tested on runbot

PR #288, opened 2019-03-25
[12.0][MIG] agreement legal (from v11 agreement)

ygol:
ref. #189 

This module depends on module agreement in PR #287

PR #287, opened 2019-03-19
[MIG] agreement: Migration to 12.0

ygol:
I started over. Replacing #259.

I include my previous [comment]https://github.com/OCA/contract/pull/259#issuecomment-453753891

> I haven't added groups, permissions as they will be different when used by @alexis-via (part of account) than by @max3903 (specific module permissions).
> 
> I did add a top menu "agreement" though.

PR #292, opened 2019-03-27
[12.0][WIP][MIG] agreement_sale

murtuzasaleh:
WIP


murtuzasaleh:
@max3903 When I merge the agreement_sale v10 and v11 code in v12 I have one question regarding [This method](https://github.com/OCA/contract/pull/292/commits/92480fda3ef3721eba1eb80234b3248de7c16841#diff-995aaf109ee1e6b58bf33ad8e227f14cR11) because agreement_id is added in In agreement_account module [account.invoice](https://github.com/OCA/contract/blob/10.0/agreement_account/models/account_invoice.py) 

So can I add this field in agreement_sale v12 or In which module this field functionality will be added?

max3903:
@murtuzasaleh `sale` depends on `payment` which depends on `account`.

I think we can have `agreement_sale` depends on `agreement_account`.

agreement_sale/__manifest__.py
```diff
@@ -0,0 +1,26 @@
...
+{
+    'name': 'Agreement - Sale',
+    'summary': 'Create an agreement when the sales order is confirmed',
+    'version': '12.0.1.0.0',
+    'license': 'AGPL-3',
+    'author': 'Open Source Integrators, Odoo Community Association (OCA)',
+    'category': 'Agreement',
+    'website': 'https://github.com/OCA/contract',
+    'depends': [
+        'agreement_legal',
```
max3903, line 13: `agreement` and `sale`, not `agreement_legal`

murtuzasaleh: But [this view](https://github.com/OCA/contract/pull/292/files#diff-13b9083d98e59c16fb7d525111063146R12) is created in agreement_legal module.

murtuzasaleh:
> @murtuzasaleh `sale` depends on `payment` which depends on `account`.
> 
> I think we can have `agreement_sale` depends on `agreement_account`.

An agreement_account module is becoming agreement module in v12 so is there any other functionality for agreement_account in v12 because agreement_account is not in v11.

agreement_sale/__manifest__.py
max3903: The `agreement` module should define the notebook (and it could be empty) and the `agreement_sale` one should add a page to it.

bealdav:
@murtuzasaleh don't forget there is https://github.com/OCA/contract/tree/10.0/agreement_sale which is the continuity of agreement (older agreement_account)

I think this implementation must be split because module makes a lot more than the module name indicates

EDIT

> the agreement template is copied into an agreement with a copy of the sections, clauses, recitals and appendices

It refers to legal part of agreement legal. I think feature are upper than just the minimal feature of an `agreement sale`

murtuzasaleh:
> @murtuzasaleh don't forget there is https://github.com/OCA/contract/tree/10.0/agreement_sale which is the continuity of agreement (older agreement_account)
> 
> I think this implementation must be split because module makes a lot more than the module name indicates

@bealdav I have Merged agreement_sale V10 and V11 code in New V12 branch.

Do you mean remaining agreement_account module functionality is created in v12 agreement_account like invoice code?

max3903:
@bealdav In our scenario, we set an agreement template on the quotation and the SO confirmation creates the agreement and link it to the SO. Is that an issue for your scenario? When the invoice is created, the agreement is set on the SO so you can do the same as in v10 or am I missing something?

bealdav:
@max3903, on our side we only populate the sale.order with a m2o agreement. 

It seems agreement template is an advanced use case (legal or other) but not a minimal one.

Sorry for my remarks, I'm ok than to serve several uses it's complex, but it's better to split to minimal features : (think about future migrations).

It could be interesting for you to have base module migrated in v14 for example by akretion

max3903:
What do you think of having a new group and display the agreement template field on the SO only to that group? Installing `agreement_legal` would add that group in the inherited ones of the Employee group.

bealdav:
Ok for that if it's an other module like agreement_legal. Thanks

murtuzasaleh:
@max3903 @bealdav Do you mean remaining agreement_account module functionality is created in v12 agreement_account like invoice code?

max3903:
I meant the following:

# Agreement

* Add the group "Manage advanced agreements for legal department"
* Add the notebook to the view

# Agreement Sale

* Add the field agreement_template_id and agreement_id to sale.order
* Display the agreement_template_id field to the "Manage advanced agreements for legal department" group
* Add the "Sales" page in the notebook

# Agreement Legal

* Add the "Manage advanced agreements for legal department" group as an inherited group of "Employee"

murtuzasaleh:
@max3903 Ok understood,
So @ygol will create group `Manage advanced agreements for legal department` and add blank Add the notebook to the view in this [PR](https://github.com/OCA/contract/pull/287)
Also @ygol will create `Manage advanced agreements for legal department` group as an inherited group of "Employee" in this [PR](https://github.com/OCA/contract/pull/288)

bealdav:
On my side I want to use sale and agreement objects, just linked by a m2o and filled manually and propagate to the invoice (perfect) and OK for sale o2m in agreement object.

The behavior which are not necessary in agreement_sale are:

- is_serviceprofile field
- agreement_template_id field
- agreement.line obj
- group Manage advanced agreements 

Sorry for this news

max3903:
@murtuzasaleh Sounds like we will have to create a `agreement_legal_sale` module...

murtuzasaleh:
> @murtuzasaleh Sounds like we will have to create a `agreement_legal_sale` module...

@max3903 Ok I will close this PR and Create new PR based on V11 agreement_sale module to agreement_legal_sale in V12
#294

bealdav:
Thanks for your work @murtuzasaleh and @max3903

PR #307, opened 2019-04-05
[10.0] Add module contract_invoice_start_end_dates

florian-dacosta:
It is glue code between contract module and account_invoice_start_end_dates module.
I did not put it as auto install though. There is not assitional configuration to do when installing the module and the date are put automatically, so I prefer let the choice to install it or not.

It is kind of a migration of https://github.com/OCA/contract/pull/51
But this later was never merged, so I did not start from it.
Also, the implementation is way more simple since the contract module has totally changed since version 8.

@alexis-via @bguillot


alexis-via (approved):

contract_invoice_start_end_dates/__manifest__.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+# Copyright 2019 Akretion
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+{
+    'name': 'Contract Invoice Start End Dates',
+    'version': '12.0.1.0.0',
```
alexis-via: You're pushing to v10 branch, so version is wrong

alexis-via (changes requested):

PR #312, opened 2019-04-11
[12.0][IMP] agreement_legal: add custom dynamic parties

ygol:
Add the option for the user to replace the default  parties section by a dynamic custom html field.


agreement_legal/models/agreement.py
```diff
@@ -170,6 +170,20 @@ def _default_stage_id(self):
     company_contact_email = fields.Char(
         related="company_contact_id.email", string="Email"
     )
+    use_parties_content = fields.Boolean(
+        string="Use parties content",
+        default = False,
```
max3903:
```suggestion
        default=False,
```

agreement_legal/models/agreement.py
```diff
@@ -170,6 +170,20 @@ def _default_stage_id(self):
...
+        help="Use custom content for parties" )
```
max3903:
```suggestion
        help="Use custom content for parties")
```

max3903 (approved):
LGTM. No test.

agreement_legal/models/agreement.py
nikul-serpentcs: no need to default False, Boolean always False

agreement_legal/models/agreement.py
```diff
@@ -170,6 +170,19 @@ def _default_stage_id(self):
     company_contact_email = fields.Char(
         related="company_contact_id.email", string="Email"
     )
+    use_parties_content = fields.Boolean(
+        string="Use parties content",
+        help="Use custom content for parties")
+    parties = fields.Html(
+        string="Parties",
+        track_visibility="onchange",
+        help="Parties of the agreement",
```
max3903, line 206: @ygol Can you add a default value mixing the current content with the dynamic fields? User would have a base to start with and no need to figure out the dynamic fields... Thank you.

ygol: @max3903 Do you mean removing the boolean field and always show the html field with default value? 
Isn'n it cleaner to have both? 
IMO the form is much less cluttered without the html field (that is shown only if needed).

max3903: We can keep the boolean field. I just think we should help the user with a default value.

osi-scampbell (approved):

max3903 (approved):

PR #315, opened 2019-04-19
[10.0][FIX] total invoice untaxed and get journal by sequence

sergiocorato:


sergiocorato:
Splitted commits by module and rebased

rafaelbn:
Hi @sergiocorato ,

Which is the issue? how to reproduce it? I would like to test in production

Thanks

sergiocorato:
> Hi @sergiocorato ,
> 
> Which is the issue? how to reproduce it? I would like to test in production
> 
> Thanks

Hi @rafaelbn,
I added #364 to show that actual implementation do not work, I'll commit here the same tests

sergiocorato:
@rafaelbn I committed the tests, issue was that total invoiced in analytic account was from the whole invoice, instead of the invoice line with the analytic account. The amount is better to be without VAT, as this is not a revenue (and would have to be computed too).

sergiocorato:
in addition I add a search for journal by sequence, but it's no fundamental (I think it's only logic)

rafaelbn:
You are modifying:

> contract/models/account_analytic_contract.py 

and

>  contract_show_invoice/models/account_analytic_account.py 

So the fix is for contract_show_invoice?

rafaelbn:
OK, I see from @pedrobaeza https://github.com/OCA/contract/pull/364#issuecomment-521198680

> We can't merge this branch if this is going to be red. Put both commits in the same PR.

OK

rafaelbn (approved):

PR #316, opened 2019-04-19
[10.0][IMP] add contract show sale

sergiocorato:


sergiocorato:
perhaps travis errors are not related to this PR, but something similar to [this](https://github.com/OCA/account-financial-tools/pull/802)? I don't understand

contract_show_sale/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    'name': 'Contract Show Sale',
+    'summary': 'Button in contracts to show their sales',
+    'version': '10.0.1.0.1',
+    'license': 'AGPL-3',
+    'author': 'Sergio Corato,'
+              'Odoo Community Association (OCA)',
+    'website': 'https://efatto.it',
```
nikul-serpentcs: @sergiocorato  URL should be `https://github.com/OCA/contract`

contract_show_sale/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+    'depends': ['account', 'analytic', 'sale'],
```
nikul-serpentcs: @sergiocorato Remove `'account'` in depends because If you install `sale` that time `account` install automatically.

contract_show_sale/models/account_analytic_account.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+class AccountAnalyticAccount(models.Model):
+    _inherit = 'account.analytic.account'
+
+    @api.multi
+    def _get_total_sale(self):
+        sale_model = self.env['sale.order']
+        for analytic in self:
+            fetch_data = sale_model.read_group(
```
nikul-serpentcs, line 14: here Better to use `search_count` method.

```
@api.multi
    def _get_total_sale(self):
        sale_model = self.env['sale.order']
        for analytic in self:
            fetch_data = sale_model.search_count(
                [('project_id', '=', analytic.id),
                 ('state', 'not in', ['draft', 'sent', 'cancel'])],
                )
            analytic.total_sale = fetch_data
```

nikul-serpentcs (commented):
Improved Code

contract_show_sale/__manifest__.py
sergiocorato: done

contract_show_sale/__manifest__.py
sergiocorato: done

contract_show_sale/models/account_analytic_account.py
sergiocorato: mmm, doesn't it only give record count? I need total amount, not only record count

rafaelbn:
Hi @sergiocorato , thank you for this contribution!

Why not adding this to `contract_sale` ? It looks like the logic

sergiocorato:
> Hi @sergiocorato , thank you for this contribution!
> 
> Why not adding this to `contract_sale` ? It looks like the logic

not a bad idea

sergiocorato:
I close for inactivity

PR #321, opened 2019-05-05
[10.0] [FIX] product_contract. Correctly create contract from template.

NL66278:
Before a contract was created (actually not, just an analytic account with recurring_invoices not set), without the contract lines. Those were not copied from the template.

### DIFF
--- a/product_contract/models/sale_order_line.py
+++ b/product_contract/models/sale_order_line.py
@@ -1,8 +1,9 @@
 # -*- coding: utf-8 -*-
 # Copyright 2017 LasLabs Inc.
-# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
-
-from odoo import fields, models
+# Copyright 2019 Therp BV <https://therp.nl>.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+# pylint: disable=missing-docstring,protected-access
+from odoo import api, fields, models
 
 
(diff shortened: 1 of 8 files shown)


product_contract/models/sale_order_line.py
```diff
@@ -1,8 +1,9 @@
 # -*- coding: utf-8 -*-
 # Copyright 2017 LasLabs Inc.
-# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
```
umiphos: Don't delete the License

product_contract/models/sale_order_line.py
```diff
@@ -12,3 +13,32 @@ class SaleOrderLine(models.Model):
...
         string='Contract'
     )
+
+    @api.multi
+    def create_contract(self):
+        """Create contract for sale order line.
+
+        Should be called on confirmation of sale order.
+        """
+        for this in self:
```
umiphos: you can add a filtered in order to avoid the if in the loop, also, its better if we avoid random names, can you change **this** for **line** or something that makes a reference to the singularity of the object?

```python
for line in self.filtered(lambda r: not r.product_id.is_contract:
```

NL66278: Done, but simplified and corrected filter.

rafaelbn (approved):
:+1:  thanks for this FIX !!!

PR #333, opened 2019-05-17
[FIX] agreement_legal: Issue #331

### LINKED ISSUE #331: [12.0][BUG] agreement_legal not installable
Hello,

I have just spotted that module `agreement_legal` is not installable: [Travis Log](https://travis-ci.org/OCA/contract/jobs/532814199#L1573)

max3903:
Fix #331

### DIFF
--- a/agreement_legal/models/agreement.py
+++ b/agreement_legal/models/agreement.py
@@ -8,9 +8,6 @@ class Agreement(models.Model):
     _name = "agreement"
     _inherit = ["agreement", "mail.thread"]
 
-    def _default_stage_id(self):
-        return self.env.ref("agreement_legal.agreement_stage_new")
-
     # General
     name = fields.Char(string="Title", required=True)
     is_template = fields.Boolean(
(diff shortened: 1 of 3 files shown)


agreement_legal/models/agreement.py
```diff
@@ -389,6 +388,8 @@ def create(self, vals):
             vals["code"] = self.env["ir.sequence"].next_by_code(
                 "agreement"
             ) or _("New")
+        vals["stage_id"] = \
```
pedrobaeza: This should be conditional: `if not vals.get('stage_id'):`

pedrobaeza (changes requested):

pedrobaeza:
`_default_stage_id` is useless now. And you should also protect `self.env.ref` calls because you can remove the stage in UI and you'll get a code error in that case.

patrickrwilson (approved):

max3903:
@katyukha Can you review this fix?

ygol (approved):

osi-scampbell (approved):

pedrobaeza:
@max3903 why have you ignored my comments at https://github.com/OCA/contract/pull/333#issuecomment-493615492 ?

max3903:
@pedrobaeza I removed `_default_stage_id` like you suggested. Concerning the protection of `self.env.ref`, I guess I didn't understand what you meant. What did you expect?

Issue #335, opened 2019-05-20
[IMP] agreement_legal: multi-level field for dynamic content


max3903:
The dynamic fields work with 2 levels. It is fine for the description, however for the recitals, sections, clauses and appendices, we are limited to the fields of the agreement. The user cannot select the customer name of the agreement from the recitals.

The idea is to use part of the domain widget to have a popup window like this:
![Screenshot from 2019-05-20 11-32-43](https://user-images.githubusercontent.com/939952/58037424-06b9eb00-7af3-11e9-86b9-e7d49cedc329.png)

to select the field we want, without level limitation and fill in the value to copy with something like `company_id.partner_id.display_name`

max3903:
@patrickrwilson

max3903:
Ping @smangukiya

PR #340, opened 2019-05-29
[IMP] agreement_legal: Multi-level field selector

smangukiya:
w.r.t: https://github.com/OCA/contract/issues/335

Module needs upgrade:
> agreement_legal

Here is the example of agreement.recital:

Image 1: (It will provide domain widget popup to generate dynamic field string)
dynamic_content_popup

Image 2: It will show dynamically generated string:
dynamic_content

@max3903


max3903:
@smangukiya I will take care of the other fields. Thank you!

max3903 (approved):

ygol (commented):
It seems to work fine in the runbot for the recitals model but all other such fields are still using the previous "field_id". maybe we need a fresh runbot instance to be rebuild?!

ygol (approved):
LGTM

patrickrwilson (approved):
I tested this and it does seem to work as intended.

nikul-serpentcs:
`Merge branch '12.0' into 12.0-imp-16335-dynamic-content`

@max3903 Regarding Commit message
FYI You are directly used `git pull upstream`
here Best way to using rebase: `git rebase upstream`, So this msg will not come and you get latest commit's 

Thank you!!

Issue #343, opened 2019-06-05
[12.0][BUG] agreement_legal: company's primary contact selection


ygol:
1. edit an agreement.
2. select a primary contact for the company
```
Error:
Error: AttributeError: object has no attribute 'partner_id'

http://3373406-12-0-577fa0.runbot3.odoo-community.org/web/content/563-5dd5e44/web.assets_backend.js:205
Traceback:
__getattribute__@http://3373406-12-0-577fa0.runbot3.odoo-community.org/web/content/563-5dd5e44/web.assets_backend.js:205:7
py.PY_getAttr@http://3373406-12-0-577fa0.runbot3.odoo-community.org/web/content/563-5dd5e44/web.assets_backend.js:168:176
py.evaluate@http://3373406-12-0-577fa0.runbot3.odoo-community.org/web/content/563-5dd5e44/web.assets_backend.js:272:11
py.evaluate@http://3373406-12-0-577fa0.runbot3.odoo-community.org/web/content/563-5dd5e44/web.assets_backend.js:267:111
py.evaluate@http://3373406-12-0-577fa0.runbot3.odoo-community.org/web/content/563-5dd5e44/web.assets_backend.js:269:99
py.eval@http://3373406-12-0-577fa0.runbot3.odoo-community.org/web/content/563-5dd5e44/web.assets_backend.js:274:284
eval_domains/<@http://3373406-12-0-577fa0.runbot3.odoo-community.org/web/content/563-5dd5e44/web.assets_backend.js:413:136
_.forEach@http://3373406-12-0-577fa0.runbot3.odoo-community.org/web/content/557-da04fe9/web.assets_common.js:12:558
_.mixin/</_.prototype[name]@http://3373406-12-0-577fa0.runbot3.odoo-community.org/web/content/557-da04fe9/web.assets_common.js:69:526
eval_domains@http://3373406-12-0-577fa0.runbot3.odoo-community.org/web/content/563-5dd5e44/web.assets_backend.js:412:265
pyeval@http://3373406-12-0-577fa0.runbot3.odoo-community.org/web/content/563-5dd5e44/web.assets_backend.js:428:24
stringToArray@http://3373406-12-0-577fa0.runbot3.odoo-community.org/web/content/563-5dd5e44/web.assets_backend.js:383:2283
_getDomain@http://3373406-12-0-577fa0.runbot3.odoo-community.org/web/content/563-5dd5e44/web.assets_backend.js:1438:167
[...]

PR #344, opened 2019-06-05
[FIX] issue #343 [12.0][BUG] agreement_legal: company's primary contact selection

ygol:
Fix issue #343.

@max3903, regarding the domain for the field "notification_address_id"; what is the idea to add the company's contact in there? 
TIA

### DIFF
--- a/agreement_legal/views/agreement.xml
+++ b/agreement_legal/views/agreement.xml
@@ -122,7 +122,7 @@
                             <field name="partner_contact_email" widget="email" readonly="1" nolabel="1"/>
                         </group>
                         <group name="contact_right" string="Primary Contact">
-                            <field name="company_contact_id" domain="[('parent_id', '=', company_id.partner_id)]" nolabel="1"/>
+                            <field name="company_contact_id" domain="[('parent_id', '=', company_id)]" nolabel="1"/>
                             <field name="company_contact_phone" widget="phone" readonly="1" nolabel="1"/>
                             <field name="company_contact_email" widget="email" readonly="1" nolabel="1"/>
                         </group>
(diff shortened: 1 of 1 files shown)


max3903:
@ygol In most of the contracts it includes clauses where you are supposed to notify the other party for any changes. This is the field to set the person to notify for those changes.

max3903 (approved):

ygol:
> @ygol In most of the contracts it includes clauses where you are supposed to notify the other party for any changes. This is the field to set the person to notify for those changes.

Then there’s maybe a mistake. The domain add not only the partner’s contacts, but also the company’s contacts. This doesn’ Seem to make sense.

max3903:
@ygol We can also get rid of the field and set the third party contact as a follower.

ygol:
this is best IMO. shall I do that in current pr?

max3903:
I would like to get @patrickrwilson opinion on this...

ygol:
Just to add to the discussion. I have extended agreement with several m2o to res.partner for:
- partner's interlocutor
- company's interlocutor
- partner's representative company + partner's representative contact, person
- third-party  (end-user, beneficiary, etc)

Additionally, 
- I needed the possibility to use a different company for an agreement.
- I have added a domain to partner_id to only show companies.

Let me know if you want some of this in the OCA agreement_legal.

patrickrwilson:
> @ygol We can also get rid of the field and set the third party contact as a follower.

I think this is a good option

ygol:
May I simply remove the "notification_address_id" field from both the model and the view?
Should I change the version number? It is currently at 12.0.1.1.0?
What happens to users who will upgrade and lose that field in the process? Maybe they were using it?!
TIA @max3903 @patrickrwilson @osi-scampbell

max3903:
@ygol 

> May I simply remove the "notification_address_id" field from both the model and the view?

Yes

> Should I change the version number? It is currently at 12.0.1.1.0?

Yes, it should be changed to 12.0.2.0.0

> What happens to users who will upgrade and lose that field in the process?

A migration script should be written to set this partner as a follower.

PR #348, opened 2019-06-10
[IMP] agreement_serviceprofile

osi-scampbell:
This PR adds notes to agreement.serviceprofiles and adds a One2many field on agreements linking them to agreement.serviceprofiles

### DIFF
--- a/agreement_serviceprofile/models/agreement.py
+++ b/agreement_serviceprofile/models/agreement.py
@@ -1,0 +1,12 @@
+# Copyright (C) 2018 - TODAY, Open Source Integrators
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class Agreement(models.Model):
+    _inherit = 'agreement'
+
+    serviceprofile_ids = fields.One2many('agreement.serviceprofile',
+                                         'agreement_id',
+                                         string="Service Profiles")
--- a/agreement_serviceprofile/views/agreement.xml
+++ b/agreement_serviceprofile/views/agreement.xml
@@ -1,0 +1,24 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <!-- Agreement Form View -->
+    <record id="agreement_fsm_order_form_view" model="ir.ui.view">
+        <field name="name">agreement.form.fsm.order.view</field>
+        <field name="model">agreement</field>
+        <field name="inherit_id" ref="agreement_legal.partner_agreement_form_view"/>
+        <field name="arch" type="xml">
+            <notebook position="inside">
+                <page name="serviceprofiles" string="Service Profiles">
+                    <field name="serviceprofile_ids">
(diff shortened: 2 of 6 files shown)


agreement_serviceprofile/models/agreement.py
```diff
@@ -0,0 +1,13 @@
...
+
+from odoo import fields, models
+
+
+class Agreement(models.Model):
+    _inherit = 'agreement'
+
+    serviceprofile_ids = fields.One2many('agreement.serviceprofile',
+                                         'agreement_id',
+                                         default='_set_service_ids',
```
max3903: @osi-scampbell There is no need for a default value. When creating a service profile from the agreement, the views (tree and form) should not show the agreement field of the service profile. This is the exact same behavior between sale order and sale order line.

agreement_serviceprofile/views/agreement.xml
```diff
@@ -0,0 +1,20 @@
...
+        <field name="name">agreement.form.fsm.order.view</field>
+        <field name="model">agreement</field>
+        <field name="inherit_id" ref="agreement_legal.partner_agreement_form_view"/>
+        <field name="arch" type="xml">
+            <notebook position="inside">
+                <page name="serviceprofiles" string="Service Profiles">
+                    <field name="serviceprofile_ids">
+                        <tree>
+                            <field name="name"/>
+                        </tree>
```
max3903, line 14: @osi-scampbell You can do something like this:

```suggestion
                        </tree>
                        <form>
                            <field name="name"/>
                            <field name="notes"/>
                        </form>
```

osi-scampbell: > @osi-scampbell You can do something like this:
image

I think that is a great idea, when you pull up the form view from that One2many field it has <p><br></p> in notes, I think it is because of the html widget. I added a placeholder and that seems to fix that issue.

max3903 (approved):

patrickrwilson (approved):
I think it will be useful to add the notes to the tree view of the service profiles, and so it shows the notes in the list of the one2many field on the agreement. This would help users see more details when viewing a large list of service profiles on the agreements. Every thing else looks good and approving it even if the above suggestion doesn't make it in this version.

ygol (approved):
LGTM

@max3903 what is the use case for service profile? how is it used?

max3903:
@patrickrwilson can provide more details, but my understanding is that some products/services on a contract generates service profiles and it is used to track the performance of a part of a contract.

patrickrwilson:
@max3903 got it about right, service profiles help add supported services to the agreement. so the agreement itself contains the legal words and the products that were sold and the service profiles provide can add more descriptive services specific to support teams in which they can use on trouble tickets and such. Because they can be linked to other records like tickets, the performance can be measured in more detail on each supported service profile. A possible future roadmap item for service profiles can include configuration components that have been contracted for the customer that connected external systems can use but this still has to be designed and thought out.

PR #353, opened 2019-07-03
11.0 mig contract show invoice

thinkwelltwd:


rafaelbn:
Hi @thinkwelltwd , thanks for this contribution. Please could you please tell me why this module? I see is in Odoo v10 but testing I see we already have "show recurring invoices"



So this module is just a smart-button?

Maybe @angelmoya @amcor or @carlosdauden can clarify algo :smile: 

Thanks

pedrobaeza:
As stated before, this module shouldn't be here in this repo, but maybe on OCA/account-analytic, as it shows all the invoices that contains the analytic account.

thinkwelltwd:
> Please could you please tell me why this module?

1. Because it's a smart button that includes the total amount invoiced.
2. Because it's at the expected location where other smart buttons are, so the user doesn't have to look for it.

> As stated before, this module shouldn't be here in this repo,

Oh dear, I don't know just _where_ it has been stated before, or how contributors are expected to know it.  

So fine, I can yank the PR and maintain the module internally. But still it would be nice to have been apprised of such information before making the PR or at least without waiting 2 1/2 months for review on such a small PR.

pedrobaeza:
Sorry for the late reply, but I have discussed this with other contributor and I can't get track of everything.

thinkwelltwd:
> I can't get track of everything.

I understand, and hugely appreciate everything you do.  Before migrating a module, I try to check if the module is listed in github issue tracking the list of modules to migrate to the next version. At the very least it could be there.

Anyway, I'll close the PR. If you'll accept it elsewhere, let me know

pedrobaeza:
I think it could fit in OCA/account-analytic under the name `analytic_show_invoice`.

PR #354, opened 2019-07-17
[10.0][FIX] contract. Prevent wrong contract count.

NL66278:
Contract count could be wrong if the current date for the user is not the
UTC date.

So a contract created in for instance the Amsterdam timezone just after
midnight, would get maybe the 16th of july as date_start. but
_compute_contract_count would look for contracts valid on the 15th of july.

### DIFF
--- a/contract/models/res_partner.py
+++ b/contract/models/res_partner.py
@@ -14,13 +14,14 @@ class ResPartner(models.Model):
     )
 
     def _compute_contract_count(self):
-        Contract = self.env['account.analytic.account']
-        today = fields.Date.today()
+        contract_model = self.env['account.analytic.account']
+        # localized date is also the default for date_start.
+        today = fields.Date.context_today(contract_model)
         for partner in self:
-            partner.contract_count = Contract.search_count([
+            partner.contract_count = contract_model.search_count([
                 ('recurring_invoices', '=', True),
                 ('partner_id', '=', partner.id),
-                ('date_start', '<=', today),
+                ('date_start', '<=', today),  # required for contracts
                 '|',
                 ('date_end', '=', False),
                 ('date_end', '>=', today),


contract/models/res_partner.py
```diff
@@ -14,13 +14,14 @@ class ResPartner(models.Model):
     )
 
     def _compute_contract_count(self):
-        Contract = self.env['account.analytic.account']
-        today = fields.Date.today()
+        contract = self.env['account.analytic.account']
```
pedrobaeza: Please don't modify this variable name which is correct.

contract/models/res_partner.py
```diff
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 # Copyright 2017 Carlos Dauden <carlos.dauden@tecnativa.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
-
+# pylint: disable=missing-docstring
```
pedrobaeza: No need to put this in OCA linter

pedrobaeza (changes requested):

contract/models/res_partner.py
NL66278: Not according to the guidelines that say to use lowercase for variable names: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#id34

contract/models/res_partner.py
NL66278: It really helps in the editor to only see the real errors. But if there is a way to use to OCA linter in vim I would be glad to hear about it.

pedrobaeza: Yes, but it adds noise for the rest, so please remove it or use another linter. My editor doesn't highlight these things as I pass OCA pylint configuration to it, but I use VSCode.

contract/models/res_partner.py
pedrobaeza: Yeah, that's the only exception that is not explicitly written, but a real convention, even on Odoo core: as this represent the handler of the model, it's very similar in concept to python classes, and thus, they are put in upper case.

If you don't want to put it in upper, then you must use a meaningful variable name, like `contract_obj` or `contract_model`, which are longer, which is the reason for the other naming style, but not `contract`, as this can be interpreted as a record of that model.

pedrobaeza: And the most important thing: if you are not touching that line, better to leave it as is for reducing the diff, even if it's not exactly your style. Note that Travis was green even with this variable name, as it's not checked due to the reasons above.

contract/models/res_partner.py
NL66278: Even using pylint with the pylint_odoo plugin still complains about the missing docstring. So I really think for most people this is supressing a useless message.

pedrobaeza: Then are you going to add this line in all the files you touch since now? This is the same need as the old line
```
# vim:expandtab:smartindent:tabstop=4:softtabstop=4:shiftwidth=4:
```
that we suffered from OpenERP during several versions.
Please don't do that.
And even more, you are not adding a file, but only a bug fixing that is a one line diff that you are converting to 6-7...

NL66278:
I removed the pylint line, so only the change to the misbehaving function remains. I also followed the suggestion to use contract_model instead of contract.

pedrobaeza (approved):

rafaelbn (approved):

Issue #349, opened 2019-06-14
[IMP] agreement_serviceprofile: Add a state field


max3903:
# agreement.serviceprofile

- [x] Add a selection field State (state) with Draft, In Progress, Suspended, To Renew, Terminated, Cancelled
- [x] Add a many2one field "Service" (product_id) to product.template. The list of products is filtered with a domain: type = service.
- [x] Add the clickable workflow on the form view to move from one state to another
- [x] Add a Kanban view in the Dashboard menu

max3903:
FYI @patrickrwilson @ygol

brian10048:
@max3903 should the state be configurable and create a new object for it with the states you mentioned as data?

max3903:
@brian10048 I guess you are asking because you have a different workflow... so yes!

PR #355, opened 2019-07-18
[12.0][IMP] agreement_serviceprofile: add stage

brian10048:
- Add stage_type to agreement.stage
- Track stage of service profile
- Add Kanban view for service profiles to the dashboard
- Add product_id field to service profile

@max3903 for #349

### DIFF
--- a/agreement_serviceprofile/models/agreement_serviceprofile.py
+++ b/agreement_serviceprofile/models/agreement_serviceprofile.py
@@ -2,14 +2,20 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 
-from odoo import fields, models
+from odoo import fields, models, api
 
 
 class AgreementServiceProfile(models.Model):
     _name = 'agreement.serviceprofile'
     _description = 'Agreement Service Profiles'
 
+    def _default_stage_id(self):
+        return self.env.ref('agreement_serviceprofile.servpro_stage_draft')
+
     name = fields.Char(string="Name", required=True)
+    stage_id = fields.Many2one('agreement.stage', string="Stage",
+                               default=_default_stage_id, copy=False,
+                               group_expand='_read_group_stage_ids',)
     agreement_id = fields.Many2one('agreement', string="Agreement",
                                    ondelete="cascade")
     active = fields.Boolean(string="Active",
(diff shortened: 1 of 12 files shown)


brian10048:
Please excuse the mess! 

@max3903, should I re-use the agreement.stage and add a selection type field for service profile stages like was done with field service stages?

max3903:
Yes please.

max3903:
@brian10048 I added a product_id field on the design with a domain. Can you implement it please?

max3903 (approved):

max3903:
Ping @kvedder-amplex @barnarj

brian10048:
@max3903 I reworked the stage to use the agreement.stage model

> 
> @brian10048 I added a product_id field on the design with a domain. Can you implement it please?

I'm trying to figure out what you mean on the design.  Could you explain a little more maybe?

max3903:
@brian10048 The service profile is the same to the service product as the serial number to the stockable product.

Screenshot from 2019-07-12 13-51-07

So with a product_id field on the service profile, we can access all the service information defined on the product template.

brian10048:
Thank you @max3903. That helps

Also, not sure what I am missing.  The dashboard kanban view for service profiles is not showing the stages unless there is a record for that stage

image

agreement_serviceprofile/models/agreement_serviceprofile.py
```diff
@@ -2,14 +2,19 @@
...
 class AgreementServiceProfile(models.Model):
     _name = 'agreement.serviceprofile'
     _description = 'Agreement Service Profiles'
 
+    def _default_stage_id(self):
+        return self.env.ref('agreement_serviceprofile.servpro_stage_draft')
+
     name = fields.Char(string="Name", required=True)
+    stage_id = fields.Many2one('agreement.stage', string="Stage",
+                               default=_default_stage_id, copy=False)
```
max3903:
```suggestion
                               default=_default_stage_id, copy=False,
                               group_expand='_read_group_stage_ids',)
```

max3903: @brian10048 That should open all the stages, even without records.

brian10048: Thank you.  I added the method, but forgot that part in the field definition!

brian10048:
@max3903 I believe this is done now.  Cleaned up into one commit.  Let me know if there is anything else I need to do.

max3903:
:+1: LGTM. Tested on Runbot.

bodedra (approved):
LTGM :+1: Good work @brian10048 

Not blocking comment: IMO for better maintain and commit history, shouldn't we have two PRs for these ? one for `agreement_legal` and another for `agreement_serviceprofile`

PR #357, opened 2019-07-26
[12.0] Migrate subscription module from official addons v10 to v12

alexis-via:
The subscription modules was part of Odoo Community until Odoo v10 included. As I still have some odoo deployments using it, I ported it to v12 and I propose it here.
The first commit corresponds to the module from v10 and I forced the author to Odoo SA.


rafaelbn:
Hello @alexis-via maybe you could check bots to get them green :green_heart:

subscription/demo/subscription_demo.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" ?>
+<odoo noupdate="1">
+
+    <record id="subscription_document_partner0" model="subscription.document">
+        <field name="model" ref="base.model_res_partner"/>
+        <field name="name">Partner</field>
+    </record>
+
+    <record id="subscription_document_fields_2" model="subscription.document.fields">
+        <field name="field" ref="base.field_res_partner_name"/>
```
sbidoul, line 10:
```suggestion
        <field name="field" ref="base.field_res_partner__name"/>
```

subscription/demo/subscription_demo.xml
```diff
@@ -0,0 +1,20 @@
...
+        <field name="value">false</field>
+        <field model="subscription.document" name="document_id" search="[('name', '=', u'Partner')]"/>
+    </record>
+
+    <record id="subscription_document_fields_3" model="subscription.document.fields">
+        <field name="field" ref="base.field_res_partner_child_ids"/>
```
sbidoul, line 16:
```suggestion
        <field name="field" ref="base.field_res_partner__child_ids"/>
```

sbidoul (commented):
@alexis-via thanks for this migration. This is useful for simple cases.
There is a little issue with demo data, and a lint error.
Otherwise LGTM, I made a quick functional test.

subscription/demo/subscription_demo.xml
alexis-via: Fixed in akretion:12-subscription (but not shown here because the PR was closed in the meantime)

subscription/demo/subscription_demo.xml
alexis-via: Fixed in akretion:12-subscription (but not shown here because the PR was closed in the meantime)

alexis-via:
Moved to a new dedicated project https://github.com/akretion/subscription

PR #360, opened 2019-08-05
#356 fix the cron recurring invoice creation in a multi-company setup

### LINKED ISSUE #356: [10.0][BUG] contract: 'cron_recurring_create_invoice' doesn't consider a multi company setup correctly
We are running a multi company setup of odoo and a couple times we've experienced an issue where the recurring invoices were not created even though they should be (according to the``recurring_next_date``).

So I had a little peek at the contract module and it looks like the ``cron_recurring_create_invoice`` method in the **account.analytic.account** model doesn't consider a multi company setup correctly.

**Body of the method:**
```python
today = datetime.today()
company_model = self.env['res.company']
for company in company_model.search([]):
    days_before = company.contract_pregenerate_days or 0
    cutoffdate = (
        today + relativedelta(days=days_before)
    ).strftime(DEFAULT_SERVER_DATE_FORMAT)
    contracts = self.with_context(cron=True).search([
        ('company_id', '=', company.id),
        ('recurring_invoices', '=', True),
        ('recurring_next_date', '<=', cutoffdate),
        '|',
        ('date_end', '=', False),
        ('date_end', '>=', cutoffdate),
     ])
return contracts.recurring_create_invoice(limit)
```
Because the recurring invoices are created after the loop  in the return statement, the ``contracts`` variable will only contain the contract records of the last company.

So if we have three companies:
1. company one (has recurring invoices)
2. company two (doesn't have recurring invoices)
3. company three (doesn't have recurring invoices)

No recurring invoices will be created.

0xA537FD:


pedrobaeza:
I don't get the problem and why your patch fixes it.

rafaelbn:
Hello @Peter-Morawski , please could you give steps to reproduce you issue in runbot?

rafaelbn:
Steps in https://github.com/OCA/contract/issues/356

contract/models/account_analytic_account.py
```diff
@@ -352,19 +352,21 @@ def cron_recurring_create_invoice(self, limit=None):
...
         today = datetime.today()
         company_model = self.env['res.company']
+        contracts = self.with_context(cron=True)
         for company in company_model.search([]):
             days_before = company.contract_pregenerate_days or 0
             cutoffdate = (
                 today + relativedelta(days=days_before)
             ).strftime(DEFAULT_SERVER_DATE_FORMAT)
-            contracts = self.with_context(cron=True).search([
+            contracts = contracts.concat(self.with_context(cron=True).search([
```
pedrobaeza, line 361:
```suggestion
            contracts |= self.search([
```

pedrobaeza (changes requested):
Or better...

1. Directly call on each loop to the method for packing only contracts for one company.
2. On the search, include `'company_id', 'in', company_model.search([])`.

PR #361, opened 2019-08-13
[12.0][WIP][ADD] module contract_commercial_partner

hparfr:


hparfr:
ping @alexis-via

pedrobaeza:
Isn't better to include this in base module? What do you think, @sbidoul ? Any way, maybe it should be included on BI report table, not on regular one.

hparfr:
>Isn't better to include this in base module?

I'm ok for merging it in contract module.
I don't see any good reason to not use this module and all its siblings ({sale|purchase}_commercial_partner).

sbidoul:
> Isn't better to include this in base module?

It makes sense to me.

alexis-via:
Yes, let's put it in the contract module.
I developed the (sale|purchase)_commercial_partner modules because it was not possible to add this field in the sale|purchase modules from the officials addons. But, when the base module is an OCA module, it should be in the base module.

hparfr:
Merged in contract module in PR #363 
I based my PR on #207 which seams close to be merged.

PR #363, opened 2019-08-14
[12.0[IMP] contract add commercial partner

hparfr:
Add commercial_partner_id to contract.

Replaces #361 
Depends of #207 

- [x] https://github.com/OCA/contract/pull/207

### DIFF
--- a/contract/views/contract.xml
+++ b/contract/views/contract.xml
@@ -43,6 +43,7 @@
                     </div>
                     <group name="main">
                         <group>
+                            <field name="commercial_partner_id" invisible="1"/>
                             <field name="partner_id" required="1"/>
                             <field name="payment_term_id"/>
                             <field name="user_id"/>
@@ -189,6 +190,7 @@
                         domain="[('date_end', '&lt;', context_today().strftime('%Y-%m-%d')), ('recurring_next_date', '=', False)]"
                 />
                 <field name="partner_id"/>
+                <field name="commercial_partner_id"/>
                 <filter string="Archived"
                         domain="[('active', '=', False)]"
                         name="inactive"/>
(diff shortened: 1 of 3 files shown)


pedrobaeza:
Something seems not correct looking commit history...

pedrobaeza:
OK, now I see that this contains commit history from #207

hparfr:
207 is merged. I did some git merges to keep only the changes without any force.

So the history is currently messy,  but the diff change is minimal. It should work well with a rebase on merge.

contract/views/contract.xml
```diff
@@ -189,6 +190,9 @@
                         domain="[('date_end', '&lt;', context_today().strftime('%Y-%m-%d')), ('recurring_next_date', '=', False)]"
                 />
                 <field name="partner_id"/>
+                <field name="commercial_partner_id"
+                        string="Commerical Entity"
```
sbidoul: A spelling issue here. The string attribute could be removed.

sbidoul (commented):
Looks good at first sight.

pedrobaeza (changes requested):
No, it won't work. Please fix the commit history starting from current branch and cherry-picking your commit, or redo the diff, but this can't be merged as is.

hparfr:
@pedrobaeza done as requested.

contract/views/contract.xml
hparfr: Good catch

pedrobaeza:
Thanks, can you please squash both commits together (our bot doesn't still do this)?

hparfr:
@pedrobaeza, it's now squashed in one commit

pedrobaeza (approved):

carlosdauden (approved):

PR #364, opened 2019-08-14
[10.0][FIX] tests not working Rename folder test to tests to working Add test to show that actual implementation do not work, to be fixed in https://github.com/OCA/contract/pull/315

sergiocorato:


pedrobaeza:
We can't merge this branch if this is going to be red. Put both commits in the same PR.

sergiocorato:
> We can't merge this branch if this is going to be red. Put both commits in the same PR.

yeah, this was only to show bug, I'll close this PR, anyway there is something wrong with runbot i think as it gives a postgres error

PR #374, opened 2019-09-09
contract: Add "Generate before" (days) optional field

gaikaz:
It is sometimes required to be able to generate an invoice some amount of days before the actual date on the invoice itself. Before this PR, `contract` module doesn't allow this while auto-generating. 
This PR adds an optional integer field `generate_before` that is checked by the cron job, to allow this functionality.

image

### DIFF
--- a/contract/models/account_analytic_account.py
+++ b/contract/models/account_analytic_account.py
@@ -7,6 +7,7 @@
 
 from dateutil.relativedelta import relativedelta
 
+import datetime as date
 from odoo import api, fields, models
 from odoo.exceptions import ValidationError
 from odoo.tools.translate import _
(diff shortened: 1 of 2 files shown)


rafaelbn:
Dear @gaikaz , please review https://github.com/OCA/contract/pull/207 the contract module has been refactored totally with lots of improvements. So this PR don't apply, sorry. Thanks

contract/models/account_analytic_account.py
```diff
@@ -53,6 +54,22 @@ class AccountAnalyticAccount(models.Model):
     create_invoice_visibility = fields.Boolean(
         compute='_compute_create_invoice_visibility',
     )
+    generate_before = fields.Integer(
+        string="Generate before",
```
quannguyen1560, line 58: I think the system will automatically use the name of the field so we don't need to define the string for the fields again, so please remove the string.

contract/models/account_analytic_account.py
```diff
@@ -53,6 +54,22 @@ class AccountAnalyticAccount(models.Model):
...
+        help="Days before how many to generate invoice",
+    )
+    generate_invoice_date = fields.Date(
+        compute='_compute_generate_before_date',
+        copy=False,
+        store=True,
+        string='Generate Invoice Date',
```
quannguyen1560, line 65: same above

contract/models/account_analytic_account.py
```diff
@@ -7,6 +7,7 @@
 
 from dateutil.relativedelta import relativedelta
 
+import datetime as date
```
quannguyen1560, line 10: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#imports

I think we should have a extra line between library imports and odoo imports

quannguyen1560 (commented):
Please have a check about my comment and fix it if possible

gaikaz:
@quannguyen1560 It's a nice review you did and I can do the changes, but I haven't had time to check the newest version of this module, that got a huge update after #207 , that Rafael mentioned. Now this PR now has conflicts. Probably should have closed this...

PR #375, opened 2019-09-10
[11.0][FIX] contract. Prevent wrong contract count.

NL66278:
This is a cherry-pick of a merged fix for 10.0. However some of the logic in contract counts changed (now separate accounts for purchase and sale contracts). In this change I think the start_date was accidentally admitted from the domain. I put it back.

Contract count could be wrong if the current date for the user is not the
UTC date.

So a contract created in for instance the Amsterdam timezone just after
midnight, would get maybe the 16th of july as date_start. but
_compute_contract_count would look for contracts valid on the 15th of july.

### DIFF
--- a/contract/models/res_partner.py
+++ b/contract/models/res_partner.py
@@ -18,10 +18,12 @@ class ResPartner(models.Model):
 
     def _compute_contract_count(self):
         contract_model = self.env['account.analytic.account']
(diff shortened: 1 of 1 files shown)


pedrobaeza:
I would say that current logic is OK, as we want to count all the contracts this partner has, not only those that have "started".

PR #383, opened 2019-09-27
[IMP] move some notions from agreement_legal to agreement

### LINKED ISSUE #382: move some concepts out of agreement_legal
For a development I'm working on, I need some concepts on agreements that are introduced by agreement_legal, but without all the bells and whistles that are provided by that module. 

mainly:

* the agreement.type model (without necessary the subtypes)
* the is_template field on agreement

I propose making a PR which would move these definitions in the base agreement module on which agreement_legal depends. Alternatively, I could create a new module, but it seems a pity to tackle adding a dependency in agreement_legal.

gurneyalex:
we move the `is_template` field definition and the `agreement.type` model from
the `agreement_legal` module to the `agreement` module.

The fields are not displayed by default, unless the feature is enabled through a
technical feature group, this is configurable in the `agreement_sale` module (because agreement
in itself has no UI, and `agreement_legal` enables the feature by default)

Closes: #382

### DIFF
--- a/agreement/models/agreement_type.py
+++ b/agreement/models/agreement_type.py
@@ -5,11 +5,8 @@ from odoo import fields, models
 
 
 class AgreementType(models.Model):
-    _inherit = "agreement.type"
+    _name = "agreement.type"
     _description = "Agreement Types"
 
-    agreement_subtypes_ids = fields.One2many(
-        "agreement.subtype",
-        "agreement_type_id",
-        string="Subtypes"
-    )
+    name = fields.Char(string="Name", required=True)
+    active = fields.Boolean(default=True)
--- a/agreement_legal/views/agreement_type.xml
+++ b/agreement_legal/views/agreement_type.xml
@@ -5,33 +5,28 @@
     <record model="ir.ui.view" id="partner_agreement_type_list_view">
         <field name="name">Agreement Type List</field>
         <field name="model">agreement.type</field>
+        <field name="inherit_id" ref="agreement.agreement_type_list_view"/>
         <field name="arch" type="xml">
-            <tree string="Agreement Types" default_order='name'>
-                <field name="name" string="Type Name"/>
+            <field name="name" position="after">
                 <field name="agreement_subtypes_ids" string="Sub-Types"/>
-            </tree>
-        </field>
+            </field>
+       </field>
     </record>
 
     <!-- Agreement Type Form View -->
     <record model="ir.ui.view" id="partner_agreement_type_form_view">
         <field name="name">Agreement Type Form</field>
         <field name="model">agreement.type</field>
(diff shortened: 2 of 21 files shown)


max3903 (approved):
LGTM. No Test

bealdav:
No security rule for agreement.type when agreement_legal not installed.

Then impossible to complete data

Screenshot from 2019-09-30 09-47-10

pending evaluation ...

agreement_legal/views/agreement_type.xml
```diff
@@ -5,33 +5,28 @@
...
-                </sheet>
-            </form>
+            <xpath expr="//sheet" position="inside">
+                <field name="agreement_subtypes_ids"
+                       nolabel="1">
+                    <tree editable="bottom">
+                        <field name="name"/>
+                    </tree>
+                </field>
+            </xpath>
```
bealdav, line 29: @gurneyalex Partially off-topic of your PR but there is a missing part in agreement_sale as documented in the readme "This module adds Agreement field to Sales Orders"
Please could you move this view
https://github.com/OCA/contract/blob/12.0/agreement_legal_sale/views/sale_order.xml
to agreement_sale
But please add a group to agreement_template_id field please.
Thanks

bealdav (changes requested):
Ok on my side, except the 2 points explained.

Nice work. 

With these modules agreement and agreement_sale now serves at least 3 uses cases (c2c, opensource integrators and akretion)
OCA proves one more time they can managed modularity across several integrators.

Thanks a lot

bealdav:
Please @alexis-via could you have a look here for an approval, please

bealdav:
Other element found 

is_template field is visible

agree

Whatever settings on it. Is it the behavior you wanted to implement ?

tmp

gurneyalex:
thanks for the review @bealdav I'll take care ot this tomorrow.

bealdav:
second review @gurneyalex, element https://github.com/OCA/contract/pull/383#issuecomment-536458890 not solved but maybe it's intentional ?

gurneyalex:
@bealdav All your comments have been taken into account. If this is ok with you we can probably merge and move on :)

bealdav (approved):
Runbot is broken, but ok for me
Thanks a lot

agreement/models/agreement_type.py
```diff
@@ -0,0 +1,12 @@
+# Copyright (C) 2018 - TODAY, Pavlov Media
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class AgreementType(models.Model):
+    _name = "agreement.type"
```
i-vyshnevska, line 8: @gurneyalex  you need an access rule for type

PR #387, opened 2019-10-01
[12.0] [ADD] Contract Variable Qty Prorated

### LINKED ISSUE #197: Prorated invoicing of contracts
Hi,

I am currently encountering the following use case that I can't (or don't know how to) solve with the current modules. Any ideas on how to deal with this use case or any opinion on the proposed new feature?

First, I would like to invoice my contracts on the last day of each month (recurring_interval set to 'Month(s) last day'). 
But, I would like to invoice my contract based on a pro rata percentage if the contract didn't start / end on the first/ respectively last day of a month. So I would like to calculate the amount I charge to my customer based on the amount of days the contract was effectively active during the month. With the current module, a full fee is for example charged to the customer at the end of February, even though the contract only started on the 24th of February. I would like to charge my customer for only 5 days (from the 24th to the 28th of February), not for a full month.

In our opinion, the following feature/module should be added:

- Adding a 'prorated invoicing' checkbox on a contract template and on the contract lines. 
- This box can only be checked when the 'recurring interval field' is set to 'Month(s) last day'
- The 'prorated invoicing' checkbox on the contract line is automatically filled with the information coming from the contract template, but can be modified manually by the user on each contract line.

If this checkbox is checked, Odoo will apply a special 'pro rata rule' to calculate the amount to add on the invoice for the first and the last month of the contract (only!). The other months of the contract are invoiced normally (full amount). 

One of the two following options should be chosen to calculate the pro rata percentage (that should be multiplied with the full amount to calculate the amount that will be charged to the customer for its first month of contract):
1.  Amount of days the contract was active (during the first month) divided by the amount of days we have in the month. So if the contract starts on the 24th of February, the pro rata percentage when the bill will be generated (on the 28th of February) will be (5/28 = 17,86%). So only 17,86% of the total amount will be charged to the customer for February.
2. Amount of days the contract was active (during the first month) divided by a fixed amount of days (independently of the amount of days we have in the month). I divided 365 by 12 to obtain a plausible amount (= 30,42). Hence, for my example, we would have (5/30,42=16,44 %), so 16,44% of the total amount will be charged to the customer for February. This second method is easier as it doesn't take the amount of days of the month into account.

For the last month of the contract, we could either apply one of these 2 rules (the same as the one we chose for the first month) to calculate the pro rata percentage or we could decide to take the residual amount that was not charged to the customer on the first month and charge this amount to the customer on the last month (only if contracts cannot be cancelled before a full amount of months has been reached). 

Any ideas / opinion on this?

Thanks

ThomasBinsfeld:
fixes: #197


contract_variable_qty_prorated/models/contract_line.py
```diff
@@ -0,0 +1,45 @@
...
+
+    @api.model
+    def _compute_prorated(
+        self, real_next_date, real_last_date, therical_next_date,
+            therical_last_date):
+        def _invoiced_days(next_date, last_date):
+            return (next_date - last_date).days + 1
+
+        return _invoiced_days(real_next_date, real_last_date) / _invoiced_days(
+            therical_next_date, therical_last_date
```
rousseldenis: s/therical/theoretical ?

rousseldenis (changes requested):
LGTM. 

But could you add a usage section that describe a little bit more the flow?

sbidoul:
I fixed the spelling typo mentioned by @rousseldenis and rebased.

alexis-via (approved):
Tested, work fine.

alexis-via:
This feature should not be in a separate module. It should be inside the "contract_variable_quantity". And the feature of the module "contract_variable_quantity" should be in the "contract" module itself. Computing the qty pro-rata temporis seems like a basic feature of a contract module; we should not have to look for several other modules to have that feature. And it should be the default behavior ; you should not have to configure something to have pro-rata-temporis. Maybe it would be an idea for v14...

sbejaoui (approved):

gva-acsone (approved):
LGTM Functional test

PR #392, opened 2019-10-14
[ADD] 12.0 agreement_sale_condition_base

gurneyalex:
This module builds on the agreement and agreement_sale modules. It defines a
model for agreement parameter values. The intent is that additional modules can
add fields on an agreement which are m2o on to parameter values, and possibly
the allowed values for a given field could depend on the value of another one.

An agreement type can be linked to a template agreement. When the agreement
type is set on a sale order, a new agreement record is created using the
template agreement as a basis. This agreement specific to the sale order can be
customized by the sales man.

When the sale order is confirmed, the agreement of the sale order is propagated
to the procurement group of the sale order. This can be used by extension
modules to customize the generation of the stock moves and pickings to match
the agreement.

In a similar fashion, when an invoice is created from the sale order, the
agreement of the sale order, the agreement is propagated to the invoice.

depends on #383 (included to ease testing)


gurneyalex:
I don't understand why the lint tests are red :/

max3903:
https://travis-ci.org/OCA/contract/jobs/597549548#L311

agreement_sale_condition_base/demo/agreement.csv
```diff
@@ -0,0 +1,5 @@
+id,code,name,partner_id/id,agreement_type_id/id,is_template
+agreement_sale_condition_base.agreement_next_day,next_day_tmpl,Template Standard Next Day,base.main_company,agreement_sale_condition_base.agreement_type_next_day,True
+agreement_sale_condition_base.agreement_weekly_routine,weekly_routine_tmpl,Template Standard Weekly Routine,base.main_company,agreement_sale_condition_base.agreement_type_weekly_routine,True
+agreement_sale_condition_base.agreement_agreed_date,agreed_date_tmpl,Template Standard Agreed Date,base.main_company,agreement_sale_condition_base.agreement_type_agreed_date,True
+agreement_sale_condition_base.agreement_pick_up,pickup_tmpl,Template Pick Up,base.main_company,agreement_sale_condition.agreement_type_pick_up,True
```
guewen:
```suggestion
agreement_sale_condition_base.agreement_pick_up,pickup_tmpl,Template Pick Up,base.main_company,agreement_sale_condition_base.agreement_type_pick_up,True
```

PR #401, opened 2019-10-23
[12.0][FIX] - contract_mandate: fix inherited method call _prepare_invoice

sbejaoui:
there is an inconsistency between the inherited method and base one

look [here](https://github.com/OCA/contract/blob/12.0/contract/models/contract.py#L256)

### DIFF
--- a/contract_mandate/models/contract.py
+++ b/contract_mandate/models/contract.py
@@ -34,7 +34,7 @@ class ContractContract(models.Model):
     @api.multi
     def _prepare_invoice(self, date_invoice, journal=None):
         invoice_vals = super(ContractContract, self)._prepare_invoice(
-            date_invoice, journal
+            date_invoice, journal=journal
         )
         if self.mandate_id:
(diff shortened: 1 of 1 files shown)


contract_mandate/models/contract.py
```diff
@@ -32,8 +32,10 @@ def _onchange_payment_mode_id(self):
             self.mandate_id = False
 
     @api.multi
-    def _prepare_invoice(self, date_ref=False):
-        invoice_vals = super(ContractContract, self)._prepare_invoice(date_ref)
+    def _prepare_invoice(self, date_invoice, journal=None):
+        invoice_vals = super(ContractContract, self)._prepare_invoice(
+            date_invoice, journal
```
pedrobaeza: Please respect always keyword arguments:
```suggestion
            date_invoice, journal=journal
```

pedrobaeza (approved):

sbidoul (approved):

PR #405, opened 2019-10-28
[12.0][FIX] - When creating a contract from a sale order, contract company must be the same as the sale order

sbejaoui:
When creating a contract from a sale order the company must be the sale order
company and not the user company

### DIFF
--- a/product_contract/models/sale_order.py
+++ b/product_contract/models/sale_order.py
@@ -46,6 +46,7 @@ class SaleOrder(models.Model):
                 template_name=contract_template.name, sale_name=self.name
             ),
             'partner_id': self.partner_id.id,
+            'company_id': self.company_id.id,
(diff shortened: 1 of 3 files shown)


product_contract/models/sale_order.py
```diff
@@ -46,6 +46,7 @@ def _prepare_contract_value(self, contract_template):
                 template_name=contract_template.name, sale_name=self.name
             ),
             'partner_id': self.partner_id.id,
+            'company_id': self.compnay_id.id,
```
pedrobaeza: Typo

sbejaoui:
```suggestion
            'company_id': self.company_id.id,
```

sbidoul (approved):

sbidoul:
squashed

PR #402, opened 2019-10-24
[12.0][IMP] - Add contract tags

sbejaoui:

PR #412, opened 2019-11-04
[12.0][ADD] - Add note field to contract

sbejaoui:


pedrobaeza:
Shouldn't we develop the link between `agreement_legal` and `contract` instead of providing alternative text fields that serves for the same purpose?

sbejaoui:
> Shouldn't we develop the link between `agreement_legal` and `contract` instead of providing alternative text fields that serves for the same purpose?

IMO, we can add this field to 12.0 As the link to agreement and the refactoring is planned only for 13.0 only

pedrobaeza:
For 13.0, the refactoring is planned, but I think it's worse to convert the notes field to the agreements on migration than to provide a module `contract_agreement_legal` that makes already the link and on migration to 13.0, merge that module with the new base and handle the link properly, don't you think?

sbidoul:
I agree we should anticipate the merge of the two models.

In this specific case I would not force pulling `agreement_legal` (which is quite big) just have a note field.

However, we could add `note` to both `agreement` (the base module) and `contract`, to facilitate migration. The same could be done for tags (#402).

sbejaoui:
@sbidoul , @pedrobaeza 

As the work for merging agreement and contract is planned for V13 (and is not moving forward for the moment), can we merge this and see what to do when the time came for V13.

sbejaoui:
> @sbidoul , @pedrobaeza
> 
> As the work for merging agreement and contract is planned for V13 (and is not moving forward for the moment), can we merge this and see what to do when the time came for V13.

@pedrobaeza , @sbidoul 

Did you considered my proposal?

pedrobaeza (approved):
OK, this seems very harmless.

gva-acsone (approved):
LGTM +1 Functional test

PR #414, opened 2019-11-13
[12.0][IMP] - Add an action for contracts manual invoicing

sbejaoui:
It happen that a company has to trigger the invoicing action to generate invoices before
the scheduled date (to print and prepare invoices documents, check invoices, etc.).
This requires technical access for end users with the risk that it represent.

This PR adds a new wizard to run the invoicing action for a given date with a helper
to see and check the contract that will be invoiced. When the manual action is called,
the system displays all created invoices.

### DIFF
--- a/contract/models/contract.py
+++ b/contract/models/contract.py
@@ -457,7 +457,17 @@ class ContractContract(models.Model):
         This method triggers the creation of the next invoices of the contracts
         even if their next invoicing date is in the future.
         """
-        return self._recurring_create_invoice()
+        invoice = self._recurring_create_invoice()
+        if invoice:
+            self.message_post(
+                body=_(
+                    'Contract manually invoiced: '
+                    '<a href="#" data-oe-model="%s" data-oe-id="%s">Invoice'
+                    '</a>'
+                )
+                % (invoice._name, invoice.id)
+            )
+        return invoice
 
     @api.multi
     def _recurring_create_invoice(self, date_ref=False):
@@ -465,8 +475,9 @@ class ContractContract(models.Model):
         return self._finalize_and_create_invoices(invoices_values)
 
     @api.model
-    def cron_recurring_create_invoice(self):
-        domain = self._get_contracts_to_invoice_domain()
+    def cron_recurring_create_invoice(self, date_ref=None):
+        if not date_ref:
+            date_ref = fields.Date.context_today(self)
+        domain = self._get_contracts_to_invoice_domain(date_ref)
         contracts_to_invoice = self.search(domain)
-        date_ref = fields.Date.context_today(contracts_to_invoice)
-        contracts_to_invoice._recurring_create_invoice(date_ref)
+        return contracts_to_invoice._recurring_create_invoice(date_ref)
--- a/contract/wizards/contract_manually_create_invoice.py
+++ b/contract/wizards/contract_manually_create_invoice.py
@@ -1,0 +1,66 @@
+# Copyright 2019 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models, _
+
+
+class ContractManuallyCreateInvoice(models.TransientModel):
+
+    _name = 'contract.manually.create.invoice'
+    _description = 'Contract Manually Create Invoice Wizard'
+
+    invoice_date = fields.Date(string="Invoice Date", required=True)
+    contract_to_invoice_count = fields.Integer(
+        compute="_compute_contract_to_invoice_ids"
+    )
+    contract_to_invoice_ids = fields.Many2many(
+        comodel_name="contract.contract",
+        compute="_compute_contract_to_invoice_ids",
+    )
+    contract_type = fields.Selection(
+        selection=[('sale', 'Customer'), ('purchase', 'Supplier')],
+        default='sale',
+        required=True,
+    )
+
+    @api.depends('invoice_date')
+    def _compute_contract_to_invoice_ids(self):
+        if not self.invoice_date:
+            # trick to show no invoice when no date has been entered yet
+            contract_to_invoice_domain = [('id', '=', False)]
+        else:
+            contract_to_invoice_domain = self.env[
+                'contract.contract'
+            ]._get_contracts_to_invoice_domain(self.invoice_date)
+        self.contract_to_invoice_ids = self.env['contract.contract'].search(
+            contract_to_invoice_domain
+            + [('contract_type', '=', self.contract_type)]
+        )
+        self.contract_to_invoice_count = len(self.contract_to_invoice_ids)
+
+    @api.multi
+    def action_show_contract_to_invoice(self):
+        self.ensure_one()
+        return {
+            "type": "ir.actions.act_window",
+            "name": _("Contracts to invoice"),
+            "res_model": "contract.contract",
+            "domain": [('id', 'in', self.contract_to_invoice_ids.ids)],
+            "view_mode": "tree,form",
+            "context": self.env.context,
+        }
+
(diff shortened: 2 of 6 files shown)


contract/wizards/contract_manually_create_invoice.py
```diff
@@ -0,0 +1,62 @@
+# Copyright 2019 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models, _
+
+
+class ContractManuallyCreateInvoice(models.Model):
```
pedrobaeza: why not `TransientModel` and you save the extra permission?

pedrobaeza (changes requested):

contract/wizards/contract_manually_create_invoice.py
sbejaoui: I would like to avoid that this action is available for all accounting users.

pedrobaeza: Do that other way, adding a group on a menu or a button, but not this way.

sbejaoui: another advantage of making it a Model is to keep track for all manual runs in the database.

pedrobaeza: I don't see that as any advantage, as that's not browseable. If you want that, put a log entry in the chatter of the contract (with, in other hand, would be great for traceability in general))

sbejaoui: @pedrobaeza , I applied you suggestion.

Please review my last commit.

sbejaoui:
FYI @gva-acsone

contract/models/contract.py
```diff
@@ -450,6 +450,10 @@ def recurring_create_invoice(self):
         This method triggers the creation of the next invoices of the contracts
         even if their next invoicing date is in the future.
         """
+        self.message_post(
+            body=_("Manually invoiced. Invoice date: %s")
+            % self.recurring_next_date
```
pedrobaeza: IMO you should include here the link to the invoice instead of the date

contract/wizards/contract_manually_create_invoice.py
```diff
@@ -0,0 +1,71 @@
...
+
+    @api.depends('invoice_date')
+    def _compute_contract_to_invoice_count(self):
+        self.ensure_one()
+        if not self.invoice_date:
+            self.contract_to_invoice_count = 0
+        else:
+            contract_model = self.env['contract.contract']
+            self.contract_to_invoice_count = contract_model.search_count(
+                contract_model._get_contracts_to_invoice_domain(
```
pedrobaeza: For not having to do the same operation twice (DRY principle), create instead a m2m non stored computed field with this result, and make the count computed field to depend on this one, getting `len` of it.

sbejaoui:
Hi @pedrobaeza , 

Can you update your review please

gva-acsone (approved):
LGTM Functional test

sbejaoui:
@pedrobaeza , we need your review here please

contract/models/contract.py
```diff
@@ -452,16 +452,27 @@ def recurring_create_invoice(self):
         This method triggers the creation of the next invoices of the contracts
         even if their next invoicing date is in the future.
         """
-        return self._recurring_create_invoice()
+        invoice = self._recurring_create_invoice()
+        if invoice:
+            self.message_post(
+                body=_(
+                    'Contract manually invoiced: '
```
pedrobaeza, line 464: This can be triggered manually or by the cron, so the text `Contract manually invoiced:` is not correct.

contract/wizards/contract_manually_create_invoice.py
```diff
@@ -0,0 +1,59 @@
...
+    )
+    contract_to_invoice_ids = fields.Many2many(
+        comodel_name="contract.contract",
+        compute="_compute_contract_to_invoice_ids",
+    )
+
+    @api.depends('invoice_date')
+    def _compute_contract_to_invoice_ids(self):
+        if not self.invoice_date:
+            contract_to_invoice_domain = [('id', '=', False)]
```
sbidoul, line 30: Perhaps add a comment here (trick to show no invoice when no date has been entered yet)

contract/models/contract.py
```diff
@@ -452,16 +452,27 @@ def recurring_create_invoice(self):
...
-    def cron_recurring_create_invoice(self):
-        domain = self._get_contracts_to_invoice_domain()
+    def cron_recurring_create_invoice(self, date_ref=None):
+        if not date_ref:
+            date_ref = fields.Date.context_today(self)
+        domain = self._get_contracts_to_invoice_domain(date_ref)
         contracts_to_invoice = self.search(domain)
-        date_ref = fields.Date.context_today(contracts_to_invoice)
-        contracts_to_invoice._recurring_create_invoice(date_ref)
+        return contracts_to_invoice._recurring_create_invoice(date_ref)
```
sbidoul, line 483: why adding `return` here?

contract/wizards/contract_manually_create_invoice.xml
```diff
@@ -0,0 +1,59 @@
...
+        <field name="target">new</field>
+    </record>
+
+    <record model="ir.ui.menu" id="contract_manually_create_invoice_menu">
+        <field name="name">Manually Invoice Contracts</field>
+        <field name="parent_id" ref="account.menu_finance"/>
+        <field name="groups_id" eval="[(6, 0, [ref('account.group_account_manager')])]"/>
+        <field name="action"
+               ref="contract_manually_create_invoice_act_window"/>
+        <field name="sequence" eval="16"/>
```
sbidoul: Does it land in a reasonably logical place in the menu?

contract/models/contract.py
sbejaoui: it's not triggered with the cron

sbejaoui: the cron call `_recurring_create_invoice`

sbidoul: Could it be triggered by other (possibly custom) code?

sbejaoui: > Could it be triggered by other (possibly custom) code?

no, this meant to be used in the UI

contract/models/contract.py
sbejaoui: this method can be overridden in other modules, so it's simple to get the created invoices in the return

contract/models/contract.py
pedrobaeza: OK, the similar name distracted me.

contract/wizards/contract_manually_create_invoice.xml
pedrobaeza: Uhm, it's a long text for being in main menu. Maybe split it in 2 actions, one for customer and one for suppliers, and put them into their respective submenus?

sbejaoui: done, https://github.com/OCA/contract/pull/414/commits/91f537789d867b6bb0d80cb6ba5a4c31289d42d4

sbejaoui:
@sbidoul , @pedrobaeza can you update your review please

contract/models/contract.py
sbidoul: Since the method is small, I'd add a new method for manual creation, this will proved a cleaner API.

sbidoul (commented):
A minor comment, otherwise LGTM.

baimont (approved):

pedrobaeza:
Can you please squash commits a bit?

sbejaoui:
> Can you please squash commits a bit?

Done

pedrobaeza (approved):

PR #422, opened 2019-11-18
[12.0][FIX] - Run onchange contract_type when changing the contract template

sbejaoui:
Fix this use-case:

If the contract journal is not set on the contract template the contract is created
without a journal when confirming the sale order

### DIFF
--- a/contract/models/contract.py
+++ b/contract/models/contract.py
@@ -209,7 +209,8 @@ class ContractContract(models.Model):
                     field.name in self.NO_SYNC,
                 )
             ):
-                self[field_name] = self.contract_template_id[field_name]
+                if self.contract_template_id[field_name]:
+                    self[field_name] = self.contract_template_id[field_name]
 
     @api.onchange('partner_id')
     def _onchange_partner_id(self):
--- a/product_contract/models/sale_order.py
+++ b/product_contract/models/sale_order.py
@@ -82,6 +82,7 @@ class SaleOrder(models.Model):
                 )
                 contracts |= contract
                 contract._onchange_contract_template_id()
+                contract._onchange_contract_type()
                 order_lines.create_contract_line(contract)
                 order_lines.write({'contract_id': contract.id})
             for line in line_to_update_contract:


product_contract/models/sale_order.py
```diff
@@ -41,11 +41,20 @@ def _compute_is_contract(self):
     @api.multi
     def _prepare_contract_value(self, contract_template):
         self.ensure_one()
+        journal = self.env['account.journal'].search(
+            [
+                ('type', '=', contract_template.contract_type),
+                ('company_id', '=', self.company_id.id),
+            ],
+            limit=1,
+        )
```
sbidoul: Would it make sense to use `contract_template.journal_id`? I'm a bit worried that searching journal types based on contract types would lead to bugs in the future, because these two types are independent in theory.

product_contract/models/sale_order.py
```diff
@@ -41,11 +41,20 @@ def _compute_is_contract(self):
...
+
         return {
             'name': '{template_name}: {sale_name}'.format(
                 template_name=contract_template.name, sale_name=self.name
             ),
             'partner_id': self.partner_id.id,
+            'journal_id': journal.id if journal else False,
```
sbidoul:
```suggestion
            'journal_id': journal.id
```

product_contract/models/sale_order.py
sbejaoui: It's the same logic used to get the default journal when creating a contract or contract template

sbidoul: Ok. But is it not a good opportunity to let users specify the journal on the contract template?

sbejaoui: if the journal is set on the contract template, the result get by this search will be overridden by the onchange method called [here](https://github.com/OCA/contract/pull/422/files/ab177a2999156a5bca9c60b58330abcc30fdc897#diff-7e52e92beab9411dd73aa2b44d0945fdR93)

See also [this](https://github.com/OCA/contract/pull/422/files/ab177a2999156a5bca9c60b58330abcc30fdc897#diff-26c1ee0949150674b4450ea66bd9b8e9R213)

contract/models/contract.py
```diff
@@ -209,7 +209,8 @@ def _onchange_contract_template_id(self):
                     field.name in self.NO_SYNC,
                 )
             ):
-                self[field_name] = self.contract_template_id[field_name]
+                if self.contract_template_id[field_name]:
+                    self[field_name] = self.contract_template_id[field_name]
```
sbidoul, line 213: Maybe we need to play some on change here, e.g. `if field_name == "contract_type": self._onchange_contract_type()`.

sbejaoui:
@sbidoul I pushed a new commit

contract/models/contract.py
```diff
@@ -209,7 +209,9 @@ def _onchange_contract_template_id(self):
...
+        self._onchange_contract_type()
```
pedrobaeza: Aren't onchange trigger in chain if executed from UI?

sbejaoui: `_onchange_contract_template_id` is called [here](https://github.com/OCA/contract/blob/12.0/product_contract/models/sale_order.py#L84)

pedrobaeza: Yeah, but that's something that is done programmatically, not triggered by UI change.

sbejaoui: I change it,

`_onchange_contract_type` need to be called when creating the contract from an SO.

pedrobaeza (approved):

sbejaoui:
@sbidoul , can you update your review here please

sbidoul (approved):

PR #423, opened 2019-11-20
[fix_migration_script_contract_module]-Move all account_analytic_acco…

hieulucky111:
Currently, Migration script doesn't take into account the contract(`contract_id`) on invoice which don't have any  `account.analytic.invoice.line` line.


pedrobaeza:
How can you have invoices without contract lines?

hieulucky111:
> How can you have invoices without contract lines?

because our workflow want to keep the contract information for each invoice. Only invoice is repeated that will have a contract lines.

pedrobaeza:
I still don't understand how can you do that. Is it something custom on top on `contract`?

hieulucky111:
i have checked and we did some custom code, i didn't check carefully.
Sorry about that,  i close this MR.

PR #429, opened 2019-11-26
[12.0][WIP][IMP] - contract: manage a different period range for first invoice

### LINKED ISSUE #428: [CONTRACT] - Pre-paid mode, invoice the first of each month except for the first period
Hi,

We have a use case that the current version of the contract module can't handle. 

For a contract line that starts at 07/11 in a pre-paid mode, the user would like to invoice his customer the first of each month except for the first period that it will be at the 07/11.

To resolve this use case, I propose to add a new field that will be shown to the user only if the last date invoiced is not set (a contract line that was never invoiced)

This field will ask the user about the end date of the first invoicing period and will be automatically set to : `date_start + the invoicing period - 1`.

Let see what this gives with our use-case:

- start date: 07/11
- pre-paid
- montly
- next invoicing date: 07/11
- last date invoiced: None
- first period invoice to date
   - Visible to user
   - set to 06/12

1- If the user don't change the **first period invoice to date** and create the first invoice --> nothing changes comparing to the current version.
2- If the the user set the **first period invoice to date** to 30/11 and create the first invoice:
  -  last date invoiced: 30/11
  - invoice created for the period 07/11 to 30/11
  - next invoicing date: 01/12
  - first period invoice to date
    - not visible to user

What do you think about this change?

cc/ @gva-acsone, @sbidoul , @pedrobaeza

sbejaoui:
closes: #428

Set a different period for the first invoice. Usually done when a contract starts in the middle of the
invoicing period and the user would like to charge his customer the first of each period.

### DIFF
--- a/contract/models/contract_line.py
+++ b/contract/models/contract_line.py
@@ -41,6 +41,14 @@ class ContractLine(models.Model):
     last_date_invoiced = fields.Date(
         string='Last Date Invoiced', readonly=True, copy=False
     )
+    first_period_invoice_to_date = fields.Date(
+        string='First Period invoice to date',
+        help="Set a different period for the first invoice. Usually used when "
+        "a contract starts in the middle of the period and you want to "
+        "charge your customer the first of each period. In this case, "
+        "set this field at the end of the first period.",
+        copy=False,
+    )
     termination_notice_date = fields.Date(
         string='Termination notice date',
         compute="_compute_termination_notice_date",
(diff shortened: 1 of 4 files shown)


sbejaoui:
@gva-acsone, This PR is ready to be tested.

sbidoul:
This use case should now be supported in #434, as pre-paid monthlylastday.

PR #430, opened 2019-11-27
[FIX] contract: Use get_relative_delta by default in migration _init_last_date_invoiced

sergio-teruel:
cc @Tecnativa TT18710
Review @Tardo @carlosdauden

### DIFF
--- a/contract/models/contract_line.py
+++ b/contract/models/contract_line.py
@@ -602,8 +602,9 @@ class ContractLine(models.Model):
     def _init_last_date_invoiced(self):
         """Used to init last_date_invoiced for migration purpose"""
         for rec in self:
-            last_date_invoiced = rec.recurring_next_date - relativedelta(
-                days=1
+            last_date_invoiced = (
+                rec.recurring_next_date - self.get_relative_delta(
+                    rec.recurring_rule_type, rec.recurring_interval)
             )
             if rec.recurring_rule_type == 'monthlylastday':
                 last_date_invoiced = (


contract/models/contract_line.py
```diff
@@ -602,8 +602,9 @@ def _update_recurring_next_date(self):
     def _init_last_date_invoiced(self):
         """Used to init last_date_invoiced for migration purpose"""
         for rec in self:
-            last_date_invoiced = rec.recurring_next_date - relativedelta(
-                days=1
+            last_date_invoiced = (
```
sbejaoui, line 605: What this PR fix?

sbejaoui: The last date invoiced is the last day of the previous period. It can be:
* in the pre-paid mode (the invoice date is the first day of the period) 
   * = the recurring next date - one day
* in the post-paid mode (the invoice date is the first day of the next period) 
   * = the recurring next date - one period (to have first day of the period)
   * \- one day
* in the monthlylast day (the invoice day is the last day of the month)
   * = the last day of the previous month

sergio-teruel: If you have a contract with a line that does invoices every 3 months, what is the last date invoiced??  by default the relative delta substract one day, I think that tthis method would be substract all setting contract period.

sbejaoui: `get_relative_delta` return a relative delta based on the contract line invoicing period, so it don't return a relative delta for one day by default.

If we have a contract with a line that does invoices every 3 months, the last date invoiced depends on recurring_invoice_type:
- if pre-paid --> subtract one day from the invoice date
- if post-paid --> subtract one period (3 months is this case) and one day

for this example  `get_relative_delta`  return relativedelta(months=3)

sbidoul:
@sergio-teruel could you add a first commit with a failing test that illustrate the issue you want to fix?

contract/models/contract_line.py
sbidoul: @sergio-teruel this method has evolved a bit in #434. Can you check if it solves your issue?

sergio-teruel: Hi @sbidoul I think is ok, my issue was when I do a migration form old version. If I have a processed contract with this settings ( Invoice every 3 months and pre-paid ) and the date of last invoice was 10/05/2019, after run migration script (12.0.4.0.0), the last invoice date in contract line is 01/04/2020, the correct date is 10/05/2019... 
The reason was _init_last_date_invoiced method always subtract one day, I think that this method should be substract all period instead...
But with lastest changes i think that the issue is solved.
Selección_513
Selección_512

sergio-teruel: I think that I can close this PR

pedrobaeza:
@sergio-teruel is this needed then?

sergio-teruel:
I think that is not needed... So I close this PR

Issue #426, opened 2019-11-22
[CONTRACT] The recurring rule type monthly-last-day


sbejaoui:
Hi, 

On a contract line there are two fields that define how to compute the next invoice date.
1.  `recurring_rule_type`: the invoicing period: week, month..
2.  `recurring_invoicing_type`: the invoicing policy: pre-paid or post-paid

But there is one `recurring_rule_type` option that cause a problem in this logic. it's the `monthlylastday` which mix in the same notion the invoicing period and policy.

Managing this option increased code complexity by:
* the need to have a separate logic for computing the next invoice date 
* the need to have a separate logic for computing the last invoiced date 
* making the `recurring_invoicing_type` useless when this option is selected

I propose to simplify all this by a small change for the post-paid case:
- The next invoice date is the last day in the period and not the first date of the next period.

example:
* contract line starts 01/11, monthly post-paid
   * `recurring_next_date` --> 30/11
* after invoicing the first period
   * `invoice_date` --> 30/11
   * `last_date_invoiced` --> 30/11
   * `recurring_next_date` --> 31/12

cc: @gva-acsone, @sbidoul , @pedrobaeza

pedrobaeza:
@carlosdauden what do you think about this that you know post-paid cases?

pedrobaeza:
The idea behind `monthlylastday` was to set the invoicing day, but without altering the invoiced period. For example, you start the contract on 14/01. With this option, you invoice on 31/01, but taking the period 14/01 to 13/02. Maybe we should set a third field for setting the invoice date independently from the rest?

sbejaoui:
I can't imagine the invoice date completely independent from the invoicing period. 

When a user select an option from the field `recurring_invoicing_type` (pre-paid or post-paid), he select whether he will invoice his customer at the first day or the last day of the period.

There was a PR that proposed to have field to shift the invoice date from its original value. We can add it to this scoop if we really need to have a different invoice date then :

- The default behavior is:
   - pre-paid option make the invoice date the first day of the period
      - the invoicing period is [the last day invoiced + one day , next invoice date + one period]
   - post-paid option make the invoice date the last date of the period
      - the invoicing period is [the last day invoiced + one day , next invoice date]
- If a shift is set:
   - the invoicing period is the same
   - the invoice is created on the original invoice date
   - the invoice date = invoice date + shift

pedrobaeza:
[...]

PR #431, opened 2019-11-28
[12.0][IMP] - Contract: Add a new field to compute the next period end date

sbejaoui:
An alternative solution for #428 is to make the invoicing period independent from the invoice date. Original idea of @pedrobaeza ([see](https://github.com/OCA/contract/issues/426#issuecomment-558668082))

This PR add a new filed to store the next period end date. With the field `last_date_invoiced` the user can easily see and change the invoicing period.

A migration script was added to initialize the new field.

What we can add to this PR:
-  add an integer field offset to shift the invoice date 
-  make the next invoice date a computed field based on
    -  invoicing_type: pre-paid/post-paid
    -  first and last period dates
    -  offset field
- make the `monthlylastday`  a check box to be used to compute the next invoice date

CC/ @pedrobaeza , @sbidoul , @gva-acsone, @carlosdauden

### DIFF
--- a/contract/models/contract_line.py
+++ b/contract/models/contract_line.py
@@ -41,6 +41,9 @@ class ContractLine(models.Model):
     last_date_invoiced = fields.Date(
         string='Last Date Invoiced', readonly=True, copy=False
     )
+    next_period_end_date = fields.Date(
+        string='Next Period End Date', copy=False
+    )
     termination_notice_date = fields.Date(
         string='Termination notice date',
         compute="_compute_termination_notice_date",
@@ -356,7 +359,7 @@ class ContractLine(models.Model):
                     )
 
     @api.model
-    def _compute_first_recurring_next_date(
+    def _get_recurring_next_date(
         self,
         date_start,
         recurring_invoicing_type,
@@ -374,7 +377,7 @@ class ContractLine(models.Model):
         )
 
     @api.model
-    def compute_first_date_end(
+    def _get_date_end(
         self, date_start, auto_renew_rule_type, auto_renew_interval
     ):
         return (
@@ -382,6 +385,20 @@ class ContractLine(models.Model):
             + self.get_relative_delta(
                 auto_renew_rule_type, auto_renew_interval
             )
+            - relativedelta(days=1)
+        )
+
+    @api.model
+    def _get_next_period_end_date(
+        self, next_period_start_date, recurring_rule_type, recurring_interval
+    ):
+        if recurring_rule_type == 'monthlylastday':
+            return next_period_start_date + self.get_relative_delta(
+                recurring_rule_type, recurring_interval - 1
+            )
+        return (
+            next_period_start_date
+            + self.get_relative_delta(recurring_rule_type, recurring_interval)
             - relativedelta(days=1)
         )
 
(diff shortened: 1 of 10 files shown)


contract/models/contract_line.py
```diff
@@ -580,23 +601,32 @@ def _insert_markers(self, first_date_invoiced, last_date_invoiced):
...
+                if rec.date_end and next_period_end_date > rec.date_end:
+                    next_period_end_date = rec.date_end
+                if rec.date_end and last_date_invoiced >= rec.date_end:
+                    rec.last_date_invoiced = rec.date_end
+                    rec.recurring_next_date = False
+                    rec.next_period_end_date = False
+                else:
+                    rec.last_date_invoiced = last_date_invoiced
+                    rec.recurring_next_date = recurring_next_date
+                    rec.next_period_end_date = next_period_end_date
```
sbidoul, line 629: I think you can continue simplifying by doing `_update_next_period_date` instead of `_update_recurring_next_date` and make `recurring_next_date` a computed field that depends on `next_period_date` and `recurring_invoicing_type`.

contract/models/contract_line.py
```diff
@@ -41,6 +41,9 @@ class ContractLine(models.Model):
     last_date_invoiced = fields.Date(
         string='Last Date Invoiced', readonly=True, copy=False
     )
+    next_period_end_date = fields.Date(
+        string='Next Period End Date', copy=False
+    )
```
sbidoul, line 46: I'd suggest adding a `next_period_start_date` too (computed = last_date_invoiced + 1). That will simplify the code a bit. And you can display to the user instead of `last_date_invoiced`, it will improve the UX.

contract/models/contract_line.py
```diff
@@ -542,23 +577,9 @@ def _get_period_to_invoice(
...
-            else:
-                last_date_invoiced = recurring_next_date - relativedelta(
-                    days=1
-                )
+        last_date_invoiced = self.next_period_end_date
         if stop_at_date_end:
-            if self.date_end and self.date_end < last_date_invoiced:
+            if not self.next_period_end_date and self.date_end:
                 last_date_invoiced = self.date_end
         return first_date_invoiced, last_date_invoiced, recurring_next_date
```
sbidoul, line 584: When you have done `_update_next_period` this `_get_period_to_invoice` can become very dumb: `return next_period_start_date, next_period_end_date`

sbidoul:
So yes :+1: , this PR will be a refactoring of the logic by computing the next period instead of the next invoice date, and making the next invoice date depending on the next period. It will become much easier to reason about the code when this is done, and the UX should be better too by clearly showing what will be the next period to invoice and the invoice date (instead of leaving it to some interpretation of the user as it is now).

And it will pave the way for new features such as:
- end first period at the end of the month in prepaid mode too (similar to monthlylastday which is post-paid now)
- add an offset for the invoice date based on the period to allow, e.g. invoicing a few days before the end of the period
- always invoice at the end of the month even if the period overlaps a month boundary
- etc...

sbidoul:
As discussed, in a first step, let's make `recurring_next_date` computed, but with an inverse method for backward compatibility (the inverse method logging a deprecation warning).

pedrobaeza:
Is this outdated with #434?

sbidoul:
> Is this outdated with #434?

Probably yes. My idea here to make `recurring_next_date` computed from the next period is raising some backward compatibility issue. I still think it would be easier for the user but there are some edge cases that are not reproducible with this approach.

In any case #434 improves maintainability and readability (IMO), and is fully compatible (except two minor differences that you can see in the test changes in #434).

sbidoul:
Closing in favor of #434

PR #434, opened 2019-12-06
[12.0] contract: some refactoring and a UX improvement

### LINKED ISSUE #428: [CONTRACT] - Pre-paid mode, invoice the first of each month except for the first period

sbidoul:
Some simple refactoring of the contract module. The main idea is to split the computation of the next invoice date in two parts: the computation of the next period, followed by the computation of the next invoice date (aka `recurring_next_date`) based on the next period dates.

This is best reviewed one commit at a time to understand individual changes.

There is also a UX improvement: display the next period start and end date on the contract line form, so the user does not need to deduce it from other parameters.

Finally, `monthlylastday` now supports `pre-paid`/`post-paid`. This could be a breaking change
if there is code that creates `monthlylastday` contract lines without setting them to `post-paid` mode.

fixes: #428

### DIFF
--- a/contract/migrations/12.0.5.0.0/pre-migration.py
+++ b/contract/migrations/12.0.5.0.0/pre-migration.py
@@ -1,0 +1,10 @@
+def migrate(cr, version):
+    # pre-paid/post-paid becomes significant for monthlylastday too,
+    # make sure it has the value that was implied for previous versions.
+    cr.execute(
+        """\
+            UPDATE contract_line
+            SET recurring_invoicing_type = 'post-paid'
+            WHERE recurring_rule_type = 'monthlylastday'
+        """
+    )
--- a/contract/models/contract_line.py
+++ b/contract/models/contract_line.py
@@ -664,7 +664,7 @@ class ContractLine(models.Model):
         self.ensure_one()
         first_date_invoiced = False
         if not recurring_next_date:
-            return first_date_invoiced, last_date_invoiced, recurring_next_date
+            return first_date_invoiced, False, recurring_next_date
         first_date_invoiced = (
             last_date_invoiced + relativedelta(days=1)
             if last_date_invoiced
--- a/contract/models/contract_line.py
+++ b/contract/models/contract_line.py
@@ -358,20 +358,51 @@ class ContractLine(models.Model):
     @api.model
     def _get_recurring_next_date(
         self,
-        date_start,
+        next_period_date_start,
         recurring_invoicing_type,
         recurring_rule_type,
         recurring_interval,
     ):
-        if recurring_rule_type == 'monthlylastday':
-            return date_start + self.get_relative_delta(
-                recurring_rule_type, recurring_interval - 1
-            )
-        if recurring_invoicing_type == 'pre-paid':
-            return date_start
-        return date_start + self.get_relative_delta(
-            recurring_rule_type, recurring_interval
+        next_period_date_end = self._get_next_period_date_end(
+            next_period_date_start,
+            recurring_rule_type,
+            recurring_interval,
+            max_date_end=False,  # TODO
         )
+        if recurring_rule_type == 'monthlylastday':
+            return next_period_date_end
+        elif recurring_invoicing_type == 'pre-paid':
+            return next_period_date_start
+        else:  # post-paid
+            return next_period_date_end + relativedelta(days=1)
+
+    @api.model
+    def _get_next_period_date_end(
+        self,
+        next_period_date_start,
+        recurring_rule_type,
+        recurring_interval,
+        max_date_end,
+    ):
+        """Compute the end date for the next period"""
+        if recurring_rule_type == 'monthlylastday':
+            next_period_date_end = (
+                next_period_date_start
+                + self.get_relative_delta(
+                    recurring_rule_type, recurring_interval - 1
+                )
+            )
+        else:
+            next_period_date_end = (
+                next_period_date_start
+                + self.get_relative_delta(
+                    recurring_rule_type, recurring_interval
+                )
+                - relativedelta(days=1)
+            )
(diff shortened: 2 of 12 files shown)


contract/tests/test_contract.py
```diff
@@ -279,7 +279,7 @@ def test_last_invoice_post_paid(self):
         )
         self.contract.recurring_create_invoice()
         self.assertEqual(
-            self.acct_line.recurring_next_date, to_date('2018-04-01')
```
sbejaoui, line 282: The current version assure that a customer is billed the same day each month.

Assume this example:
- date start 01/01
- date end 15/03

the customer will receive an invoices at:
* for pre-paid:  01/01, 01/02 and 01/03
* for monthlylastday:  31/01, 28/02 and 31/03
* for post-paid: 01/02, 01/03 and 01/04

for me it make more sens to say post-paid mode invoice date is the last day of the invoicing period (or the last day of the invoicing period + one day).

Let make a RFC and see what other think about this change

contract/tests/test_contract.py
```diff
@@ -247,7 +247,7 @@ def test_contract_yearly_pre_paid(self):
         self.assertEqual(self.acct_line.last_date_invoiced, last_date_invoiced)
 
     def test_contract_monthly_lastday(self):
-        recurring_next_date = to_date('2018-03-31')
+        recurring_next_date = to_date('2018-02-28')
```
sbejaoui, line 251: +1

contract/migrations/12.0.5.0.0/pre-migration.py
```diff
@@ -0,0 +1,10 @@
+def migrate(cr, version):
+    # pre-paid/post-paid becomes significant for monthlylastday too,
+    # make sure it has the value that was implied for previous versions.
+    cr.execute(
+        """\
+            UPDATE contract_contract
```
sbejaoui:
```suggestion
            UPDATE contract_line
```

sbejaoui (dismissed):

contract/models/contract_line.py
```diff
@@ -664,7 +664,7 @@ def _get_period_to_invoice(
         self.ensure_one()
         first_date_invoiced = False
         if not recurring_next_date:
-            return first_date_invoiced, last_date_invoiced, recurring_next_date
+            return first_date_invoiced, False, recurring_next_date
```
sbidoul: It would be more readable to `return False, False, False`

sbejaoui: yes

sbejaoui (approved):
Fantastic

Thank you @sbidoul for this great work.

For me the most important thing in this PR is the fact that it fix a logic issue we had for a long time. Now the `recurring_invoicing_type` is no longer implicit in the `monthlylastday` option and we can configure a invoicing process to stop at the end of the month and be payed at the first day of the period (even for an interval of several months [see](https://github.com/OCA/contract/blob/eef75b7d88fef7d4385bce9b137b85e7e94381b5/contract/tests/test_contract.py#L973)).

Adding to this fix, this PR open the way for new features (invoicing offset, computed next invoice date,.. ) by simplifying the code and improve the UX.

cc/ @gva-acsone

contract/models/contract_line.py
sbejaoui:
```suggestion
            return False, False, False
```

contract/models/contract_line.py
```diff
@@ -669,7 +669,7 @@ def _get_period_to_invoice(
         self.ensure_one()
         first_date_invoiced = False
```
sbidoul: This line is useless now

sbidoul:
I added the offset field in the contract line form, to further help the user understand what is going on.

@sbejaoui will do some more testing with other depending modules (OCA and customer specific), but we don't expect issue. So this is ready to review. Reviewing individual commits is highly recommended to understand the logical evolution of the code base.

rafaelbn (approved):
After a first quick functional review it looks OK :+1:

pedrobaeza:
It's difficult to predict side effects with the great amount of changes, but let's trust in your good work and extra tests also makes this more confidence. Trying it out in real will tell us any possible problem.

/ocabot merge major

sbidoul:
To summarize breaking changes:

- Contract lines and products using `monthlylastday` MUST now specify `pre-paid` / `post-paid`. There is a migration script for that. The UI has been adapted. Code that creates such lines must be adapted or will default to `pre-paid` whereas `post-paid` was implied before.
- `get_relative_delta()` now works uniformly for all period types (it had an exception for `monthlylastday` before). Code calling it directly must be adapted.
- In `post-paid` mode, the invoice date is now *always* the first day after the end of the period being invoiced. There was an exception before for the last period of the contract.

PR #443, opened 2019-12-17
[FIX] agreement_legal_sale: Ensure Analytic Account is copied from SO to Agreement

dreispt:
If `sale_timesheet` is installed, the `_action_confirm()` may be setting an Analytic Account on the SO.

But since it is not a dependency, that can happen after we create the Agreement.

To work around that, we check if that is the case, and make sure the SO Analytic Account is copied to the Agreement.


max3903 (approved):

dreispt:
Fixed the red build!

ygol (approved):

dreispt:
@max3903 Can you merge please?

PR #444, opened 2019-12-18
[12.0][IMP] - Add a company config option to decide whether to create or to extend contract line at renew action

sbejaoui:
Add a company config option to decide whether to create or to extend contract
line at renew action

### DIFF
--- a/contract/models/contract_line.py
+++ b/contract/models/contract_line.py
@@ -1156,17 +1156,18 @@ class ContractLine(models.Model):
         }
 
     @api.multi
-    def _get_renewal_dates(self):
+    def _get_renewal_new_date_end(self):
         self.ensure_one()
         date_start = self.date_end + relativedelta(days=1)
         date_end = self._get_first_date_end(
             date_start, self.auto_renew_rule_type, self.auto_renew_interval
         )
-        return date_start, date_end
+        return date_end
 
     @api.multi
-    def _renew_create_line(self, date_start, date_end):
+    def _renew_create_line(self, date_end):
         self.ensure_one()
+        date_start = self.date_end + relativedelta(days=1)
         is_auto_renew = self.is_auto_renew
         self.stop(self.date_end, post_message=False)
         new_line = self.plan_successor(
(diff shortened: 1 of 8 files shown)


sbidoul:
@sbejaoui We can probably avoid a breaking change by adding an option on contract l ine that controls how the renewal is done: by adding a new line (current behavior) or by changing the end date.

sbejaoui:
> @sbejaoui We can probably avoid a breaking change by adding an option on contract l ine that controls how the renewal is done: by adding a new line (current behavior) or by changing the end date.

ok, good idea

sbejaoui:
cc @gva-acsone

contract/models/contract_line.py
```diff
@@ -1164,17 +1164,33 @@ def _get_renewal_dates(self):
...
+            company = rec.contract_id.company_id
             date_start, date_end = rec._get_renewal_dates()
-            new_line = rec.plan_successor(
-                date_start, date_end, is_auto_renew, post_message=False
-            )
-            new_line._onchange_date_start()
+            if company.create_new_line_at_contract_line_renew:
+                new_line = rec._renew_create_line(date_start, date_end)
+            else:
+                new_line = rec._renew_extend_line(date_end)
```
sbidoul, line 1195: It bothers me that one method takes `date_start` and the other not. It should be symmetric or bugs will creep in later. Since it does not make sense for `_renew_extend_line` to take `date_start` then it must be removed from `_renew_create_line` (in which case it is always date_end + 1 day), and `_get_renewal_dates` must be changed to `_get_renewal_new_date_end`.

sbejaoui: make sense

contract/models/contract_line.py
```diff
@@ -1156,25 +1156,43 @@ def action_stop_plan_successor(self):
...
-            is_auto_renew = rec.is_auto_renew
-            rec.stop(rec.date_end, post_message=False)
-            date_start, date_end = rec._get_renewal_dates()
-            new_line = rec.plan_successor(
-                date_start, date_end, is_auto_renew, post_message=False
-            )
-            new_line._onchange_date_start()
+            company = rec.contract_id.company_id
+            date_end = rec._get_renewal_new_date_end()
+            date_start = rec.date_end + relativedelta(days=1)
```
sbidoul, line 1191:
```suggestion
```

sbidoul (approved):
LGTM

contract/models/contract_line.py
sbejaoui: it's used in the contract posted message

https://github.com/OCA/contract/blob/c15a876603df32c82bd051df95015c04c2939f1e/contract/models/contract_line.py#L1197-L1209

sbidoul: Ah, yes, sorry.

gva-acsone (approved):
LGTM Functional test

PR #445, opened 2019-12-19
[12.0][FIX][CONTRACT] - Recompute invoice date after contract line stop

sbejaoui:
When a contract line is stopped and its date_end is changed, the system must update the invoicing_date

### DIFF
--- a/contract/models/contract_line.py
+++ b/contract/models/contract_line.py
@@ -540,7 +540,7 @@ class ContractLine(models.Model):
     def _onchange_date_start(self):
         for rec in self.filtered('date_start'):
             rec.recurring_next_date = self.get_next_invoice_date(
-                rec.date_start,
+                rec.next_period_date_start,
                 rec.recurring_invoicing_type,
                 rec.recurring_invoicing_offset,
                 rec.recurring_rule_type,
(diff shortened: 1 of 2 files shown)


sbejaoui:
@sbidoul , @gva-acsone, can you review this please

gva-acsone (approved):
LGTM Functional test

contract/models/contract_line.py
```diff
@@ -818,6 +818,17 @@ def stop(self, date_end, manual_renew_needed=False, post_message=True):
                     }
                     if rec.last_date_invoiced == date_end:
```
sbidoul: Do we have a test for this case (`rec.last_date_invoiced == date_end`)? Is it still a special case (at first glance I'd say it should be covered by `get_next_invoice_date`)?

sbejaoui: yes, 

https://github.com/OCA/contract/blob/659b91b0f0c50d29cf0b685fe9ab6bd1fb840543/contract/tests/test_contract.py#L2278-L2282

sbidoul: @sbejaoui I still think the `== date_end` is now useless as this cased should be covered by the `get_next_invoice_date` in the else.

sbejaoui: I will test it

sbejaoui: True, get_next_invoice_date cover the case, I will make the change

sbidoul (approved):
LGTM (code review)

PR #454, opened 2020-01-07
[12.0][FIX] - forecast period for auto renew contract lines

sbejaoui:
The contract forecast is a simulation of the upcoming invoices that we can expect from a contract based on different parameters like the start date, end date, ..

A special case is for contract line with an undefined end date or set to auto-renew, we can't decide when to stop the forecast.

A company paramétré was added to define the forecast end date for those cases.  

If a contract line is set to auto-renew the forecast should continue  after the date end and stop at the company forecast period and the same for contract lines with an undefined end date

### DIFF
--- a/contract_forecast/models/contract_line.py
+++ b/contract_forecast/models/contract_line.py
@@ -74,6 +74,7 @@ class ContractLine(models.Model):
                     if rec.last_date_invoiced
                     else rec.date_start - relativedelta(days=1)
                 )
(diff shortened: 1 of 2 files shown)


sbejaoui:
cc/ @sbidoul @gva-acsone

sbidoul (approved):

gva-acsone (approved):
Functional test +1

PR #455, opened 2020-01-08
[12.0][FIX+IMP] contract: Handle properly multi-company setups

pedrobaeza:
If you have contracts in several companies, cron will create all of them, but property fields will be populated with incorrect data as the taken company is the main from the cron user (usually admin).

@Tecnativa TT21285

### DIFF
--- a/contract/models/contract.py
+++ b/contract/models/contract.py
@@ -281,7 +281,9 @@ class ContractContract(models.Model):
         invoice_type = 'out_invoice'
         if self.contract_type == 'purchase':
             invoice_type = 'in_invoice'
-        vinvoice = self.env['account.invoice'].new({
+        vinvoice = self.env['account.invoice'].with_context(
+            force_company=self.company_id.id,
+        ).new({
             'partner_id': self.invoice_partner_id.id,
             'type': invoice_type,
         })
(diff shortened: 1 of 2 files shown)


pedrobaeza:
@sbidoul @sbejaoui not sure if you override `_finalize_invoice_values` in your custom modules. If that's the case, do you see how to do it other way with the already provided hooks or is it a problem?

sbidoul:
Hi @pedrobaeza,

I quickly checked and we do have overrides of `_finalize_invoice_values`. I see the redundancy but I'd like to wait until @sbejaoui is back from holidays to check all the implications.

I'm fine with the multi company part, so if you want to move quickly with that part, you can do a separate PR for it.

pedrobaeza:
It depends on when he comes back. If it's next week, we can wait.

sbidoul:
End of next week. Travis is red, though.

pedrobaeza:
Yeah, I'll check

pedrobaeza:
OK, let's go first only for the multi-company thing, although this means to add an extra context also on finalize method, but the other involves other modules. I'm doing the PR after this.

carlosdauden (approved):

sbidoul (approved):
(code review)

pedrobaeza:
@sbidoul should I proceed with the merge then?

PR #456, opened 2020-01-08
[12.0][IMP] contract: invoice generation redundant method

pedrobaeza:
I have detected a method that was created as redundant and with the same technique used when preparing the line values, so better to have everything together in the same method instead of having it spread.

Contains #455 and must be rebased when the other is merged.

@Tecnativa TT21285

### DIFF
--- a/contract/models/contract.py
+++ b/contract/models/contract.py
@@ -330,56 +330,32 @@ class ContractContract(models.Model):
 
     @api.model
     def _finalize_invoice_values(self, invoice_values):
-        """
-        This method adds the missing values in the invoice lines dictionaries.
-
-        If no account on the product, the invoice lines account is
-        taken from the invoice's journal in _onchange_product_id
-        This code is not in finalize_creation_from_contract because it's
-        not possible to create an invoice line with no account
-
-        :param invoice_values: dictionary (invoice values)
-        :return: updated dictionary (invoice values)
-        """
-        # If no account on the product, the invoice lines account is
-        # taken from the invoice's journal in _onchange_product_id
-        # This code is not in finalize_creation_from_contract because it's
-        # not possible to create an invoice line with no account
-        new_invoice = self.env['account.invoice'].with_context(
-            force_company=invoice_values['company_id'],
-        ).new(invoice_values)
-        for invoice_line in new_invoice.invoice_line_ids:
-            name = invoice_line.name
-            account_analytic_id = invoice_line.account_analytic_id
-            price_unit = invoice_line.price_unit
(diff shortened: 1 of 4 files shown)


pedrobaeza:
Rebased and waiting for feedback from @sbejaoui when he comes back from holidays

contract/models/contract.py
```diff
@@ -379,12 +344,7 @@ def _finalize_and_create_invoices(self, invoices_values):
         """
         if isinstance(invoices_values, dict):
```
sbejaoui: I believe this is useless now.

sbejaoui (approved):
LGTM, 

I managed to rewrite the override we made on  `_finalize_invoice_values` and got rid of it.

Thank you @pedrobaeza

contract/models/contract.py
pedrobaeza: Removed and improved docstring in involved methods

pedrobaeza: And welcome back from holidays!

pedrobaeza:
@sbidoul can you please proceed to merge this?

sbidoul:
is there no way to keep backward compatibility? The problem I see is if people have overrides of, e.g. `_finalize_invoice_values`, they will not immediately notice it because it will be silently ignored.

pedrobaeza:
@sbidoul I have kept the method for compatibility.

contract/models/contract.py
```diff
@@ -327,59 +327,34 @@ def action_contract_send(self):
...
-            invoice_line.update(
-                {
-                    'name': name,
-                    'account_analytic_id': account_analytic_id,
-                    'price_unit': price_unit,
-                }
-            )
-        return new_invoice._convert_to_write(new_invoice._cache)
+        """Provided for keeping compatibility in this version."""
+        # TODO: Must be removed in >=13.0
```
sbidoul, line 334: nit: log a deprecation warning?

sbidoul (commented):
> @sbidoul I have kept the method for compatibility.

Thanks! Can you check why travis is red?

pedrobaeza:
It was red due to flake8 thing. It should be good now.

contract/models/contract.py
pedrobaeza: I prefer not as we are going to keep it in this version, so deprecation doesn't apply at all, and when migrating then the TODO is clear. Populating the logger needs also several lines that doesn't worth IMO.

pedrobaeza:
@sbidoul @sbejaoui  can we merge this then?

sbidoul:
yes, thanks!
/ocabot merge patch

PR #457, opened 2020-01-27
[ADD] agreement_legal_sale_fieldservice

osi-scampbell:
The following PR adds the module agreement_legal_sale_fieldservice. This module ensures that when an agreement is created from a sales order, the fsm_location of the agreement is set with the fsm_location of the partner on the fsm_order


agreement_legal_sale_fieldservice/models/sale_order.py
```diff
@@ -0,0 +1,18 @@
...
+    _inherit = 'sale.order'
+
+    @api.multi
+    def _action_confirm(self):
+        res = super(SaleOrder, self)._action_confirm()
+        for order in self:
+            agreement_id = self.env['agreement'].\
+                search([('code', '=', order.name)])
+            agreement_id.fsm_location_id = order.partner_id.\
+                service_location_id
```
max3903: @osi-scampbell What about using the FSM location of the order?

max3903:
@osi-scampbell Can you remove the `i18n` directory from the module you copied?

agreement_legal_sale_fieldservice/models/sale_order.py
osi-scampbell: @max3903 This module only depends on agreement_legal_sale and fieldservice_agreement, I can implement that workflow if I add fieldservice_sale to the dependencies which I think makes sense.

agreement_legal_sale_fieldservice/__manifest__.py
```diff
@@ -0,0 +1,25 @@
...
+        'fieldservice_sale'
+    ],
+    'data': [
+    ],
+    'installable': True,
+    'development_status': 'Beta',
+    'maintainers': [
+        'osi-scampbell',
+        'max3903',
+    ],
```
dreispt, line 24: I suggest making this an autoinstallable  "glue" module. For that add the key `"auto_install": True`

agreement_legal_sale_fieldservice/models/sale_order.py
```diff
@@ -0,0 +1,18 @@
...
+
+class SaleOrder(models.Model):
+    _inherit = 'sale.order'
+
+    @api.multi
+    def _action_confirm(self):
+        res = super(SaleOrder, self)._action_confirm()
+        for order in self:
+            agreement_id = self.env['agreement'].\
+                search([('code', '=', order.name)])
```
dreispt: Not sure, but isn't there a sale_id field in the Agreements model, that can be used for this?

agreement_legal_sale_fieldservice/models/sale_order.py
dreispt: What if we have a delivery address different from the customer main address? I so, should we use it instead?

osi-scampbell: @dreispt The client requested the fsm_location so I think we should stick with that for now, if someone raises an issue regarding this then I would be willing to look into solutions for that workflow

dreispt (approved):
LGTM

dreispt:
Travis error was on another module. Restarted the build to see if that has been solved.

agreement_legal_sale_fieldservice/models/sale_order.py
max3903: @osi-scampbell  We have the Service Location on the sale order. Why not using it?

agreement_legal_sale_fieldservice/models/sale_order.py
```diff
@@ -0,0 +1,18 @@
+# Copyright (C) 2019 - TODAY, Open Source Integrators
```
max3903:
```suggestion
# Copyright (C) 2019 Open Source Integrators
```

agreement_legal_sale_fieldservice/__manifest__.py
```diff
@@ -0,0 +1,26 @@
+# Copyright (C) 2019 - TODAY, Open Source Integrators
```
max3903:
```suggestion
# Copyright (C) 2019 Open Source Integrators
```

max3903:
@osi-scampbell Please use the icon from #470

agreement_legal_sale_fieldservice/models/sale_order.py
dreispt: Tests on this line result in a singleton error: https://travis-ci.org/OCA/contract/jobs/652175553#L2030

dreispt:
Max's concern's were fixed. I'm merging.

PR #465, opened 2020-02-11
[12.0][FIX] product_contract: set 'Contract template' field company depend

### LINKED ISSUE #464: [12.0][contract][product_contract] Is contract module multi-company?
Hello,

I test the module in multi-company enviroment and it doesn't look like multi-company is supported when you add `product_contract`

- In Odoo products are multi-company, this means that you can use 1 product for n companies
- Contract templates are not multi-company, this means 1 "contract template" per company.

In a real environment with 30 products for contract, when you stablish a `contract template` then **the product becomes single-company** :cry:

ernestotejeda:
Fixes: #464
Cc @Tecnativa TT21084

### DIFF
--- a/product_contract/models/product_template.py
+++ b/product_contract/models/product_template.py
@@ -10,8 +10,10 @@ class ProductTemplate(models.Model):
     _inherit = 'product.template'
 
     is_contract = fields.Boolean('Is a contract')
-    contract_template_id = fields.Many2one(
-        comodel_name='contract.template', string='Contract Template'
+    property_contract_template_id = fields.Many2one(
+        comodel_name='contract.template',
+        string='Contract Template',
+        company_dependent=True,
     )
     default_qty = fields.Integer(string="Default Quantity", default=1)
     recurring_rule_type = fields.Selection(
@@ -42,13 +44,13 @@ class ProductTemplate(models.Model):
         string='Termination Notice type',
     )
 
-    @api.onchange('is_contract')
-    def _change_is_contract(self):
-        """ Clear the relation to contract_template_id when downgrading
-        product from contract
-        """
-        if not self.is_contract:
-            self.contract_template_id = False
+    @api.multi
+    def write(self, vals):
+        if 'is_contract' in vals and vals['is_contract'] is False:
+            for company in self.env['res.company'].search([]):
+                self.with_context(force_company=company.id).write(
+                    {'property_contract_template_id': False})
+        super().write(vals)
 
     @api.constrains('is_contract', 'type')
     def _check_contract_product_type(self):
--- a/product_contract/models/sale_order.py
+++ b/product_contract/models/sale_order.py
@@ -2,7 +2,8 @@
 # Copyright 2018 ACSONE SA/NV.
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-from odoo import fields, api, models
+from odoo import _, fields, api, models
+from odoo.exceptions import ValidationError
 
 
 class SaleOrder(models.Model):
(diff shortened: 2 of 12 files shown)


rafaelbn (approved):
Functional reviewed in local :+1: 

Thanks @ernestotejeda @pedrobaeza for the efforts in doing this possible. You know I don't like this solution as contract product_contract IMHO has a bad design for a easy a general approach with multi-companies. But this could solve the problem in customers.

pedrobaeza:
Please add the migration script and finish the PR now that https://github.com/OCA/openupgradelib/pull/197 is merged.

product_contract/migrations/12.0.4.0.0/post-migration.py
```diff
@@ -0,0 +1,12 @@
+# Copyright 2019 Tecnativa - Ernesto Tejeda
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.convert_to_company_dependent(
```
pedrobaeza: It must be in pre-migration, and you need to add the field with `add_fields` method (or at least to add the column).

product_contract/models/sale_order_line.py
```diff
@@ -19,7 +19,7 @@ class SaleOrderLine(models.Model):
     contract_template_id = fields.Many2one(
         comodel_name='contract.template',
         string='Contract Template',
-        related='product_id.product_tmpl_id.contract_template_id',
+        related='product_id.product_tmpl_id.property_contract_template_id',
```
pedrobaeza: This field is not valid anymore, as the value depends on the company. You can do:

1. A computed field that takes company_id of the sale.order and get the property linked (with context `force_company=company_id`.
2. Do it a regular field, and populate it with the onchange of the company or the product.
3. Remove it and change any reference to it in the code by the proper related field.

My favorite one is 3 if possible, then 1.

product_contract/models/sale_order.py
```diff
@@ -71,11 +71,11 @@ def action_create_contract(self):
                 )
             )
             for contract_template in line_to_create_contract.mapped(
-                'product_id.contract_template_id'
+                'product_id.property_contract_template_id'
             ):
                 order_lines = line_to_create_contract.filtered(
```
pedrobaeza, line 86: It would be safer to add `with_context(force_company=line_to_create_contract.company_id)` for being sure the contract template is got from the right company. What awful long variable names by the way...

product_contract/models/product_template.py
```diff
@@ -48,7 +50,7 @@ def _change_is_contract(self):
         product from contract
         """
         if not self.is_contract:
-            self.contract_template_id = False
+            self.property_contract_template_id = False
```
pedrobaeza: This is not valid in a multi-company context. Instead, you must overwrite `write` method and if `is_contract` is False, remove the value for all the companies.

product_contract/tests/test_product.py
```diff
@@ -19,10 +19,11 @@ def test_change_is_contract(self):
         """ It should verify that the contract_template_id is removed
         when is_contract is False """
         self.service_product.is_contract = True
-        self.service_product.contract_template_id = self.contract.id
+        self.service_product.property_contract_template_id = self.contract.id
         self.service_product.is_contract = False
         self.service_product.product_tmpl_id._change_is_contract()
```
pedrobaeza: Change this test according my comment on the onchange.

product_contract/views/product_template.xml
```diff
@@ -24,7 +24,7 @@
                       name="contract"
                       attrs="{'invisible': [('is_contract', '=', False)],}">
                     <group>
-                        <field name="contract_template_id"
+                        <field name="property_contract_template_id"
                                attrs="{'required':[('is_contract', '=', True)]}"/>
```
pedrobaeza: I don't think this should be made required in multi-company context. Instead, raise an exception while creating the contract on sales order validation if not found.

pedrobaeza (changes requested):

ernestotejeda:
Changes done

@pedrobaeza , Related to calling the `add_fields` method:
I saw that `add_fields` method adds a column in the table associated to the specified model, but does not take into account that in case of property field the column shouldn't be added in the table. Maybe could be a good idea to incorporate that functionality in the `add_fields` method, but now, what I did was add here in the migration script the insertion of that field in the ir_model_field table before calling `convert_to_company_dependent` method

pedrobaeza (approved):

pedrobaeza:
> Maybe could be a good idea to incorporate that functionality in the add_fields method...

Yeah, you're right. It's a missing feature.

pedrobaeza:
@sbejaoui is it ok for you?

@ernestotejeda add you as contributor of the module and bump manually the manifest version for avoiding any problem.

ernestotejeda:
Changes done.
@pedrobaeza , I also added you as a contributor because you were who made the design and you were who guided me

sbejaoui:
Thank you  @ernestotejeda  for this PR

> @sbejaoui is it ok for you?

I need sometime to test it in our multi-company setup

pedrobaeza:
@sbejaoui please try to check this quickly, as this keeps getting in conflict due to the hardcoded manifest version for migration scripts.

sbejaoui (approved):

pedrobaeza:
Thanks, merging:

/ocabot merge

Issue #463, opened 2020-02-10
Logo SVG (Legal Agreements)


marcelsavegnago:
Hi, 

Does anyone have the logo (SVG) of the Legal Agreements module?

marcelsavegnago:
The current icon is smaller than the other icons when using the web_responsive module. If nobody has the SVG file, I propose a new logo. What do you think ?

![icon](https://user-images.githubusercontent.com/3595132/74282011-330b3c80-4cfe-11ea-8f65-55a3f09b83c2.png)


[icon.zip](https://github.com/OCA/contract/files/4188562/icon.zip)

marcelsavegnago:
@ygol Do you have the agreement_legal module logo vector?

ygol:
I don't have the vector. The current icon comes from the "agreement" module in v11.

marcelsavegnago:
> I don't have the vector. The current icon comes from the "agreement" module in v11.

I saw it now ... but it doesn't have the vector either.

marcelsavegnago:
@ygol Do you believe that it is worth opening a PR to replace the Icon? I understand that it is only visual but it makes a small difference ?

![image](https://user-images.githubusercontent.com/3595132/74285846-69988580-4d05-11ea-8a0d-b4d64f9d69d0.png)

![image](https://user-images.githubusercontent.com/3595132/74285891-7917ce80-4d05-11ea-9f75-0e6dcf0e3880.png)

ygol:
On runbot it is not showing smaller:
http://3407763-12-0-2eb377.runbot2.odoo-community.org/web#action=32&model=ir.module.module&view_type=kanban&menu_id=5
![image](https://user-images.githubusercontent.com/2945174/74316953-27ae2480-4d83-11ea-9e86-582efbb64d07.png)

marcelsavegnago:
> On runbot it is not showing smaller: http://3407763-12-0-2eb377.runbot2.odoo-community.org/web#action=32&model=ir.module.module&view_type=kanban&menu_id=5 ![image](https://user-images.githubusercontent.com/2945174/74316953-27ae2480-4d83-11ea-9e86-582efbb64d07.png)

In this case you are looking at the list of applications. Install the Web Responsive module and see that in this case the icon is impacted due to the image resolution. See the print screen below.

![image](https://user-images.githubusercontent.com/3595132/74365230-660c0980-4dac-11ea-9672-bfcb6e40e351.png)

ygol:
[...]

PR #470, opened 2020-02-15
[12.0][IMP] New icon for agreement modules (svg file in agreement_legal)

marcelsavegnago:
I made a new icon for the agreement modules since I didn't get the SVG file for the current icon. In the main menu, the current icon is smaller than the other icons when the web_responsive module is activated. In addition, the proposed new icon already makes it close to the standard of current Odoo icons.

image

https://github.com/OCA/contract/issues/463


max3903 (approved):

pedrobaeza:
I'm missing the XML for loading that icon in the main menu.

It would be also interesting to add little variations of the icon according the module for differentiating them on the module list.

patrickrwilson (approved):
Looks great! LGTM

max3903:
@pedrobaeza It's already there:
https://github.com/OCA/contract/blob/12.0/agreement_legal/views/menu.xml#L11

For the variations, please open an issue.

PR #473, opened 2020-02-17
[12.0][FIX] agreement_legal: Change company_contact_id domain

Tardo:
Before this commit, the view of agreement uses a dotted domain on the 'right leaf' in the company_conact_id field causing an javascript error.

cc @tecnativa TT22261

### DIFF
--- a/agreement_legal/models/agreement.py
+++ b/agreement_legal/models/agreement.py
@@ -129,6 +129,7 @@ class Agreement(models.Model):
     use_parties_content = fields.Boolean(
         string="Use parties content",
         help="Use custom content for parties")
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):

max3903 (approved):

ygol (commented):
If not mistaking, it was discussed to get rid of the field and replace it by subscribing as followers.
ref. https://github.com/OCA/contract/pull/344

pedrobaeza:
Well, meanwhile you decide about that, we need to fix the bug, so let's go with this and then #344 can continue on its own:

/ocabot merge minor

PR #479, opened 2020-02-25
[IMP] - Improve UX, open contract form if the sale order is related to one (same for contract)

sbejaoui:


sbejaoui:
CC/ @gva-acsone

flotho (approved):
LGTM

carlosdauden (approved):

PR #468, opened 2020-02-14
Added correct domain for supplier contract's on payment mode

vinclar:
Added view for supplier contracts, respecting the domain for payment_type. 
Before was showing the inbound types in the supplier payment

PR #483, opened 2020-03-09
Added correct domain for supplier contract's on payment mode (squashed #468)

dreispt:
Added view for supplier contracts, respecting the domain for payment_type.
Before was showing the inbound types in the supplier payment

Supersedes #468


pedrobaeza (approved):
OK, this one can serve

PR #486, opened 2020-03-17
[12.0][IMP] contract: add section and notes in contract line

ernestotejeda:
Allows users to add sections and notes in contract lines
Screenshot from 2020-03-17 18-17-41

Cc @Tecnativa TT22935

### DIFF
--- a/contract/models/contract_line.py
+++ b/contract/models/contract_line.py
@@ -146,6 +146,9 @@ class ContractLine(models.Model):
     def _compute_state(self):
         today = fields.Date.context_today(self)
         for rec in self:
+            if rec.display_type:
+                rec.state = False
+                continue
             if rec.is_canceled:
                 rec.state = 'canceled'
                 continue
@@ -648,15 +651,15 @@ class ContractLine(models.Model):
 
     @api.depends('recurring_next_date', 'date_start', 'date_end')
     def _compute_create_invoice_visibility(self):
+        # TODO: depending on the lines, and their order, some sections
+        # have no meaning in certain invoices
         today = fields.Date.context_today(self)
         for rec in self:
-            if rec.date_start:
-                if today < rec.date_start:
-                    rec.create_invoice_visibility = False
-                else:
-                    rec.create_invoice_visibility = bool(
-                        rec.recurring_next_date
-                    )
+            if (not rec.display_type and
+                    rec.date_start and today >= rec.date_start):
+                rec.create_invoice_visibility = bool(rec.recurring_next_date)
+            else:
+                rec.create_invoice_visibility = False
 
     @api.multi
     def _prepare_invoice_line(self, invoice_id=False, invoice_values=False):
@@ -665,6 +668,7 @@ class ContractLine(models.Model):
             self.last_date_invoiced, self.recurring_next_date
         )
         invoice_line_vals = {
+            'display_type': self.display_type,
             'product_id': self.product_id.id,
             'quantity': self._get_quantity_to_invoice(*dates),
             'uom_id': self.uom_id.id,
@@ -1299,10 +1303,11 @@ class ContractLine(models.Model):
     @api.multi
     def unlink(self):
         """stop unlink uncnacled lines"""
-        if not all(self.mapped('is_canceled')):
-            raise ValidationError(
-                _("Contract line must be canceled before delete")
-            )
+        for record in self:
+            if not (record.is_canceled or record.display_type):
+                raise ValidationError(
+                    _("Contract line must be canceled before delete")
+                )
         return super().unlink()
 
     @api.multi
--- a/contract/models/contract_line.py
+++ b/contract/models/contract_line.py
@@ -251,6 +251,8 @@ class ContractLine(models.Model):
             ]
         if state == 'canceled':
             return [('is_canceled', '=', True)]
+        if not state:
+            return [('display_type', '!=', False)]
 
     @api.model
     def _search_state(self, operator, value):
@@ -261,11 +263,8 @@ class ContractLine(models.Model):
             'upcoming-close',
             'closed',
             'canceled',
+            False,
         ]
-        if operator == '!=' and not value:
-            return []
-        if operator == '=' and not value:
-            return [('id', '=', False)]
         if operator == '=':
             return self._get_state_domain(value)
         if operator == '!=':
@@ -278,8 +277,6 @@ class ContractLine(models.Model):
             return domain
         if operator == 'in':
             domain = []
-            if not value:
-                return [('id', '=', False)]
             for state in value:
                 if domain:
                     domain.insert(0, '|')
@@ -287,6 +284,8 @@ class ContractLine(models.Model):
             return domain
 
         if operator == 'not in':
+            if set(value) == set(states):
+                return [('id', '=', False)]
             return self._search_state(
                 'in', [state for state in states if state not in value]
             )
--- a/contract/tests/test_contract.py
+++ b/contract/tests/test_contract.py
@@ -42,9 +42,16 @@ class TestContractBase(common.SavepointCase):
             'recurring_rule_type': 'yearly',
             'recurring_interval': 1,
         }
+        cls.section_template_vals = {
+            'display_type': 'line_section',
+            'name': 'Test section',
+        }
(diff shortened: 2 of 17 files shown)


contract/models/contract_line.py
```diff
@@ -146,6 +146,10 @@ def _compute_termination_notice_date(self):
     def _compute_state(self):
         today = fields.Date.context_today(self)
         for rec in self:
+            if rec.display_type:
+                rec.state = False
```
pedrobaeza, line 150: Is it correct to put `False` as `state`

contract/models/contract_line.py
```diff
@@ -146,6 +146,10 @@ def _compute_termination_notice_date(self):
...
+                continue
+
```
pedrobaeza: Don't add empty extra lines

contract/models/contract_line.py
```diff
@@ -1279,10 +1281,6 @@ def fields_view_get(
         self, view_id=None, view_type='form', toolbar=False, submenu=False
     ):
         default_contract_type = self.env.context.get('default_contract_type')
-        if view_type == 'tree' and default_contract_type == 'purchase':
```
pedrobaeza: Why removing this?

contract/models/contract_line.py
```diff
@@ -1299,7 +1297,7 @@ def fields_view_get(
     @api.multi
     def unlink(self):
         """stop unlink uncnacled lines"""
-        if not all(self.mapped('is_canceled')):
+        if not all(self.mapped(lambda r: r.is_canceled or r.display_type)):
```
pedrobaeza: This is more performant with a `for` loop stopping at the first occurrence instead of forcing to process whole recordset.

contract/tests/test_contract.py
```diff
@@ -2280,16 +2280,9 @@ def test_action_stop_plan_successor(self):
         )
 
     def test_purchase_fields_view_get(self):
-        purchase_tree_view = self.env.ref(
```
pedrobaeza: I think this has been removed due to an old 12.0 base branch

contract/views/contract.xml
```diff
@@ -96,8 +96,75 @@
...
+                                   widget="section_and_note_one2many"
+                                   context="{'default_contract_type': contract_type}">
+                                <tree decoration-muted="is_canceled"
+                                      decoration-info="create_invoice_visibility and not is_canceled">
+                                    <control>
+                                        <create string="Add a line"/>
+                                        <create string="Add a section" context="{'default_display_type': 'line_section'}"/>
+                                        <create string="Add a note" context="{'default_display_type': 'line_note'}"/>
+                                    </control>
+                                    <field name="display_type" invisible="1"/>
```
pedrobaeza, line 110: Why do you need to declare inline tree view? The idea is to keep the isolated one.

contract/views/contract_line.xml
```diff
@@ -77,97 +79,4 @@
...
-                <button name="renew" string="Renew"
-                        type="object"
-                        icon="fa-fast-forward text-success"
-                        groups="base.group_no_one"
-                        attrs="{'invisible': [('is_auto_renew', '=', False)]}"/>
-            </tree>
-        </field>
-    </record>
-
-    <!--Supplier TREE view-->
```
pedrobaeza: Again the same lost code due to old base branch

pedrobaeza (changes requested):

contract/models/contract_line.py
ernestotejeda: what state should I set to lines that correspond to the sections and notes? they aren't common contract lines

contract/views/contract.xml
ernestotejeda: These <control> tags can only be defined in embedded tree views. If I defined that in an isolated tree view we get an error

ernestotejeda:
Last changes:
- I fixed the contract report because it looked bad and I also adapted that report so the lines corresponding to sections and notes are correctly printed
- I restored the deleted views to avoid possible errors in other modules (for example because of adding fields to those views).

ping @pedrobaeza

contract/models/contract_line.py
```diff
@@ -650,13 +653,11 @@ def _check_start_end_dates(self):
...
-                    rec.create_invoice_visibility = False
-                else:
-                    rec.create_invoice_visibility = bool(
-                        rec.recurring_next_date
-                    )
+            if (not rec.display_type and
+                    rec.date_start and today >= rec.date_start):
+                rec.create_invoice_visibility = bool(rec.recurring_next_date)
+            else:
+                rec.create_invoice_visibility = False
```
pedrobaeza, line 661: This has risen me a great question: how do you control which sections/notes are invoiced each time? AFAIK, you are adding them each time you invoice, but depending on the lines, and their order, some sections have no meaning in certain invoices. Please put a big TODO about this and add a point in the ROADMAP about this. We are not going to cover it because we are overpassed in the budget and it's not our customer use-case, but we have to note it.

pedrobaeza (approved):
But add the ROADMAP as stated

pedrobaeza:
@sbejaoui this improvement adds a valuable feature to the module. What do you think about it?

sbejaoui:
> @sbejaoui this improvement adds a valuable feature to the module. What do you think about it?

Thank you for this interesting feature. It's a real improvement. 

It will be nice to add this to `product_contract` module to get note and section lines from the SO to the contract.

https://github.com/OCA/contract/blob/14e46b14006ba9391c9f7ccff5d0aa9794eb68e0/product_contract/models/sale_order_line.py#L143

There is the contract_forecast module that simulate contract income that need to consider this change too.

https://github.com/OCA/contract/blob/14e46b14006ba9391c9f7ccff5d0aa9794eb68e0/contract_forecast/models/contract_line.py#L53

But, can we keep this open for a sometime so we can think on how to resolve this https://github.com/OCA/contract/pull/486#discussion_r397773004.

pedrobaeza:
Well, we are not going to work on that and it's not a problem at all for normal usage with a contract with same intervals for all their lines. The same if you don't uses sections/notes but have different intervals, so this shouldn't be blocking IMO.

pedrobaeza:
And the same for the rest of the modules: they don't provide the feature but they are also compatible, so no problem.

sbejaoui:
> And the same for the rest of the modules: they don't provide the feature but they are also compatible, so no problem.

I am not sure about contract_forecast need to test it. but it's ok for me

pedrobaeza:
OK, maybe you need to exclude explicitly only at count effects, but for amounts it shouldn't be a problem as notes/sections account as 0.

contract/models/contract_line.py
ernestotejeda: True, I'm going to add that to the roadmap.
And now that you mention that, I think this change does not make sense since the sections and notes alone don't have to influence whether or not the invoices are generated. Am I right?

pedrobaeza: Well, you can keep it this way any way for now while the ROADMAP is not solved.

pedrobaeza: Or maybe the contrary (`True`), if you base what to pass to invoice on this field.

pedrobaeza:
@sbejaoui any news about this?

sbejaoui (approved):
> @sbejaoui any news about this?

I made a full test and it works out of the box for `contract_forecast`. Thank you.

Can you take the time to see what it can be done to make the coverage green again.

pedrobaeza:
I have improved the test coverage adding a section in tests and testing `state` and `create_invoice_visibility` computed fields, and it has served for discovering a bug in `_search_state` method, so let's go with this! Thanks for the review

/ocabot merge major

contract/models/contract_line.py
```diff
@@ -278,15 +277,15 @@ def _search_state(self, operator, value):
...
-            if not value:
-                return [('id', '=', False)]
             for state in value:
                 if domain:
                     domain.insert(0, '|')
                 domain.extend(self._get_state_domain(state))
             return domain
 
         if operator == 'not in':
+            if set(value) == set(states):
```
pedrobaeza, line 287: This is the bug I have discovered FYI @sbejaoui

PR #488, opened 2020-03-18
[12.0][ADD] - Add new module Product Contract Transmit Method

sbejaoui:


product_contract_transmit_method/README.rst
```diff
@@ -0,0 +1,73 @@
+================================
+Product Contract Transmit Method
```
bodi000: I think the name is misleading, as it is a sale order to contract link, not dealing with product. The `product_contract` family deals precisely with products linked to contracts. I suggest `sale_contract_transmit_method` or `contract_sale_transmit_method` to align it to the naming of other similar modules.

product_contract_transmit_method/models/sale_order.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+class SaleOrder(models.Model):
+
+    _inherit = "sale.order"
+
+    @api.multi
+    def _prepare_contract_value(self, contract_template):
+        contract_values = super()._prepare_contract_value(contract_template)
+        contract_values["transmit_method_id"] = self.transmit_method_id.id
```
bodi000: usually the simple quotas are used for fields, like so `'transmit_method_id'`

bodi000 (changes requested):
IMO some small changes are needed. thank you

sbejaoui:
cc/ @len-foss

product_contract_transmit_method/README.rst
len-foss: +1

product_contract_transmit_method/models/sale_order.py
len-foss: It's 13.0 (Black) coding convention (easier to forward-port).

len-foss (approved):
I agree that the name should reflect better the module. Besides that, all good.

sbejaoui:
@gva-acsone , can you review this PR please

gva-acsone (approved):
LGTM Functional test

PR #491, opened 2020-03-23
[12.0] Set company_id earlier in the invoice preparation process

svalaeys:
Some of the methods that inherit from _onchange_partner_id on account.invoice depend on the company_id field of the invoice. 

You can find such an example in account_payment_partner.

It is therefore important to set this field prior to calling this method.


sbejaoui (approved):
LGTM, Thanks

pedrobaeza (approved):

Issue #492, opened 2020-03-23
[12.0] error in migration from 11.0


sergiocorato:
This function https://github.com/OCA/contract/blob/14e46b14006ba9391c9f7ccff5d0aa9794eb68e0/product_contract/migrations/12.0.3.0.0/pre-migration.py#L6 generate an error when upgrading from 11.0 because in pre-migration the field is not yet created.

```
2020-03-23 17:45:00,070 12921 ERROR pretecno odoo.sql_db: bad query:             UPDATE product_template
            SET recurring_invoicing_type = 'post-paid'
            WHERE recurring_rule_type = 'monthlylastday'
        
ERROR: ERRORE:  la colonna "recurring_rule_type" non esiste
LINE 3:             WHERE recurring_rule_type = 'monthlylastday'
```

pedrobaeza:
I have migrated from 11.0 without problem. Do you have latest code in 11.0?

sergiocorato:
mm, yes, I pull every time

pedrobaeza:
Well, as said, I haven't found this problem, but I haven't checked code.

sbejaoui:
This migration was for 12.0.2.0.0 to 12.0.3.0.0 to introduce the monthly last day prepaid option but apparently we forget about v 11.0 where the field wasn't there.

You can add a check on the version like here 

https://github.com/OCA/contract/blob/14e46b14006ba9391c9f7ccff5d0aa9794eb68e0/contract/migrations/12.0.4.0.0/post-migration.py#L57-L60

sergiocorato:
So the migration has to be done only if version is == '12.0.2.0.0' I understand, I propose a PR

pedrobaeza:
Closing this as PR is merged.

PR #493, opened 2020-03-24
[FIX] migration from 11.0

sergiocorato:
solve #492

### DIFF
--- a/product_contract/migrations/12.0.3.0.0/pre-migration.py
+++ b/product_contract/migrations/12.0.3.0.0/pre-migration.py
@@ -1,10 +1,14 @@
+from odoo.tools import parse_version
+
+
 def migrate(cr, version):
-    # pre-paid/post-paid becomes significant for monthlylastday too,
-    # make sure it has the value that was implied for previous versions.
-    cr.execute(
-        """\
-            UPDATE product_template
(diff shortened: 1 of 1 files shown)


product_contract/migrations/12.0.3.0.0/pre-migration.py
```diff
@@ -1,10 +1,16 @@
...
-            UPDATE product_template
-            SET recurring_invoicing_type = 'post-paid'
-            WHERE recurring_rule_type = 'monthlylastday'
-        """
-    )
+from openupgradelib import openupgrade
+from odoo.tools import parse_version
+
+
+@openupgrade.migrate()
```
sbejaoui: you don't need `openupgrade` here

```suggestion
```

product_contract/migrations/12.0.3.0.0/pre-migration.py
```diff
@@ -1,10 +1,16 @@
...
+def migrate(env, version):
```
sbejaoui:
```suggestion
def migrate(cr, version):
```

sbejaoui (changes requested):

sbejaoui (approved):

AaronHForgeFlow (approved):
Just tested and works :+1:

PR #496, opened 2020-04-01
[FIX]contract_mandate: Search mandate from correct company if we get …

javierjcf:
…into this method from the contract cron. In such case we could get a mandate from a diferent company of the contract one


pedrobaeza (approved):

pedrobaeza:
Fast-tracking as obvious:

/ocabot merge patch

PR #497, opened 2020-04-03
[IMP] agreement_legal_sale: Allow Multiple SP's From SO Line

osi-scampbell:
The following PR makes changes to the agreement_legal_sale module so if a user orders 3 units of a product that is_serviceprofile then 3 Service Profiles are created. This PR also adjusts the structure of the methods so they are more easily inherited. This module also needs to depend on agreement_serviceprofile or else when running by itself it will generate a key error.

### DIFF
--- a/agreement_legal_sale/models/sale_order.py
+++ b/agreement_legal_sale/models/sale_order.py
@@ -40,7 +40,7 @@ class SaleOrder(models.Model):
         """ Create line.product_uom_qty SP's """
         if line.product_id.product_tmpl_id.is_serviceprofile:
             for i in range(1, int(line.product_uom_qty)+1):
-                sp = self.env['agreement.serviceprofile'].\
+                self.env['agreement.serviceprofile'].\
                     create(self._get_sp_vals(line, order, i))
 
     def _get_agreement_line_vals(self, line):
--- a/agreement_legal_sale/models/sale_order.py
+++ b/agreement_legal_sale/models/sale_order.py
@@ -17,33 +17,48 @@ class SaleOrder(models.Model):
         res = super(SaleOrder, self)._action_confirm()
         for order in self:
             if order.agreement_template_id:
-                order.agreement_id = order.agreement_template_id.copy(default={
-                    'name': order.name,
-                    'code': order.name,
-                    'is_template': False,
-                    'sale_id': order.id,
-                    'partner_id': order.partner_id.id,
-                    'analytic_account_id': order.analytic_account_id and
-                    order.analytic_account_id.id or False,
-                })
+                order.agreement_id = order.\
+                    agreement_template_id.copy(default={
+                        'name': order.name,
+                        'code': order.name,
+                        'is_template': False,
+                        'sale_id': order.id,
+                        'partner_id': order.partner_id.id,
+                        'analytic_account_id': order.analytic_account_id and
+                        order.analytic_account_id.id or False,
+                    })
                 for line in order.order_line:
                     # Create agreement line
-                    self.env['agreement.line'].create({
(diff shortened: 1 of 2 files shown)


agreement_legal_sale/models/sale_order.py
```diff
@@ -17,34 +17,50 @@ def _action_confirm(self):
...
-                        })
+                    self.env['agreement.line'].\
+                        create(self._get_agreement_line_vals(line, order))
+                    # Create SP's based on product_id config
+                    if line.product_id.is_serviceprofile:
+                        self.create_sp_qty(line, order)
         return res
 
+    def create_sp_qty(self, line, order):
+        # Create line.product_uom_qty SP's
```
dreispt: This explains what the method does, so you could have made it a docstring
```suggestion
        """ Create line.product_uom_qty SP's """
```

agreement_legal_sale/models/sale_order.py
```diff
@@ -17,34 +17,50 @@ def _action_confirm(self):
...
+        if line.product_id.product_tmpl_id.is_serviceprofile:
+            for i in range(1, int(line.product_uom_qty)+1):
+                sp = self.env['agreement.serviceprofile'].\
+                    create(self._get_sp_vals(line, order))
+                sp.write({'name': sp.name + ' ' + str(i)})
```
dreispt: Nitpicking, but this looks simpler:
```suggestion
            ServiceProfile = self.env['agreement.serviceprofile']
            for i in range(line.product_uom_qty):
                sp = Serviceprofile.create(self._get_sp_vals(line))
                sp.write({'name': sp.name + ' ' + str(i + 1)})
```
Does the write need to be done after the `create()`?
Would it work writing it to the values dictionary before the create?

agreement_legal_sale/models/sale_order.py
```diff
@@ -17,34 +17,50 @@ def _action_confirm(self):
...
+
+    def _get_agreement_line_vals(self, line, order):
+        return {
+            'product_id': line.product_id.id,
+            'name': line.name,
+            'agreement_id': order.agreement_id.id,
+            'qty': line.product_uom_qty,
```
dreispt: The `order` argument is not necessary:
```suggestion
    def _get_agreement_line_vals(self, line):
        return {
            'product_id': line.product_id.id,
            'name': line.name,
            'agreement_id': line.order_id.agreement_id.id,
            'qty': line.product_uom_qty,
```

dreispt (approved):
LGTM, change makes sense
I suggested some minor improvements to the proposed code.

osi-scampbell:
@dreispt Changes made, thank you for the feedback

dreispt:
@osi-scampbell please fix this unused variable to get a green build: https://travis-ci.org/github/OCA/contract/jobs/672129797#L313

agreement_legal_sale/models/sale_order.py
```diff
@@ -17,34 +17,49 @@ def _action_confirm(self):
...
+                    # Create SP's based on product_id config
+                    if line.product_id.is_serviceprofile:
+                        self.create_sp_qty(line, order)
         return res
 
+    def create_sp_qty(self, line, order):
+        """ Create line.product_uom_qty SP's """
+        if line.product_id.product_tmpl_id.is_serviceprofile:
+            for i in range(1, int(line.product_uom_qty)+1):
+                sp = self.env['agreement.serviceprofile'].\
```
dreispt:
```suggestion
                self.env['agreement.serviceprofile'].\
```

bodedra (approved):
Code review. LGTM :+1:

PR #506, opened 2020-04-27
[WIP][IMP] agreement: Minor improvements

sergio-teruel:


ygol (approved):

pedrobaeza:
@sergio-teruel is this still WIP?

sergio-teruel:
Yes, I think... but maybe this PR can be merged...

agreement/security/ir.model.access.csv
```diff
@@ -1,3 +1,5 @@
 id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
 access_agreement_read,Read access on agreement to Employees,model_agreement,base.group_user,1,0,0,0
 access_agreement_full,Full access on agreement grp,model_agreement,base.group_no_one,1,1,1,1
+access_agreement_type_read,Read access on agreement type to Employees,model_agreement_type,base.group_user,1,0,0,0
+access_agreement_type_full,Full access on agreement type grp,model_agreement_type,base.group_no_one,1,1,1,1
```
pedrobaeza: Why giving full access to developer mode? I think it should be to settings group.

sergio-teruel:
I have kept the original line of the module

pedrobaeza:
I think both are incorrect: it should be settings group. If not, anyone can change that switching to developer mode. @ygol @max3903 do you agree?

pedrobaeza:
Please do the switch of the permissions as @ygol seems to agree and they are logical.

agreement/security/ir.model.access.csv
sergio-teruel: Ok. What groups can I use here??

sergio-teruel:
@pedrobaeza Ok. What groups can I use here??

pedrobaeza:
For me it should be `base.group_system`

agreement/models/agreement.py
```diff
@@ -10,7 +10,7 @@ class Agreement(models.Model):
     _description = 'Agreement'
     _inherit = ["mail.thread", "mail.activity.mixin"]
 
-    code = fields.Char(required=True, copy=False)
+    code = fields.Char(required=True)
```
alexis-via, line 13: why did you do that ? There is a unicity constrain...

sergio-teruel: If field is required you can not duplicate record. Maybe it would be necessary to create a sequence..

carlosdauden:
Another option could be to add "(Copy)" to the code:
https://github.com/odoo/odoo/blob/c5a8051a2c840b2be545eee2f173261a908de737/addons/account/models/account.py#L253
:sweat_smile:

agreement/models/agreement.py
carlosdauden: Another option could be to add "(Copy)" to the code:
https://github.com/odoo/odoo/blob/c5a8051a2c840b2be545eee2f173261a908de737/addons/account/models/account.py#L253
:sweat_smile:

sergio-teruel:
Changes done!!

pedrobaeza (approved):

PR #507, opened 2020-04-28
[12.0][FIX] contract: Select the correct sections/notes to be invoiced

ernestotejeda:
Select the correct sections/notes to be invoiced.
For example: Empty contract line sections shouldn't be invoiced.

Cc @Tecnativa TT23706

### DIFF
--- a/contract/models/abstract_contract_line.py
+++ b/contract/models/abstract_contract_line.py
@@ -125,9 +125,27 @@ class ContractAbstractContractLine(models.AbstractModel):
         required=True,
         ondelete='cascade',
     )
-    display_type = fields.Selection([
-        ('line_section', "Section"),
-        ('line_note', "Note")], default=False, help="Technical field for UX purpose.")
+    display_type = fields.Selection(
+        selection=[
+            ('line_section', "Section"),
+            ('line_note', "Note"),
+        ],
+        default=False,
+        help="Technical field for UX purpose."
+    )
+    note_invoicing_mode = fields.Selection(
+        selection=[
+            ('with_previous_line', 'With previous line'),
+            ('with_next_line', 'With next line'),
+            ('custom', 'Custom'),
+        ],
+        default='with_previous_line',
+        help="Defines when the Note is invoiced:\n"
+             "- With previous line: If the previous line can be invoiced.\n"
+             "- With next line: If the next line can be invoiced.\n"
+             "- Custom: Depending on the recurrence to be define."
+    )
+    is_recurring_note = fields.Boolean(compute="_compute_is_recurring_note")
 
     @api.model
     def _get_default_recurring_invoicing_offset(
@@ -140,6 +158,13 @@ class ContractAbstractContractLine(models.AbstractModel):
             return 0
         else:
             return 1
+
+    def is_recurring_note(self):
+        for record in self:
+            record.is_recurring_note = (
+                record.display_type == 'line_note'
+                and record.note_invoicing_mode == 'custom'
+            )
 
     @api.depends('recurring_invoicing_type', 'recurring_rule_type')
     def _compute_recurring_invoicing_offset(self):
--- a/contract/models/contract.py
+++ b/contract/models/contract.py
@@ -194,7 +194,8 @@ class ContractContract(models.Model):
     def _compute_recurring_next_date(self):
         for contract in self:
             recurring_next_date = contract.contract_line_ids.filtered(
-                lambda l: l.recurring_next_date and not l.is_canceled
+                lambda l: (l.recurring_next_date and not l.is_canceled
+                           and (not l.display_type or l.is_recurring_note))
             ).mapped('recurring_next_date')
             if recurring_next_date:
                 contract.recurring_next_date = min(recurring_next_date)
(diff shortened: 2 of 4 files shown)


contract/models/contract.py
```diff
@@ -412,11 +413,28 @@ def _get_lines_to_invoice(self, date_ref):
...
-            and l.recurring_next_date <= date_ref
-        )
+
+        def can_be_invoiced(l):
+            return (not l.is_canceled and l.recurring_next_date
+                    and l.recurring_next_date <= date_ref)
+
+        res = current_section = self.env['contract.line']
+        add_current_section = add_next_note = False
+        for line in self.contract_line_ids.sorted('sequence'):
```
pedrobaeza: Why do you need `sorted`?

ernestotejeda: If I have contract lines sorted by 'sequence', iterating those lines with a for loop I can assure that lines that go after a certain section correspond to that section and lines inmediatly before a note correspond to that note. With that, i can determine with a single loop, which section/Notes and which product lines can be invoiced.

contract/models/abstract_contract_line.py
```diff
@@ -125,9 +125,27 @@ class ContractAbstractContractLine(models.AbstractModel):
...
-        ('line_note', "Note")], default=False, help="Technical field for UX purpose.")
+    display_type = fields.Selection(
+        selection=[
+            ('line_section', "Section"),
+            ('line_note', "Note"),
+        ],
+        default=False,
+        help="Technical field for UX purpose."
+    )
+    notes_invoicing_mode = fields.Selection(
```
pedrobaeza: Avoid plural

contract/models/abstract_contract_line.py
```diff
@@ -125,9 +125,27 @@ class ContractAbstractContractLine(models.AbstractModel):
...
+            ('with_next_line', 'With next line'),
+            ('custom', 'Custom'),
+        ],
+        default='with_previous_line',
+        help="Defines when the Note is invoiced:\n"
+             "- With previous line: If the previous line can be invoiced.\n"
+             "- With next line: If the previous line can be invoiced.\n"
+             "- Custom: Depending on the recurrence to be define."
+    )
+    is_recurring_note = fields.Boolean(compute="_compute_is_recurring_note")
```
pedrobaeza, line 148: No need of this extra field as the visibility computation can be done with domain and it's very simple

contract/views/abstract_contract_line.xml
```diff
@@ -8,6 +8,7 @@
         <field name="arch" type="xml">
             <form>
                 <field name="display_type" invisible="1"/>
+                <field name="notes_invoicing_mode" invisible="1"/>
```
pedrobaeza: No need to add it twice

contract/views/abstract_contract_line.xml
```diff
@@ -62,7 +66,7 @@
                             </div>
                         </group>
                     </group>
-                    <group name="recurrence_info" attrs="{'invisible': [('display_type', '!=', False)]}">
+                    <group name="recurrence_info" attrs="{'invisible': ['|', ('display_type', '=', 'line_section'), ('display_type', '=', 'line_note'), ('note_invoicing_mode', '!=', 'custom')]}">
```
pedrobaeza:
```suggestion
                    <group name="recurrence_info" attrs="{'invisible': ['|', ('display_type', '=', 'line_section'), '&', ('display_type', '=', 'line_note'), ('note_invoicing_mode', '!=', 'custom')]}">
```

pedrobaeza:
@sbejaoui here we are fixing the problem on the invoiceability of sections and notes, with a good approach IMO:

- Sections are only invoiced if one of the included lines in it is invoiceable.
- Notes can be configured for being invoiceable with the previous or next line, or to be able to define its recurrency independently.

pedrobaeza (approved):

carlosdauden (approved):

sbejaoui (approved):
LGTM, Thanks

contract/models/abstract_contract_line.py
```diff
@@ -125,9 +125,27 @@ class ContractAbstractContractLine(models.AbstractModel):
...
+    note_invoicing_mode = fields.Selection(
+        selection=[
+            ('with_previous_line', 'With previous line'),
+            ('with_next_line', 'With next line'),
+            ('custom', 'Custom'),
+        ],
+        default='with_previous_line',
+        help="Defines when the Note is invoiced:\n"
+             "- With previous line: If the previous line can be invoiced.\n"
+             "- With next line: If the previous line can be invoiced.\n"
```
sbejaoui:
```suggestion
             "- With next line: If the next line can be invoiced.\n"
```

sbejaoui:
@gva-acsone , This should fix the issue we faced as we can now define when the note is invoiced:
- With previous line: If the previous line can be invoiced.
- With next line: If the next line can be invoiced.
- Custom: Depending on the recurrence to be define.

pedrobaeza:
Let's cancel build for changing that text.

pedrobaeza:
Changed now, so let's merge again, but better to put it as minor change, due to fields changing:

/ocabot merge minor

PR #446, opened 2019-12-19
13.0 mig contract

devCbssolutionsRo:
contact/contract   for version 13.0  ( from 12.0.5.0)

PR #511, opened 2020-05-13
[UPD] Contract type on LINE and Translate MN

altanmur:


pedrobaeza:
Thanks for the PR. Are you trying to migrate the module to v13? There's already another PR with the migration in #446 and you haven't preserved commit history, so closing as duplicated.

PR #522, opened 2020-06-02
[12.0][FIX] contract: Currency is not editable + pricelist from partner

pedrobaeza:
Previous related field was not accurated nor editable. Now the field is got properly from a computed field.

Reviewing this, as the currency was taken (and it continues being taken) from the partner pricelist if no pricelist is explicitly set, automatic price should use the same logic for using partner pricelist.

cc @Tecnativa TT24071


sbejaoui (approved):
LGTM, Thank you

rafaelbn (approved):
Thanks :+1:

ernestotejeda (approved):
Code review :+1:

PR #523, opened 2020-06-05
[12.0][IMP] contract: Allow to set a manual currency

pedrobaeza:
Done through an extra field + inverse in the existing compute. This way, we have the flexibility of auto-remove custom value if it matches the computed one + we don't need extra changes in the rest of the code.

cc @Tecnativa TT24071

### DIFF
--- a/contract/models/contract.py
+++ b/contract/models/contract.py
@@ -34,8 +34,12 @@ class ContractContract(models.Model):
     )
     currency_id = fields.Many2one(
         compute="_compute_currency_id",
+        inverse="_inverse_currency_id",
         comodel_name="res.currency",
         string="Currency",
+    )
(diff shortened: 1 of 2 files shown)


carlosdauden (approved):
I think simply assigning `inverse=lambda self: self` would avoid all this code, but since I know you don't like it and that in v13 that same inverse is achieved with `readonly=False`, which is to your liking, I have no problem approving this PR. :wink:

pedrobaeza:
Well, I thought about that, but the problem using `lambda` (or the equivalent whole method), is that the manual currency is assigned as soon as you save the first time, so the compute remains only as a "default value". That's why I have put full inverse method with the check for a different value.

rafaelbn (approved):
:+1:

PR #526, opened 2020-06-12
[12.0][IMP] contract: Add link to contract line in contract chatter messages

qgroulard:
On contracts, it might be difficult to know which contract line chatter messages refer to.
Thus, this PR adds a link to the contract line form view in the messages.


sbejaoui (approved):
LGTM, thanks

gva-acsone (approved):
LGTM Functional test

marcelsavegnago (approved):

pedrobaeza (changes requested):
Please adapt all of the existing translations, or the fuzzy match will break all languages, not having the number of needed placeholders.

And this is `[IMP]`, not `[ADD]`, as this is not a new module.

sbidoul:
> or the fuzzy match will break all languages

@pedrobaeza do you think we should disable fuzzy matching in weblate?

pedrobaeza:
If not 100% match, I think it should be disabled. Anyway, doing this change will force to re-translate that strings, so it's better to make the substitution in .po files, as there's no translatable content added. I would admit a refactoring of the strings for letting outside the placeholders and avoid errors due to this.

But wait...

pedrobaeza (changes requested):
`_(` is applied after placeholder replacing, so this is not going to be translated ever.

sbejaoui:
@qgroulard can you check @pedrobaeza comment,

rousseldenis (approved):
Code review

rousseldenis:
@qgroulard Could you rebase ?

rousseldenis:
@pedrobaeza Could you update your review ?

pedrobaeza:
It still has the same problem. You should apply .format after the call to translate _

rousseldenis:
@qgroulard Indeed. If you let like this, source strings will have different values to be translated...

rousseldenis:
@qgroulard

rousseldenis (approved):
@pedrobaeza Could you update your comments?

pedrobaeza:
The `for` word doesn't make sense in this context, as now we are referring directly to the line, not to the product.

qgroulard:
> The `for` word doesn't make sense in this context, as now we are referring directly to the line, not to the product.

The href leads to the line but the text displays the product name.

rousseldenis:
> The `for` word doesn't make sense in this context, as now we are referring directly to the line, not to the product.

Product is just the alias for line name. So, it is not really referred to product only.

image

pedrobaeza (approved):

pedrobaeza:
Yes, it's OK, as the link text is still the product.

pedrobaeza:
The PO files are not correct after this change. Please fix them.

qgroulard:
@pedrobaeza I believe the " not escaped was the cause of the error.
Could you please retry ?

PR #527, opened 2020-06-13
[10.0] [FIX] Remove part of the optimization of invoice creation

fcayre:
The `norecompute` context manager was used around a call to method
`_create_invoice`, which is overriden by at least the
contract_payment_auto module (which is arguably a bug in itself as
the method name starts with an underscore). This breaks some
automations like the closing of invoices with no residual after
payment.

To fix this without changing the de facto API of this module, the
`norecompute context manager` was removed and the
`_onchange_invoice_line_ids` method of account.invoice model is now
explicitely called.

The invoice creation context is now set around the `_create_invoice`
method call so that it can be used in the module that override it too.

### DIFF
--- a/contract/models/account_analytic_account.py
+++ b/contract/models/account_analytic_account.py
@@ -258,15 +258,13 @@ class AccountAnalyticAccount(models.Model):
     def _create_invoice(self):
         self.ensure_one()
         # Re-read contract with correct company
-        _self = self.with_context(self.get_invoice_context())
-        invoice_vals = _self._prepare_invoice()
-        invoice = _self.env['account.invoice'].create(invoice_vals)
+        invoice_vals = self._prepare_invoice()
+        invoice = self.env['account.invoice'].create(invoice_vals)
         # Lines are read from an env where expensive values are precomputed
         for line in self.recurring_invoice_line_ids:
-            invoice_line_vals = _self._prepare_invoice_line(line, invoice.id)
-            _self.env['account.invoice.line'].create(invoice_line_vals)
-        # Update next invoice date for current contract
-        _self.recurring_next_date = _self.env.context['next_date']
+            invoice_line_vals = self._prepare_invoice_line(line, invoice.id)
+            self.env['account.invoice.line'].create(invoice_line_vals)
+            invoice._onchange_invoice_line_ids()
         return invoice
 
     @api.multi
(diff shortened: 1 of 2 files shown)


fcayre:
Hi @Yajo please review this PR as it would basically backout some of your optimizations. The rational is explained in the commit message: the `_create_invoice` method is a _de facto_ public API as it is overridden by the contract_payment_auto module (which is arguably a bug if your consider python conventions regarding function naming). As a consequence, the `_do_auto_pay` and the `_pay_invoice` methods defined in contract_payment_auto are called from `_create_invoice`, which in turn call the `s2s_do_transaction` payment method and the corresponding methods from _any_ payment module... With this in mind, it is no more reasonable to use the `norecompute` context manager around it, as it leads to a lot of difficult to find-out bugs.
My PR may be sub-optimal however (even with this restriction in mind), that's why I ask you to review it. Many thanks.

pedrobaeza:
Sorry, my team (including @Yajo) are not working anymore on v10 (but later versions), so we can't review it anymore with confidence. Anyway, this has been working since a lot without problems. You should find other reviewers.

mourad-ehm:
it's possible to make inheritance of recurring_create_invoice method like this 
```python
@api.multi
    def recurring_create_invoice(self, limit=None):
        """ If automatic payment is enabled, perform auto pay actions. """
        invoices = super(AccountAnalyticAccount, self).recurring_create_invoice()
        for invoice in invoices:
             if inv.is_auto_pay:
                  inv._do_auto_pay(invoice)
        return invoices
```
May be resolve this problem ?

fcayre:
@mourad-ehm this is the easiest change indeed, but it does not solve the problem of the invoices being created with norecompute that is a very aggressive optimization, which consequences depend on the other modules you have installed (this is why "works for me" is not helping here). I will try a minimum patch adding an `invoices.recompute()` instead of or after the `compute_taxes()` call in recurring_create_invoice; maybe it is enough and would be more respectful of the previous optimization work.

mourad-ehm:
I think that "_create_invoice" is privet method and should not be inherited. May be add a comment on description off this method can alert developers on this problem.

fcayre:
Hello @alexis-via you may be interested in this PR, if you have time your review would be more than welcome !

PR #528, opened 2020-06-18
[12.0][FIX] - duplicate payement mode in supplier form views

flotho:
allow to prevent duplicate fields on supplier views

### DIFF
--- a/contract_payment_mode/views/contract_view.xml
+++ b/contract_payment_mode/views/contract_view.xml
@@ -45,8 +45,8 @@
 		<field name="priority">18</field>
         <field name="inherit_id" ref="contract.contract_contract_supplier_form_view"/>
         <field name="arch" type="xml">
-            <field name="partner_id" position="after">
-               <field name="payment_mode_id" domain="[('payment_type', '=', 'outbound')]"/>
+            <field name="payment_mode_id" position="attributes">
+               <attribute name="domain" >[('payment_type', '=', 'outbound')]</attribute>
             </field>
         </field>
     </record>


sbejaoui (approved):
LGTM, thank you

pedrobaeza (changes requested):
Don't replace, but change `domain` attribute if that's the problem.

flotho:
@pedrobaeza the point is that this field is also called in another view. So it's really a duplicate issue in the field view get and not a domain issue.
That's why I used replaced

sbejaoui:
@flotho ,

@pedrobaeza point is to avoid losing other attributes from the base view. 

It's not the case today, but it make sense in a scalability logic if one day we add `attrs` or change the label with `string`.

Changing the domain will resolve the duplicate issue

flotho:
@pedrobaeza @sbejaoui ,

Thanks I read too fast and too late. I change this

pedrobaeza:
Travis is red. I think you have to encode the quotes. Please also squash in one commit.

flotho:
Hi @pedrobaeza ,

Travsi is Green and runbot failed.
i didin't succeed in getting the logs to see if I can do something from my side, sorry for this newbie question

pedrobaeza:
Runbot error:
```

Field 'domain' used in attributes must be present in view but is missing:
 - 'domain' in attrs="{'invisible': [('domain', '!=', 'purchase')]}"
 - 'domain' in attrs="{'invisible': [('domain', '!=', 'sale')]}"

Error context:
View `invoice.button.agreement.form`
[view_id: 1086, xml_id: n/a, model: agreement, parent_id: 334]
None" while parsing /home/odoo/dependencies/contract/agreement_account/views/agreement.xml:36, near
<record id="agreement_form" model="ir.ui.view">
        <field name="name">invoice.button.agreement.form</field>
        <field name="model">agreement</field>
        <field name="inherit_id" ref="agreement.agreement_form"/>
            <div name="button_box" position="inside">
                <button class="oe_stat_button" type="action" name="%(account.action_invoice_refund_out_tree)d" context="{'search_default_agreement_id': active_id, 'default_agreement_id': active_id, 'default_partner_id': partner_id}" attrs="{'invisible': [('domain', '!=', 'sale')]}" icon="fa-pencil-square-o">
                        <field string="Customer Invoices" name="out_invoice_count" widget="statinfo"/>
                </button>
                <button class="oe_stat_button" type="action" name="%(account_invoice_in_invoice_refund_action)d" context="{'search_default_agreement_id': active_id, 'default_agreement_id': active_id, 'default_partner_id': partner_id}" attrs="{'invisible': [('domain', '!=', 'purchase')]}" icon="fa-pencil-square-o">
                        <field string="Vendor Bills" name="in_invoice_count" widget="statinfo"/>
                </button>

            </div>
        </field>
    </record>
```

marylla (approved):
Thanks for solving this. I was very confused about the duplicated field with the wrong domain.
I would have done this with the attribute node as well, so looks good to me. :+1: 
And it also works checking the runbot.

@pedrobaeza Could you re-review, because I think it is correct now.

pedrobaeza (approved):

PR #533, opened 2020-07-14
[13.0][MIG] contract: Migration to 13.0

pedrobaeza:
- Standard procedure
- Adapt invoice creation to new account.move
- Tests adapted

@Tecnativa TT19723


pedrobaeza:
I have added an improvement for handling recurrence information at header level and thus simplify the UI a lot for such cases, that are the most frequent. GitHub is not refreshing the PR anyway...

em230418:
```python3
from datetime import date, timedelta


def create_contract(self):
    partner = self.env.ref("base.partner_demo")
    recurring_rule_type = "yearly"

    today = date.today()
    recurring_next_date = today + timedelta(days=7)

    product = self.env.ref("product.product_product_1")
    contract = self.env["contract.contract"].create({
        "name": "{}'s SaaS Contract".format(partner.name),
        "partner_id": partner.id,

        "contract_line_ids": [(0, 0, {
            "name": product.name,
            "product_id": product.id,
            "uom_id": self.env.ref("uom.product_uom_unit").id,
            "quantity": 1,
            "price_unit": product.list_price,
            "recurring_interval": 1,
            "recurring_rule_type": recurring_rule_type,
            "recurring_invoicing_type": "pre-paid",
            "recurring_next_date": recurring_next_date,
            "is_cancel_allowed": False,
            "date_start": today,
            "date_end": today + timedelta(days=365),
        })]
    })

    print(contract.contract_line_ids.recurring_next_date, recurring_next_date)
    assert contract.contract_line_ids.recurring_next_date == recurring_next_date

create_contract(self)
```

On odoo 12.0 `create_contract(self)` runs successfully, on odoo 13.0 assertion error

UPD: added create_contract(self) to snippet above
UPD2: setting `"recurring_next_date": recurring_next_date` instead of `"recurring_next_date": today + timedelta(days=7)` to be more obvious

pedrobaeza:
I don't get the problem. Please give reproducible steps for the problem, not pasting a bunch of code. If that's custom code, of course you need to adjust it to new version with the new requirements.

em230418:
@pedrobaeza, I edited my code above.

In odoo 12 and odoo 13:
1. Install contract module on fresh database
```
odoo-bin -d test333 -i contract --stop-after-init
```
2. Create file test.py in current directory with code from https://github.com/OCA/contract/pull/533#issuecomment-659366730
3. Execute test.py on odoo shell
```
cat test.py |  odoo-bin shell -d test333
```

In odoo 12.0 two equal dates will be printed, in odoo 13.0 with this PR two different dates will be printed and AssertionError is raised.

The problem is that value of `contract.contract_line_ids.recurring_next_date` is not the same as `recurring_next_date` which is passed to `create` method.

pedrobaeza:
Yes, indeed, because you are creating it directly... You need to adjust your code as stated, or incongruences will arise.

em230418:
I see. Thanks for the tip!

contract/models/contract.py
```diff
@@ -0,0 +1,509 @@
...
+                )
+            ).mapped("recurring_next_date")
+            if recurring_next_date:
+                contract.recurring_next_date = min(recurring_next_date)
+            else:
+                contract.recurring_next_date = False
+
+    def _inverse_recurring_next_date(self):
+        for record in self:
+            record.contract_line_ids.recurring_next_date = record.recurring_next_date
```
em230418: Perhaps this change?
```diff
diff --git a/contract/models/contract.py b/contract/models/contract.py
index 090ce7fe..055ae0fe 100644
--- a/contract/models/contract.py
+++ b/contract/models/contract.py
@@ -212,7 +212,7 @@ class ContractContract(models.Model):
                 contract.recurring_next_date = False
 
     def _inverse_recurring_next_date(self):
-        for record in self:
+        for record in self.filtered(lambda x: not x.line_recurrency):
             record.contract_line_ids.recurring_next_date = record.recurring_next_date
 
     @api.depends("contract_line_ids.create_invoice_visibility")
```

Without this I cannot create contract:
- "Recurrency at line level? = Yes"
- One product line with filled "Date of Next Invoice" (for example 07/22/2020)

When I press "Save" on contract, it fails with message "You must supply a date of next invoice for contract line 'xxxx'" which is not true.

em230418:
Cannot add line on existing contract with recurrency at line level. Here is what I do:
- Create contract, "Recurrency at line level? - Yes", add line, save line, save contract.
- Edit contract, add one more line, save line fails with following message: https://i.imgur.com/gYMKkMX.png
Stack trace:
```
test_1      | Traceback (most recent call last):
test_1      |   File "/usr/lib/python3/dist-packages/odoo/api.py", line 745, in get
test_1      |     value = self._data[field][record._ids[0]]
test_1      | KeyError: <NewId origin=3>
test_1      | 
test_1      | During handling of the above exception, another exception occurred:
test_1      | 
test_1      | Traceback (most recent call last):
test_1      |   File "/usr/lib/python3/dist-packages/odoo/fields.py", line 996, in __get__
test_1      |     value = env.cache.get(record, self)
test_1      |   File "/usr/lib/python3/dist-packages/odoo/api.py", line 751, in get
test_1      |     raise CacheMiss(record, field)
test_1      | odoo.exceptions.CacheMiss: ('contract.line(<NewId origin=3>,).date_start', None)
test_1      | 
test_1      | During handling of the above exception, another exception occurred:
test_1      | 
test_1      | Traceback (most recent call last):
test_1      |   File "/usr/lib/python3/dist-packages/odoo/api.py", line 745, in get
test_1      |     value = self._data[field][record._ids[0]]
test_1      | KeyError: <NewId origin=3>
test_1      | 
test_1      | During handling of the above exception, another exception occurred:
test_1      | 
test_1      | Traceback (most recent call last):
test_1      |   File "/usr/lib/python3/dist-packages/odoo/http.py", line 624, in _handle_exception
test_1      |     return super(JsonRequest, self)._handle_exception(exception)
test_1      |   File "/usr/lib/python3/dist-packages/odoo/http.py", line 310, in _handle_exception
test_1      |     raise pycompat.reraise(type(exception), exception, sys.exc_info()[2])
test_1      |   File "/usr/lib/python3/dist-packages/odoo/tools/pycompat.py", line 14, in reraise
test_1      |     raise value
test_1      |   File "/usr/lib/python3/dist-packages/odoo/http.py", line 670, in dispatch
test_1      |     result = self._call_function(**self.params)
test_1      |   File "/usr/lib/python3/dist-packages/odoo/http.py", line 350, in _call_function
test_1      |     return checked_call(self.db, *args, **kwargs)
test_1      |   File "/usr/lib/python3/dist-packages/odoo/service/model.py", line 94, in wrapper
test_1      |     return f(dbname, *args, **kwargs)
test_1      |   File "/usr/lib/python3/dist-packages/odoo/http.py", line 339, in checked_call
test_1      |     result = self.endpoint(*a, **kw)
test_1      |   File "/usr/lib/python3/dist-packages/odoo/http.py", line 916, in __call__
test_1      |     return self.method(*args, **kw)
test_1      |   File "/usr/lib/python3/dist-packages/odoo/http.py", line 515, in response_wrap
test_1      |     response = f(*args, **kw)
test_1      |   File "/usr/lib/python3/dist-packages/odoo/addons/web/controllers/main.py", line 1322, in call_kw
test_1      |     return self._call_kw(model, method, args, kwargs)
test_1      |   File "/usr/lib/python3/dist-packages/odoo/addons/web/controllers/main.py", line 1314, in _call_kw
test_1      |     return call_kw(request.env[model], method, args, kwargs)
test_1      |   File "/usr/lib/python3/dist-packages/odoo/api.py", line 387, in call_kw
test_1      |     result = _call_kw_multi(method, model, args, kwargs)
test_1      |   File "/usr/lib/python3/dist-packages/odoo/api.py", line 374, in _call_kw_multi
test_1      |     result = method(recs, *args, **kwargs)
test_1      |   File "/usr/lib/python3/dist-packages/odoo/models.py", line 6023, in onchange
test_1      |     new_lines.mapped(subname)
test_1      |   File "/usr/lib/python3/dist-packages/odoo/models.py", line 5206, in mapped
test_1      |     recs = recs._mapped_func(operator.itemgetter(name))
test_1      |   File "/usr/lib/python3/dist-packages/odoo/models.py", line 5165, in _mapped_func
test_1      |     vals = [func(rec) for rec in self]
test_1      |   File "/usr/lib/python3/dist-packages/odoo/models.py", line 5165, in <listcomp>
test_1      |     vals = [func(rec) for rec in self]
test_1      |   File "/usr/lib/python3/dist-packages/odoo/models.py", line 5610, in __getitem__
test_1      |     return self._fields[key].__get__(self, type(self))
test_1      |   File "/usr/lib/python3/dist-packages/odoo/fields.py", line 1023, in __get__
test_1      |     value = env.cache.get(record, self)
test_1      |   File "/usr/lib/python3/dist-packages/odoo/api.py", line 751, in get
test_1      |     raise CacheMiss(record, field)
test_1      | odoo.exceptions.CacheMiss: ('contract.line(<NewId origin=3>,).date_start', None)
```

No idea what causes it. Happens on my Odoo version 13.0-20200417 with patch from https://github.com/OCA/contract/pull/533#discussion_r458576391 and runbot from this PR without mentioned patch at the time of writing

em230418:
> Create contract, "Recurrency at line level? - Yes", add line, save line, save contract.

Pardon me. That would work with patch from https://github.com/OCA/contract/pull/533#discussion_r458576391

Without this patch do this instead:
- Create contract, "Recurrency at line level? - Yes", save contract
- Edit contract, add line, save line, save contract
- Edit contract, add one more line, save line fails with following message: https://i.imgur.com/gYMKkMX.png

pedrobaeza:
@em230418 I have applied your suggestion. Is now working properly?

em230418:
@pedrobaeza found another bug. Previous one are fixed.

- Create contract, "Recurrency at line level? - Yes", add line, save line, save contract
- Edit contract, check out "Recurrency at line level?" and error will appear.

```
odoo_1      | Traceback (most recent call last):
odoo_1      |   File "/usr/lib/python3/dist-packages/odoo/api.py", line 745, in get
odoo_1      |     value = self._data[field][record._ids[0]]
odoo_1      | KeyError: <NewId origin=443>
odoo_1      | 
odoo_1      | During handling of the above exception, another exception occurred:
odoo_1      | 
odoo_1      | Traceback (most recent call last):
odoo_1      |   File "/usr/lib/python3/dist-packages/odoo/fields.py", line 996, in __get__
odoo_1      |     value = env.cache.get(record, self)
odoo_1      |   File "/usr/lib/python3/dist-packages/odoo/api.py", line 751, in get
odoo_1      |     raise CacheMiss(record, field)
odoo_1      | odoo.exceptions.CacheMiss: ('contract.line(<NewId origin=443>,).date_start', None)
odoo_1      | 
odoo_1      | During handling of the above exception, another exception occurred:
odoo_1      | 
odoo_1      | Traceback (most recent call last):
odoo_1      |   File "/usr/lib/python3/dist-packages/odoo/api.py", line 745, in get
odoo_1      |     value = self._data[field][record._ids[0]]
odoo_1      | KeyError: <NewId origin=443>
odoo_1      | 
odoo_1      | During handling of the above exception, another exception occurred:
odoo_1      | 
odoo_1      | Traceback (most recent call last):
odoo_1      |   File "/usr/lib/python3/dist-packages/odoo/http.py", line 624, in _handle_exception
odoo_1      |     return super(JsonRequest, self)._handle_exception(exception)
odoo_1      |   File "/usr/lib/python3/dist-packages/odoo/http.py", line 310, in _handle_exception
odoo_1      |     raise pycompat.reraise(type(exception), exception, sys.exc_info()[2])
odoo_1      |   File "/usr/lib/python3/dist-packages/odoo/tools/pycompat.py", line 14, in reraise
odoo_1      |     raise value
odoo_1      |   File "/usr/lib/python3/dist-packages/odoo/http.py", line 670, in dispatch
odoo_1      |     result = self._call_function(**self.params)
odoo_1      |   File "/usr/lib/python3/dist-packages/odoo/http.py", line 350, in _call_function
odoo_1      |     return checked_call(self.db, *args, **kwargs)
odoo_1      |   File "/usr/lib/python3/dist-packages/odoo/service/model.py", line 94, in wrapper
odoo_1      |     return f(dbname, *args, **kwargs)
odoo_1      |   File "/usr/lib/python3/dist-packages/odoo/http.py", line 339, in checked_call
odoo_1      |     result = self.endpoint(*a, **kw)
odoo_1      |   File "/usr/lib/python3/dist-packages/odoo/http.py", line 916, in __call__
odoo_1      |     return self.method(*args, **kw)
odoo_1      |   File "/usr/lib/python3/dist-packages/odoo/http.py", line 515, in response_wrap
odoo_1      |     response = f(*args, **kw)
odoo_1      |   File "/usr/lib/python3/dist-packages/odoo/addons/web/controllers/main.py", line 1322, in call_kw
odoo_1      |     return self._call_kw(model, method, args, kwargs)
odoo_1      |   File "/usr/lib/python3/dist-packages/odoo/addons/web/controllers/main.py", line 1314, in _call_kw
odoo_1      |     return call_kw(request.env[model], method, args, kwargs)
odoo_1      |   File "/usr/lib/python3/dist-packages/odoo/api.py", line 387, in call_kw
odoo_1      |     result = _call_kw_multi(method, model, args, kwargs)
odoo_1      |   File "/usr/lib/python3/dist-packages/odoo/api.py", line 374, in _call_kw_multi
odoo_1      |     result = method(recs, *args, **kwargs)
odoo_1      |   File "/usr/lib/python3/dist-packages/odoo/models.py", line 6023, in onchange
odoo_1      |     new_lines.mapped(subname)
odoo_1      |   File "/usr/lib/python3/dist-packages/odoo/models.py", line 5206, in mapped
odoo_1      |     recs = recs._mapped_func(operator.itemgetter(name))
odoo_1      |   File "/usr/lib/python3/dist-packages/odoo/models.py", line 5165, in _mapped_func
odoo_1      |     vals = [func(rec) for rec in self]
odoo_1      |   File "/usr/lib/python3/dist-packages/odoo/models.py", line 5165, in <listcomp>
odoo_1      |     vals = [func(rec) for rec in self]
odoo_1      |   File "/usr/lib/python3/dist-packages/odoo/models.py", line 5610, in __getitem__
odoo_1      |     return self._fields[key].__get__(self, type(self))
odoo_1      |   File "/usr/lib/python3/dist-packages/odoo/fields.py", line 1023, in __get__
odoo_1      |     value = env.cache.get(record, self)
odoo_1      |   File "/usr/lib/python3/dist-packages/odoo/api.py", line 751, in get
odoo_1      |     raise CacheMiss(record, field)
odoo_1      | odoo.exceptions.CacheMiss: ('contract.line(<NewId origin=443>,).date_start', None)
```

contract/models/account_move.py
```diff
@@ -0,0 +1,21 @@
+# Copyright 2016 Tecnativa - Carlos Dauden
+# Copyright 2018 ACSONE SA/NV.
+# Copyright 2020 Tecnativa - Pedro M. Baeza
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class AccountInvoice(models.Model):
```
yajo: To match the model name:

```suggestion
class AccountMove(models.Model):
```

contract/models/contract.py
```diff
@@ -106,26 +107,22 @@ class ContractContract(models.Model):
...
-            self.env["account.invoice.line"]
+            self.env["account.move.line"]
             .search([("contract_line_id", "in", self.contract_line_ids.ids,)])
-            .mapped("invoice_id")
-        )
-        invoices |= self.env["account.invoice"].search(
-            [("old_contract_id", "=", self.id)]
+            .mapped("move_id")
         )
+        invoices |= self.env["account.move"].search([("old_contract_id", "=", self.id)])
```
yajo, line 133: I don't understand... :thinking: 

`old_contract_id` is supposed to be here only for migration purposes. However, it's being used here for real business logic. Then is it needed to work or not?

If yes, then it should be not marked as "migration only field". If not, then please add some comments clarifying its usage where used (like here).

contract/models/contract.py
```diff
@@ -472,21 +418,15 @@ def _prepare_recurring_invoices_values(self, date_ref=False):
...
-                invoice_values.setdefault("invoice_line_ids", [])
-                invoice_line_values = line._prepare_invoice_line(
-                    invoice_values=invoice_values,
-                )
-                if invoice_line_values:
-                    invoice_values["invoice_line_ids"].append(
-                        (0, 0, invoice_line_values)
-                    )
-            invoices_values.append(invoice_values)
+                invoice_line_vals = line._prepare_invoice_line(move_form=move_form)
```
yajo, line 443: Now that v13 has on-memory ORM, would it be a good moment to get rid of these "prepare" methods? They are not so helpful for performance anymore...

contract/models/abstract_contract_line.py
```diff
@@ -47,46 +48,32 @@ class ContractAbstractContractLine(models.AbstractModel):
...
+        readonly=False,
         required=True,
+        copy=True,
+    )
+    date_start = fields.Date(
+        compute="_compute_date_start", store=True, readonly=False, copy=True,
+    )
+    recurring_next_date = fields.Date(
+        compute="_compute_recurring_next_date", store=True, readonly=False, copy=True,
     )
```
yajo: If fields are computed and `readonly=False`, shouldn't they have an inverse method? :thinking:

contract/models/contract.py
```diff
@@ -43,6 +44,12 @@ class ContractContract(models.Model):
         inverse_name="contract_id",
         copy=True,
     )
+    # Trick for being able to have 2 different views for the same o2m
+    contract_line_fixed_ids = fields.One2many(
+        string="Contract lines (fixed)",
+        comodel_name="contract.line",
+        inverse_name="contract_id",
+    )
```
yajo: Shouldn't this have a `domain`?

yajo (commented):
I'm not very aware of the businsess logic behind the changes, but here are some general comments over the code.

Good job overall! :blush:

contract/models/abstract_contract.py
```diff
@@ -0,0 +1,72 @@
...
+        )
+
+    @api.model
+    def _default_journal(self):
+        company_id = self.env.context.get("company_id", self.env.user.company_id.id)
+        domain = [
+            ("type", "=", self.contract_type),
+            ("company_id", "=", company_id),
+        ]
+        return self.env["account.journal"].search(domain, limit=1)
```
em230418: When creating contract using `create` with given "contract_type" value, self.contract_type is false. So default journal is not set.
Anyway it seems like Odoo's bug. At least Odoo 13.0-20200417

flachica (approved):

CarlosRoca13 (approved):
LGTM :+1:

contract/models/abstract_contract.py
```diff
@@ -0,0 +1,72 @@
...
+        domain="[('type', '=', contract_type)," "('company_id', '=', company_id)]",
+        index=True,
+    )
+    company_id = fields.Many2one(
+        "res.company",
+        string="Company",
+        required=True,
+        default=lambda self: self.env.company.id,
+    )
+    line_recurrency = fields.Boolean(
```
sbejaoui: IMO, it's a company strategical choice more than a user decision on individual contract. I was imagining it as a company config.

sbejaoui: After the first invoice, only setting this option to true is possible.

The other case, we need to check that all lines have the same recurrence values (if not, ask the user to handle it manually before).

contract/models/abstract_contract.py
sbejaoui: `default_contract_type` is set in context from the menu action. `self.contract_type` is not false if the contract is created from the UI.

Need to be explicitly added to context if the contract creation is scripted.

contract/views/contract.xml
```diff
@@ -0,0 +1,653 @@
...
+                                    <field
+                                        name="is_plan_successor_allowed"
+                                        invisible="1"
+                                    />
+                                    <field name="is_stop_allowed" invisible="1" />
+                                    <field name="is_cancel_allowed" invisible="1" />
+                                    <field name="is_un_cancel_allowed" invisible="1" />
+                                    <field name="is_canceled" invisible="1" />
+                                    <button
+                                        name="action_plan_successor"
```
sbejaoui, line 253: those actions must be hidden in case `line_recurrency=False`.

We need to have the same actions in contract level

contract/views/contract.xml
```diff
@@ -0,0 +1,653 @@
...
+                                name="recurring_invoicing_type"
+                                attrs="{'required': [('line_recurrency', '=', False)]}"
+                            />
+                        </group>
+                        <group attrs="{'invisible': [('line_recurrency', '=', True)]}">
+                            <field
+                                name="date_start"
+                                attrs="{'required': [('line_recurrency', '=', False)]}"
+                            />
+                            <field name="date_end" />
```
sbejaoui, line 173: - `date_end` must be editable
- `recurring_next_date` must be computed after recurrence change

sbejaoui:
I get compute error after unchecking `line_recurrency`

sbejaoui (changes requested):

contract/models/abstract_contract.py
sbejaoui: default_contract_type is set in context from the menu action. self.contract_type is not false if the contract is created from the UI.

Need to be explicitly added to context if the contract creation is scripted.

NachoAlesLopez (approved):
LGTM, with the caveats that the others commented. I've left a minor remark about the code used to create invoices, but it's non blocking.

contract/models/contract.py
```diff
@@ -0,0 +1,516 @@
...
+            )
+        if not journal:
+            raise ValidationError(
+                _("Please define a %s journal for the company '%s'.")
+                % (self.contract_type, self.company_id.name or "")
+            )
+        invoice_type = "out_invoice"
+        if self.contract_type == "purchase":
+            invoice_type = "in_invoice"
+        move_form = Form(
```
NachoAlesLopez, line 322: Is it really needed to use `odoo.tests.Form` in this situation? Shouldn't initializing a dictionary be more efficient?

pedrobaeza: Yes, as this is not something about performance, but reliability: it tries to mimic how you input an invoice, for not having side effects of other modules that add some fields over invoices not being populated. Example: the payment mode added by `account_payment_partner` module. Not doing this forces us to make tricks like:

https://github.com/OCA/bank-payment/blob/12.0/account_payment_partner/models/account_invoice.py#L68-L84

Ideally, this wouldn't be needed, but with current ORM, this is the only way to ensure correctness. We can patch all possible extension modules, but that will be even worst. In a near future, when computed writable fields are totally implemented in core modules, we won't need this.

contract/models/contract.py
pedrobaeza: It's not exactly a migration only field, but a caveat we have from prior implementation (pre-v12): the only link between the contract and the invoice was at header level. Now it's contract line to invoice line. As you can't deduce each link for lines before this new implementation (invoices can be manipulated after the contract generation, etc), the commitment solution was to keep this field "forever" and search through it. For >=v12, this field will never be filled, and for <v12, it will contain data and you can't remove the field if you want that the smart-button continues showing the same number of invoices.

I have added a comment in this line clarifying its usage.

contract/models/contract.py
pedrobaeza: I'm afraid it's not possible to remove this hook methods if you need to expand this module, due to 3 things:

- Creation is performed split from writes, so you can only optimize second writing and subsequent ones, but the performance will suffer a bit.
- You still need a hook method in which performing that write, and the general one that creates all invoices is not the ideal one.
- And most important, you lose the context of which things is generating which other things, so the hooks are still very valuable for keeping clean code and promote encapsulation.

contract/models/abstract_contract_line.py
pedrobaeza: No, this is the new API feature. What we have been calling "computed writable fields". They are like normal fields, but react to several depends and recompute its value when that dependencies are changed.

contract/models/contract.py
pedrobaeza: No, as both views must contain the same number of lines. It's only for providing 2 different views of the same records (one with all the fields, and the other with simplified fields), which is not possible in regular one2many widget (that can't be put 2 times with different views). The main question here that forbids me to put `attrs` in fields for controlling visibility is that one of the tree must be editable inline, and the other not, so that can't be done through `attrs`. I'm expanding the comment for noting this.

contract/models/abstract_contract.py
pedrobaeza: I'm replacing current code with a computed writable field that should get rid of this problem. Please check when pushed.

contract/models/abstract_contract.py
pedrobaeza: I don't think so, as you can handle both type of contracts, or change them across time. Example:

- You start with a contract with 2 lines invoiced yearly.
- You create the contract without checking the line recurrency, and create easily with popups those lines.
- After 3 years, you need to add a monthly service. Then you check line, and define that third line through the popup, having the other two with previous recurrency data.

For company default configuration, just define a global default value for this field.

pedrobaeza: They should have, as computed writable fields are for that. In fact, the contract lines to invoice code is not changed. We are only playing with assignations of these values from parent to children. Have you found any case where this propagation hasn't happened?

contract/views/contract.xml
pedrobaeza: I let some actions like stopping a line, as you can do this even if the recurrence is activated. This one was also preserved because although the misleading name, it can be used for resuming a paused line, which is also a possibility with global recurrency.

pedrobaeza:
Pushed a new version with the commented changes, plus the cherry-pick of #540 that was missing + a renaming from `line_recurrency` to `line_recurrence`, as the last one is the most common ones (and now it's the moment to do these things).

On monday I will continue seeing the problems found by @sbejaoui and hopefully we will be able to merge this then.

contract/models/abstract_contract.py
sbejaoui: What I mean, after the the first invoice, contract line may have different next invoice date and different next invoicing period. If the user un-check `line_recurrency` in the contract, the compute method will override the recurrence values.

contract/views/contract.xml
pedrobaeza: Fixed with latest changes that cleans a bit the definition across inheritance and make more use of computed.

contract/models/abstract_contract.py
pedrobaeza: I think that we shouldn't avoid that situations, as if someone changes the line recurrence flag, it's their duty to check current state, and if a forbidden situation is got, then the constraint will arise.

pedrobaeza:
This should be all for fixing remaining issues. Please let me know anything else.

rescola:
Hi! I don't understand the change from contract_id to old_contract_id on account.move. I think it was ok the possibility to link contract from invoice (Generated by contract). Only in one click you'll get there. 
It's for something special this change?

pedrobaeza:
@novis82 this is not something of 13.0, but the refactoring on 12.0. The link is at line level (contract line vs invoice line), so for finding the invoices generated from a contract, you just need to find the invoices lines linked to the contract lines, and traverse the invoices to which that lines belongs to. In fact, there's a method that does that for you at contract.contract level.

rescola:
> @novis82 this is not something of 13.0, but the refactoring on 12.0. The link is at line level (contract line vs invoice line), so for finding the invoices generated from a contract, you just need to find the invoices lines linked to the contract lines, and traverse the invoices to which that lines belongs to. In fact, there's a method that does that for you at contract.contract level.

Thanks @pedrobaeza. I found it.

pedrobaeza:
@sbejaoui can you give your final blessing?

rescola:
Hi again. In a contract starting for example on 1/1/20 and ending 31/12/20. Then you are creating bills until the last invoice that can generate, recurring_next_date is doing well increasing. But when you reach the last bill it turns the first period instead of "False" like in recurrence at line level. Is it ok? I think it will be False because there's not next date to bill, isn't it?

contract/views/contract.xml
sbejaoui: the issue here, is after un-canceling a line or a resume after a stop, the system ask for the next invoice date. The date selected by the user is ignored. This is not user-friendly.




I propose to hide all actions in line level if the global mode is selected (only keep end-date editable) and add to the roadmap that we need to adpat those actions to the new mode

sbejaoui:
Can we add auto-renewal process to the contract too

contract/views/contract.xml
pedrobaeza: And what about hiding that field in case of contract recurrence?

pedrobaeza: I still think that it's useful to allow pause/resume even when contract recurrence.

sbejaoui:
bug:

If I add a new line after generating some invoices, the contract next invoice date will be reset and the invoicing will be blocked.

the contract next invoice must consider the earlier last_date_invoiced of all lines.

contract/views/contract.xml
sbejaoui: it's useful yes, but we need to adapt views to not ask for next invoice date if we are in global mode

pedrobaeza: OK, I'll go that way then, and check the rest of the comments.

rescola:
Captura de pantalla de 2020-09-04 13-30-12

Also I tried a contract starting 1/1/20 ending 31/12/20. I create all 12 invoices. Then I create new line with a single condition with invoice date to be generated on 1/10/20. It doesn't show the button to generate the bill, but it show correctly that the next date of next invoice is 1/10/20.

rousseldenis:
> bug:
> 
> If I add a new line after generating some invoices, the contract next invoice date will be reset and the invoicing will be blocked.
> 
> the contract next invoice must consider the earlier last_date_invoiced of all lines.
> 
> 

Indeed, I can confirm that.

AntoniRomera:
> > bug:
> > If I add a new line after generating some invoices, the contract next invoice date will be reset and the invoicing will be blocked.
> > the contract next invoice must consider the earlier last_date_invoiced of all lines.
> > 
> 
> Indeed, I can confirm that.

Right now I don't see this error, is it fixed? @pedrobaeza

pedrobaeza:
No, I have done yet anything. I'm checking today.

pedrobaeza:
@sbejaoui I'm not able to reproduce the problem as well. Can you provide details steps to reproduce it (that I will reflect in a test)?

pedrobaeza:
@rescola your case is incorrect, as you have as date end the same as date start, and thus, no period to invoice, and has nothing to do with the migration/refactoring itself.

sbejaoui:
> @sbejaoui I'm not able to reproduce the problem as well. Can you provide details steps to reproduce it (that I will reflect in a test)?

1. Add a contract line
2. generate some invoices
3. add a new contract line (notice that the contract next invoice will be reset to the new line date)


ezgif com-video-to-gif

rousseldenis:
> > @sbejaoui I'm not able to reproduce the problem as well. Can you provide details steps to reproduce it (that I will reflect in a test)?
> 
> 1. Add a contract line
> 2. generate some invoices
> 3. add a new contract line (notice that the contract next invoice will be reset to the new line date)
> 
> ezgif com-video-to-gif

I can reproduce. The thing I can say at this time is that that happens when:

- More than one contract line
- the 'recurrence at line level' is not checked

rousseldenis:
When opening contract line, product is required but not required on contract form level:

image

rousseldenis:
It seems that when constrains() is evaluated just after invoice creation, the recurring_next_date is inconsistent between lines recordset. One is at the good next month date, one is at former date (e.g.: one line at 01/06/2020 and one at 01/07/2020 - european date format).

Maybe an orm problem or...

sbejaoui:
The `recurring_next_date` compute in contract level consider the the oldest date from contract line. This make sense in case of 'recurrence at line level' is checked as we want that the system invoice the oldest line before.

In case of 'recurrence at contract level', the `recurring_next_date` must consider the most recent date to be able to invoice all lines at once.

rousseldenis:
@pedrobaeza Nothing new under the sun ?

pedrobaeza:
Sorry, very busy... I will check it soon

victoralmau:
> When opening contract line, product is required but not required on contract form level:
> 
> image

IMO (after check it) product is only required if you check "_Recurrence at line level?_"

victoralmau:
> > > @sbejaoui I'm not able to reproduce the problem as well. Can you provide details steps to reproduce it (that I will reflect in a test)?
> > 
> > 
> > 
> > 1. Add a contract line
> > 2. generate some invoices
> > 3. add a new contract line (notice that the contract next invoice will be reset to the new line date)
> > 
> > ezgif com-video-to-gif
> 
> I can reproduce. The thing I can say at this time is that that happens when:
> 
>     * More than one contract line
> 
>     * the 'recurrence at line level' is not checked



> > > @sbejaoui I'm not able to reproduce the problem as well. Can you provide details steps to reproduce it (that I will reflect in a test)?
> > 
> > 
> > 
> > 1. Add a contract line
> > 2. generate some invoices
> > 3. add a new contract line (notice that the contract next invoice will be reset to the new line date)
> > 
> > ezgif com-video-to-gif
> 
> I can reproduce. The thing I can say at this time is that that happens when:
> 
>     * More than one contract line
> 
>     * the 'recurrence at line level' is not checked

I can reproduce it (or similar).

- Create a contract (with 1 line)
- Create invoice/s (with button "Create Invoice").
- Edit contract and add other line
In save process apear error related to constrain (_check_last_date_invoiced)

I will investigate about it and suggeste a correction about it.

Not necessary to define product in lines (if don't check `Recurrence at line level?`) to do it.

contract/models/contract_line.py
```diff
@@ -0,0 +1,1040 @@
...
+                rec.recurring_next_date
+                and rec.recurring_next_date <= rec.last_date_invoiced
+            ):
+                raise ValidationError(
+                    _(
+                        "You can't have the next invoice date before the date "
+                        "of last invoice for the contract line '%s'"
+                    )
+                    % rec.name
+                )
```
victoralmau:
```suggestion
      for rec in self.filtered("last_date_invoiced"):
            if rec.date_end and rec.date_end < rec.last_date_invoiced:
                raise ValidationError(
                    _(
                        "You can't have the end date before the date of last "
                        "invoice for the contract line '%s'"
                    )
                    % rec.name
                )

            if rec.contract_id.line_recurrence:
                if rec.date_start and rec.date_start > rec.last_date_invoiced:
                    raise ValidationError(
                        _(
                            "You can't have the start date after the date of last "
                            "invoice for the contract line '%s'"
                        )
                        % rec.name
                    )
                if (
                    rec.recurring_next_date
                    and rec.recurring_next_date <= rec.last_date_invoiced
                ):
                    raise ValidationError(
                        _(
                            "You can't have the next invoice date before the date "
                            "of last invoice for the contract line '%s'"
                        )
                        % rec.name
                    )
```

With this suggestion i have try to solved a problem that exists (and other users comment about it).

**Problem**:
- Create contrat (`Recurrence at line level?` = False)
- Add a line
- Create a invoice(s)
- Edit, add new line and try to Save > Error
- Try to Create invoices > Error

**Solution**
This _check_constrain_ appear ever because when create new invoices and update info about contract_lines, in a little moment appear `last_date_invoiced` > `recurring_next_date`, because "recurring_next_date" re-calculation not apply yet.
With this suggestion IMO is not neccessary to check "impossible things" because ((`Recurrence at line level?` = False) `recurring_next_date` field is readonly (and auto-change when is need)

Probably exists more solutions about it but IMO it's the most easy.

What do you think @pedrobaeza ?

rousseldenis:
@victoralmau Did you see this ? https://github.com/Tecnativa/contract/pull/3/

victoralmau:
> @victoralmau Did you see this ? [Tecnativa#3](https://github.com/Tecnativa/contract/pull/3)

Thanks @rousseldenis . I haven't see it until now becasue we don't check PRs in our repositories almost never.

I have check it and wouldn't solve problem exist explained previously.

rousseldenis:
> > @victoralmau Did you see this ? [Tecnativa#3](https://github.com/Tecnativa/contract/pull/3)
> 
> Thanks @rousseldenis . I haven't see it until now becasue we don't check PRs in our repositories almost never.
> 
> I have check it and wouldn't solve problem exist explained previously.

No ? My test is showing it

victoralmau:
@pedrobaeza I have check this changes https://github.com/Tecnativa/contract/pull/3 and it's necessary to add it too (explained problem in PR). 
Thanks to @rousseldenis

pedrobaeza:
I have applied @victoralmau's suggestion to avoid the transient problem while adding a line, but with a friendlier for diff version.

@sbejaoui can you please recheck for removing the requested changes and finally merge this?

pedrobaeza:
@sbejaoui please, we have several PRs depending on this one. Can you check if it's OK for you?

sbejaoui:
The issue is fixed but there is a new one, you can reproduce it following this steps:

- create a contract, Recurrence at line level? = False
- add product A and add to the description `#START# ---> #END# ` to be able to follow the invoicing periods
- invoice the contract 3 times
- add product B and add to the description `#START# ---> #END# ` to be able to follow the invoicing periods
- invoice the contract 2 times

Expected behavior: 5 invoices

1. invoice 1: period 1: product A
2. invoice 2: period 2: product A
3. invoice 3: period 3: product A
4. invoice 4: period 1: product B
5. invoice 5: period 2: product B

Current behavior: 5 invoices

1. invoice 1: period 1: product A
2. invoice 2: period 2: product A
3. invoice 3: period 3: product A
4. invoice 4: period 1: product B
5. invoice 5: Both products are invoiced with mixed periods



you can check the contract in the review db [link](http://3445076-533-1fb4be.runbot2.odoo-community.org/web?debug=1#id=3&action=279&model=contract.contract&view_type=form&cids=1&menu_id=98)

contract/views/contract.xml
```diff
@@ -0,0 +1,653 @@
...
+                        name="action_contract_send"
+                        type="object"
+                        string="Send by Email"
+                        attrs="{'invisible': [('is_terminated','=',True)]}"
+                        groups="base.group_user"
+                    />
+                    <button
+                        name="recurring_create_invoice"
+                        type="object"
+                        attrs="{'invisible': ['|', ('create_invoice_visibility', '=', False)]}"
```
hparfr:
```suggestion
                        attrs="{'invisible': [('create_invoice_visibility', '=', False)]}"
```

pedrobaeza:
Hi @sbejaoui thanks for the extensive checking. Fixed your comment with latest push: https://github.com/OCA/contract/compare/1fb4be6ef2a6a23c3a861715af2ecb8e899b679e..f7f03f4f0e64c8eb863959b3eec180bebe9b8554 .

contract/views/contract.xml
pedrobaeza: Fixed as well.

sbejaoui (approved):
LGTM, Thank you

pedrobaeza:
Thanks to all the reviewers and contributors. Let's merge it and continue improving this great module through subsequent PRs:

/ocabot merge nobump

rousseldenis:
@pedrobaeza Just one remark, this commit https://github.com/OCA/contract/pull/533/commits/f7f03f4f0e64c8eb863959b3eec180bebe9b8554

has its own meaning, so why having added view modification that has nothing linked to...

pedrobaeza:
Sorry, it was that I amended commit and yours is the last. Aborting and fixing that commit history.

pedrobaeza:
Fixed:

/ocabot merge nobump

rousseldenis:
@pedrobaeza It seems that another test does not pass now.

pedrobaeza:
I think that is a test not resilient with timezones and midnight, but I haven't touched that test. Let me try again:

/ocabot merge nobump

pedrobaeza:
It turns out that the test started to fail now in my local as well... Not sure the reason, but clearly the failing constraint shouldn't apply to line sections. Let's see if this is the definitive:

/ocabot merge nobump

PR #535, opened 2020-07-27
[FIX] contract: multi-company assignment notification

yajo:
Steps to reproduce the problem:

1. Log in as Mitchell Admin.
2. Create contract CNT-A for company CMP-A, assigned to Marc Demo.
3. Create contract CNT-B for company CMP-B, assigned to Marc Demo.
4. Run cron to create recurring invoices.

Actual results:

- Odoo sends automated assignment emails to Marc Demo, which indicate the name of the company activated for `__system__` user whlie the cron was being executed.

Expected results after this patch:

- Odoo sends automated assignment emails to Marc Demo, which indicate the invoice company.

@Tecnativa TT24657

### DIFF
--- a/contract/models/contract.py
+++ b/contract/models/contract.py
@@ -546,8 +546,15 @@ class ContractContract(models.Model):
         if not date_ref:
             date_ref = fields.Date.context_today(self)
         domain = self._get_contracts_to_invoice_domain(date_ref)
-        contracts_to_invoice = self.search(domain)
-        return contracts_to_invoice._recurring_create_invoice(date_ref)
+        invoices = self.env["account.invoice"]
+        # Invoice by companies, so assignation emails get correct context
+        companies_to_invoice = self.read_group(domain, ["company_id"], ["company_id"])
+        for row in companies_to_invoice:
+            contracts_to_invoice = self.search(row["__domain"]).with_context(
+                allowed_company_ids=[row["company_id"][0]]
+            )
+            invoices |= contracts_to_invoice._recurring_create_invoice(date_ref)
+        return invoices
 
     @api.multi
     def action_terminate_contract(self):
--- a/contract/README.rst
+++ b/contract/README.rst
@@ -74,6 +74,7 @@ Known issues / Roadmap
 ======================
 
 * Recover states and others functional fields in Contracts.
+* Remove ``models/ir_ui_view.py`` in v13, where the workaround it contains is supported upstream.
 
 Bug Tracker
 ===========
--- a/contract/models/__init__.py
+++ b/contract/models/__init__.py
@@ -13,3 +13,4 @@ from . import contract_tag
 from . import res_company
 from . import res_config_settings
 from . import contract_terminate_reason
+from . import ir_ui_view
--- a/contract/models/ir_ui_view.py
+++ b/contract/models/ir_ui_view.py
@@ -1,0 +1,18 @@
+# Copyright 2020 Tecnativa - Jairo Llopis
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import api, models
+
+
+class IrUiView(models.Model):
+    _inherit = "ir.ui.view"
+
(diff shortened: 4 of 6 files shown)


pedrobaeza:
I don't really like to mangle with the user's company, as there can be other operations in parallel that are breaking due to this change. Have you checked how that assignation notification is done instead? I prefer to modify that code to take document's company instead.

yajo:
I don't think that is really going to happen because there's no DB commit between the company changes, so theoretically the company is never going to change in other cursors. I don't know if odoo will end up issuing an `UPDATE` query, but in any case it's going to update to the same value, so it shouldn't clash.

Before you ask, the notification mails are queued, not sent directly, so they don't produce a DB commit or race condition either.

> Have you checked how that assignation notification is done instead?

Considering what I mentioned above, I didn't see this necessary; but yes, I have checked this option and it would mean to patch [this line](https://github.com/odoo/odoo/blob/bfd2b669c056cc9fa6e97992944e66aac7c1092a/odoo/addons/base/models/ir_ui_view.py#L1343).

Do you still think the change is needed/benefitial? :thinking: 

---

BTW neither of these will be needed in v13, where all can be fixed with a context key.

pedrobaeza:
Well, we don't know what overrides are in the system, and modifying wildly that company for whole process seems to risky for me, and more knowing that this is not a problem anymore in v13, so I would write this in the ROADMAP of v12 and notify users that it won't be a problem there (but are you sure the link to Odoo reference is changed in v13 to take document's company?). I would prefer the patch in Odoo to take there the document's company.

yajo:
> are you sure the link to Odoo reference is changed in v13 to take document's company?

Nope, it takes the company from the context. We will still need the workaround, it will just be simpler.

> I would prefer the patch in Odoo to take there the document's company.

That is a big patch, and it doesn't affect upstream code, so I see too many possibilities that Odoo rejects it.

Let me just write the alternative workaround here and add the roadmap note.

yajo:
Patch ready.

pedrobaeza (approved):
OK, it can work in a less risky way.

contract/models/contract.py
```diff
@@ -546,8 +546,13 @@ def cron_recurring_create_invoice(self, date_ref=None):
...
         domain = self._get_contracts_to_invoice_domain(date_ref)
-        contracts_to_invoice = self.search(domain)
-        return contracts_to_invoice._recurring_create_invoice(date_ref)
+        # Invoice by companies, so assignation emails get correct context
+        companies_to_invoice = self.read_group(domain, ["company_id"], ["company_id"])
+        for row in companies_to_invoice:
+            contracts_to_invoice = self.search(row["__domain"]).with_context(
+                allowed_company_ids=[row["company_id"][0]]
+            )
+            contracts_to_invoice._recurring_create_invoice(date_ref)
```
sbejaoui: please keep the return

sbejaoui (changes requested):
We didn't notice this case as we configure one cron per company with a regular user.

contract/models/contract.py
pedrobaeza: invoices records should be then stacked and return at the end.

sbejaoui: yes that what I mean :)

yajo: I saw it's used nowhere, but OK no problem on this.

yajo:
All fixed.

sbejaoui (approved):
thanks @Yajo

pedrobaeza:
Included this in v13 in latest push:

https://github.com/OCA/contract/compare/21ff7d808df330a728fe56dff401a6da4ee65a77..647a4417466e6264fc1c36976992735d7ec247b2

yajo:
Mm? What do you mean? The addon isn't in v13 yet AFAICS.

pedrobaeza:
I added to the migration PR: https://github.com/OCA/contract/pull/533

You can review it by chance :angel:

PR #536, opened 2020-07-28
[FIX] contract: return qcontext result

yajo:
Duh... 🤓 Obvious regression from #535

@Tecnativa TT24657


chienandalu (approved):

pedrobaeza:
Fast-tracking this for fixing the bug:

/ocabot merge patch

PR #539, opened 2020-08-07
[FIX][12.0]product_contract: bad date_end regarding the contract renew strategy

flotho:
The product contract has options for setting the auto renw strategy.
The strategy was incoherent regarding the contract line strategy :
* Date_end is mandatory if auto renew is set on contract line
* date_end is always mandatory on sale.order.line whatever the reneww option is set

The purpose is to have the same behaviour in both modules


flotho:
ping @sbejaoui , @pedrobaeza

PR #541, opened 2020-08-24
[12.0][IMP] - Different payment mode for contracts generated from sale orders

sbejaoui:
New option allow to set different payment mode for the contracts generated from the sale order than the one will be used for invoices.

### DIFF
--- a/contract_sale_payment_mode/models/res_company.py
+++ b/contract_sale_payment_mode/models/res_company.py
@@ -1,0 +1,13 @@
+# Copyright 2020 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+
+    _inherit = 'res.company'
+
+    specific_contract_payment_mode = fields.Boolean(
+        string="Specific payment mode for contracts created from sale orders"
+    )
--- a/contract_sale_payment_mode/models/sale_order.py
+++ b/contract_sale_payment_mode/models/sale_order.py
@@ -1,16 +1,43 @@
 # Copyright 2019 ACSONE SA/NV
+# Copyright 2020 ACSONE SA/NV
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo import api, models
+from odoo import api, fields, models
 
 
 class SaleOrder(models.Model):
     _inherit = 'sale.order'
+
+    contract_payment_mode_id = fields.Many2one(
+        comodel_name='account.payment.mode',
+        string='Contract Payment Mode',
+        domain=[('payment_type', '=', 'inbound')],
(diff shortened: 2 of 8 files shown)


gva-acsone (approved):
LGTM Functional test

contract_sale_payment_mode/models/sale_order.py
```diff
@@ -1,16 +1,43 @@
...
     @api.multi
     def _prepare_contract_value(self, contract_template):
         self.ensure_one()
         vals = super(SaleOrder, self)._prepare_contract_value(
             contract_template)
-        vals['payment_mode_id'] = self.payment_mode_id.id
+        if self.specific_contract_payment_mode:
+            vals['payment_mode_id'] = self.contract_payment_mode_id.id
+        else:
+            vals['payment_mode_id'] = self.payment_mode_id.id
```
nikul-serpentcs, line 42: @sbejaoui your Code is not wrong but IMO use like this
`vals.update({'payment_mode_id': self.contract_payment_mode_id.id if self.specific_contract_payment_mode else self.payment_mode_id.id})`

sbejaoui:
cc/ @len-foss

contract_sale_payment_mode/models/res_company.py
```diff
@@ -0,0 +1,14 @@
...
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+
+    _inherit = 'res.company'
+
+    specific_contract_payment_mode = fields.Boolean(
+        string="Different payment mode for contracts generated from sale "
+        "orders"
```
len-foss: Oh come on you can't leave that word orphaned like this.
Besides that, maybe add a bit more verbose help message? 
(that also applies below for the other fields).

contract_sale_payment_mode/models/sale_order.py
len-foss: Outside of Odoo, it's generally frowned upon to have 300+ character long lines of code.

contract_sale_payment_mode/models/sale_order.py
```diff
@@ -1,16 +1,43 @@
...
 class SaleOrder(models.Model):
     _inherit = 'sale.order'
 
+    contract_payment_mode_id = fields.Many2one(
+        comodel_name='account.payment.mode',
+        string='Contract Payment Mode',
+        domain=[('payment_type', '=', 'inbound')],
+        index=True,
+    )
+    specific_contract_payment_mode = fields.Boolean(
```
len-foss, line 17: I don't think adding this field is necessary?

len-foss (approved):
LGTM with a few minor comments.

contract_sale_payment_mode/models/sale_order.py
sbejaoui: it's used for the new field visibility in SO form view

PR #543, opened 2020-08-25
[IMP] agreement_serviceprofile (Copy True)

patrickrwilson:
Service profiles should be copied when an agreement is created from a template or when duplicating from another agreement. This adds copy=True to the field.


patrickrwilson:
ping @max3903 @osi-scampbell

max3903 (approved):

bodedra (approved):

osi-scampbell (approved):

PR #544, opened 2020-08-26
[12.0][IMP] - use next period computed fields to init forecast

sbejaoui:
after the add of computed fields for next invoicing period, it's more efficient to use them to init the forecast generation.

### DIFF
--- a/contract_forecast/models/contract_line.py
+++ b/contract_forecast/models/contract_line.py
@@ -58,8 +58,8 @@ class ContractLine(models.Model):
         if not self.date_end or self.is_auto_renew:
             return period_date_end < contract_forecast_end_date
         return (
-            period_date_end < self.date_end
-            and period_date_end < contract_forecast_end_date
+            period_date_end <= self.date_end
+            and period_date_end <= contract_forecast_end_date
         )
 
     @api.multi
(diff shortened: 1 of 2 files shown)


bodedra (approved):
LG :+1:

sbejaoui:
@gva-acsone

sbejaoui:
cc/ @len-foss

contract_forecast/models/contract_line.py
```diff
@@ -69,18 +69,24 @@ def _generate_forecast_periods(self):
...
                 max_date_end = rec.date_end if not rec.is_auto_renew else False
                 while (
                     period_date_end
                     and rec._get_generate_forecast_periods_criteria(
                         period_date_end
                     )
                 ):
+                    if period_date_end and recurring_next_date:
+                        values.append(
+                            rec._prepare_contract_line_forecast_period(
```
len-foss: In general it's a bit awkward to have indented nested call like this compared to 
```
new_value = rec._prepare_contract_line_forecast_period(...)
values.append(new_value)
```
(and you win one line). But it's really a trivial detail.

len-foss (approved):
LGTM (is it really an IMP and not a REF?)

marcelsavegnago (approved):
LGTM

PR #558, opened 2020-10-21
[13.0][MIG] contract_payment_mode: Migration to 13.0

willer-mo:
Migration to 13.0

* [x] Depends on https://github.com/OCA/contract/pull/533


sbejaoui:
Please rebase, #533 was merged

contract_payment_mode/models/contract.py
```diff
@@ -0,0 +1,27 @@
...
+
+    def _prepare_invoice(self, date_invoice, journal=None):
+        res = super(ContractContract, self)._prepare_invoice(
+            date_invoice=date_invoice, journal=journal
+        )
+        invoice_vals = res[0]
+        if self.payment_mode_id:
+            invoice_vals["payment_mode_id"] = self.payment_mode_id.id
+            invoice = self.env["account.move"].new(invoice_vals)
+            invoice_vals = invoice._convert_to_write(invoice._cache)
```
sbejaoui: @pedrobaeza come with the idea to use Form test class to trigger the onchanges. Can you do the same here.

Please check

https://github.com/OCA/contract/blob/6177f9de43a7acd26bca05e7dc905f7c9622f07a/contract/models/contract.py#L322-L344

sbejaoui (dismissed):

contract_payment_mode/models/contract.py
pedrobaeza: Yes, the idea is to reuse `move_form` variable if needed to perform changes with onchanges, or directly inject in `invoice_vals` any value in the dictionary. That way, you have the flexibility of using any of this mechanism, but you should respect at least the expected return value.

willer-mo: Although onchanges won't be executed this way, I have done it injecting it in `invoice_vals`.

contract_payment_mode/models/contract.py
```diff
@@ -15,28 +15,11 @@ class ContractContract(models.Model):
     def on_change_partner_id(self):
         self.payment_mode_id = self.partner_id.customer_payment_mode_id.id
 
-    @api.multi
     def _prepare_invoice(self, date_invoice, journal=None):
-        invoice_vals = super(ContractContract, self)._prepare_invoice(
+        res = super(ContractContract, self)._prepare_invoice(
```
CarlosRoca13: This is not necessary in python3 when using the same class and self field
```suggestion
        res = super()._prepare_invoice(
```

contract_payment_mode/tests/test_contract_payment.py
```diff
@@ -15,7 +15,7 @@ def setUp(self):
         super(TestContractPaymentInit, self).setUp()
```
CarlosRoca13:
```suggestion
        super().setUp()
```

CarlosRoca13 (approved):
Just non blocking comments... I think that you will be a contributor of this module too. You can add youself to CONTRIBUTORS.rst and run the oca-gen-addon-readme

contract_payment_mode/models/contract.py
pedrobaeza: Please unfold directly the returned variables:
```suggestion
        invoice_vals, move_form = super()._prepare_invoice(
```
and then do:
```python
return invoice_vals, move_form
```

willer-mo: Ok, I have applied these changes :+1:

pedrobaeza (approved):

PR #567, opened 2020-10-27
[13.0] [MIG] contract_variable_quantity: Migration to v13.0

CarlosRoca13:
cc @Tecnativa TT20652

Please @pedrobaeza @ernestotejeda, can you review this?


sbejaoui (approved):
LGTM, (code review)

Thank you

contract_variable_quantity/models/contract.py
```diff
@@ -13,3 +13,13 @@ class ContractContract(models.Model):
         string="Skip Zero Qty Lines",
         help="If checked, contract lines with 0 qty don't create invoice line",
     )
+
+    def _prepare_recurring_invoices_values(self, date_ref=False):
+        vals = super()._prepare_recurring_invoices_values(date_ref=date_ref)
```
pedrobaeza: I don't get why are you doing this, but it seems very incorrect, as this module only operates in the quantity field of the invoice line, not on invoice header.

pedrobaeza (changes requested):

contract_variable_quantity/models/contract.py
CarlosRoca13: I do it because when `_prepare_invoice_line` return `vals` as `{}`. This method create a line on the contract, I do it for avoid this line creation. Based on test, I think that's the original scope...

CarlosRoca13:
@pedrobaeza Chage done!

pedrobaeza (approved):

PR #569, opened 2020-10-30
[12.0] Removing group_by if this is in context.

Lopsanz:
I think there's a trouble with tree-form-tree context  that is not being to be solved by Odoo.
So with this change I prevent error.

### DIFF
--- a/contract_sale_generation/__manifest__.py
+++ b/contract_sale_generation/__manifest__.py
@@ -5,7 +5,7 @@
 
 {
     'name': 'Contracts Management - Recurring Sales',
-    'version': '12.0.1.0.1',
+    'version': '12.0.1.0.2',
     'category': 'Contract Management',
     'license': 'AGPL-3',
     'author': "PESOL, "


pedrobaeza (approved):

contract/models/res_partner.py
```diff
@@ -46,14 +46,15 @@ def act_show_contract(self):
...
-        )
+        t_context = dict(
+            self.env.context,
+            search_default_partner_id=self.id,
+            default_partner_id=self.id,
+            default_pricelist_id=self.property_product_pricelist.id,
+            )
+        if 'group_by' in t_context:
+            del t_context['group_by']
+        res.update(context=t_context)
```
len-foss: You can gain a few lines:

```suggestion
action_context = {k: v for k, v in self.env.context if k != "groupby"}
action_context["search_default_partner_id"] = self.id
action_context["default_partner_id"] = self.id
action_context["default_pricelist_id"] = self.property_product_pricelist.id

res["context"] = action_context  
```
(or you could simply use `t_context.pop('group_by', None)`).
Anyway, LGTM.

len-foss (approved):

sbejaoui:

Lopsanz:
Me and git commands aren't finally friends. So... can anybody take these changes in his own PR??

PR #571, opened 2020-11-04
[13.0][FIX] contract: Only include 'section_and_note_one2many' js widget in view fields that contain it

victoralmau:
Now in contract list in runbot appear an error like this

contract-js-error

This error is related to include .js related to widget `section_and_note_one2many` in any places BUT it's only need to load when some fields that contain it show.

This error **NOT** appear in V12.

Please @pedrobaeza and @Tardo  review it

@Tecnativa TT26617

### DIFF
--- a/contract/static/src/js/section_and_note_fields_backend.js
+++ b/contract/static/src/js/section_and_note_fields_backend.js
@@ -11,6 +11,7 @@ this can be applied directly to Odoo*/
 odoo.define("contract.section_and_note_backend", function(require) {
     "use strict";
 
+    require("account.section_and_note_backend");
     var fieldRegistry = require("web.field_registry");
     var section_and_note_one2many = fieldRegistry.get("section_and_note_one2many");
 


contract/static/src/js/section_and_note_fields_backend.js
```diff
@@ -13,26 +13,27 @@ odoo.define("contract.section_and_note_backend", function(require) {
```
Tardo, line 13: I think that this issue happens because the script is loading before that the main widget script.

This change should fix it... with this you ensure that the main widget script is loaded:
```suggestion
require('account.section_and_note_backend');
```

Tardo (changes requested):

contract/static/src/js/section_and_note_fields_backend.js
victoralmau: Thanks, I have changed it with your suggestion, is it correct now?

pedrobaeza (approved):

Tardo (approved):

PR #573, opened 2020-11-06
[12.0][Fix] Recurring config missing on order line with product configurator

Martronic-SA:
The fields recurring_rule_type and recurring_invoicing_type are not present in the sale order form in lines tree view, so they are not set when creating a sales order line from the product configurator.
Here in the fix they are only added as invisible but maybe they could be shown in the line tree view.


len-foss (approved):
LGTM

sbejaoui:

sbejaoui (approved):

dreispt:
Travis needs fixing.

PR #574, opened 2020-11-10
[IMP] product_contract: manage contracts at variant level, enforce contract field

len-foss:
There are two improvements here:
(1) allow to manage contract template per variant. The one set on the template is the default for the variants.
(2) add a constraint to enforce a contract template for products that are contracts

The point (2) is managed through a (company) settings which defaults to False, as to not break existing installs.

### DIFF
--- a/product_contract/migrations/12.0.6.0.0/pre-migration.py
+++ b/product_contract/migrations/12.0.6.0.0/pre-migration.py
@@ -1,0 +1,18 @@
+# Copyright 2020 ACSONE SA/NV.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    """Copy the product template contract on their variants."""
+    contract_key = "property_contract_template_id"  # TODO: check that mig (do in SQL?)
+
+    for company in env["res.company"].with_context(active_test=False).search([]):
+        company_env = env["product.template"].with_context(
+            force_company=company.id, active_test=False
+        )
+        templates = company_env.search([(contract_key, "!=", False)])
+        for template in templates:
+            template.variant_ids.write({contract_key: template[contract_key].id})
--- a/product_contract/models/product_product.py
+++ b/product_contract/models/product_product.py
@@ -1,0 +1,56 @@
+# Copyright 2020 ACSONE SA/NV.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
+
+
+class ProductProduct(models.Model):
+    """Adds the possibility to override the template contract at variant level."""
+
+    _inherit = "product.product"
+
+    property_contract_template_id = fields.Many2one(
+        comodel_name="contract.template",
+        string="Contract Template",
+        company_dependent=True,
+    )
+
+    def _format_names_for_constraint_message(self, limit=None):
+        exceeds = limit and len(self) > limit
+        to_format = self[:limit] if exceeds else self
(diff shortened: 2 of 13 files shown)


len-foss:
@sbejaoui Hi! As a resident expert on this module, could you give your opinion on this? Thanks in advance!

sbejaoui:
Hi @len-foss ,

Thank you for this PR. I am completely ok to move all contract fields from the product template to the variant level. It's an improvement to cover the use-case where two variants have different recurrence options.

My first thoughts are:
- Yes we need a company option to decide whether this field is required or not.
- We may need a constraint to refuse `property_contract_template_id` if it's not a contract.

product_contract/models/product_product.py
```diff
@@ -0,0 +1,34 @@
+# Copyright 2017 LasLabs Inc.
```
sbejaoui, line 1:
```suggestion+1
# Copyright 2020 ACSONE SA/NV.
```

product_contract/models/product_product.py
```diff
@@ -0,0 +1,34 @@
...
+# Copyright 2018 ACSONE SA/NV.
```
sbejaoui:
```suggestion
# Copyright 2020 ACSONE SA/NV.
```

len-foss:
Thank you @sbejaoui 
I have implemented the constraints both ways, one way being dependent on the company settings (which I think is what you meant).
Diff is a tad larger than it should be because of pre-commit though.

Besides checking the migration, I think everything is essentially done here.

sbejaoui:
@len-foss , pre-commit and black were introduced in >= 13.0 oca branches. IMO, it's better to keep it that way and minimize the diff in this PR avoiding useless conflict for other open PRs.

len-foss:
It's true that there are a lot of open PRs, I'll go through everything to minimize the diff then.

product_contract/models/product_product.py
```diff
@@ -0,0 +1,56 @@
...
+
+    _inherit = "product.product"
+
+    property_contract_template_id = fields.Many2one(
+        comodel_name="contract.template",
+        string="Contract Template",
+        company_dependent=True,
+    )
+
+    def _format_names(self, limit=None):
```
sbejaoui:
```suggestion
    def _format_names_for_constraint_message(self, limit=None):
```
Just for code readability and to be confused with variant display name methods

sbejaoui (approved):
Looks good, thank you. Just a small suggestion

sbejaoui:

product_contract/migrations/12.0.6.0.0/pre-migration.py
```diff
@@ -0,0 +1,18 @@
...
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    """Copy the product template contract on their variants."""
+    contract_key = "property_contract_template_id"  # TODO: check that mig (do in SQL?)
+
+    for company in env["res.company"].with_context(active_test=False).search([]):
```
lmignon, line 12: IMO we should avoid the use of the ORM.... but  The query could be a little bit tricky to write for company dependent fields

lmignon (approved):
LGTM (code review)

len-foss:
I'll test a migration before merge.

len-foss:
Based on my current understanding, the approach is seriously flawed at best, unworkable at worst.
It was based on the understanding that "contract attributes" were on the contract template, while they are on the product themselves. 
While transferring all the contract fields at the variant is possible (it would make the proposed code _slightly more horrible_), it would not even completely align with our requirements. 
So I think this would be better suited in a deeper refactoring, and not suitable in its current state.

Some parts could be salvaged though (constraint checking) but it does not look necessary for now.

PR #575, opened 2020-11-13
[12.0][FIX] - Fix travis

sbejaoui:
Travis in 12.0 is red


pedrobaeza (approved):
AFAIK this is a warning, nothing blocking, but it's good to remove it. Please check better other possible errors.

len-foss (approved):
So this is the source of these red builds!

pedrobaeza:
This last one was! It's due to that check arriving 2 weeks ago or so in pylint-odoo. Let's merge it:

/ocabot merge patch

PR #583, opened 2020-11-30
[12.0][IMP] contract: Add followers to invoices created

victoralmau:
Add followers to invoices created (related to followers subtype "Invoice create" in contracts).

@Tecnativa TT27027

### DIFF
--- a/contract/data/mail_message_subtype.xml
+++ b/contract/data/mail_message_subtype.xml
@@ -1,0 +1,7 @@
+<?xml version="1.0" ?>
+<odoo noupdate="1">
+    <record id="mail_message_subtype_invoice_created" model="mail.message.subtype">
+        <field name="name">Invoice created</field>
+        <field name="res_model">contract.contract</field>
+    </record>
+</odoo>
--- a/contract/models/contract.py
+++ b/contract/models/contract.py
@@ -418,6 +418,18 @@ class ContractContract(models.Model):
         invoices.compute_taxes()
 
     @api.model
+    def _invoice_followers(self, invoices):
+        invoice_create_subtype = self.sudo().env.ref(
+            'contract.mail_message_subtype_invoice_created'
+        )
+        for item in self:
+            partner_ids = item.message_follower_ids.filtered(
+                lambda x: invoice_create_subtype in x.subtype_ids
+            ).mapped('partner_id')
+            if partner_ids:
+                invoices.message_subscribe(partner_ids=partner_ids.ids)
+
+    @api.model
     def _finalize_and_create_invoices(self, invoices_values):
         """This method:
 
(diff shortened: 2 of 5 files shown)


joao-p-marques (approved):
Code a functional review :+1: 
Good job!

contract/models/contract.py
```diff
@@ -417,6 +417,18 @@ def _finalize_invoice_creation(self, invoices):
         """
         invoices.compute_taxes()
 
+    @api.model
+    def _followers_invoice(self, invoices):
+        for item in self:
+            invoice_create_subtype = self.sudo().env.ref(
```
sbejaoui: this can be outside the for loop

sbejaoui (approved):

contract/data/mail_message_subtype.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" ?>
+<odoo noupdate="1">
+    <record id="mail_message_subtype_invoice_create" model="mail.message.subtype">
+        <field name="name">Invoice create</field>
```
pedrobaeza:
```suggestion
        <field name="name">Invoice created</field>
```

contract/models/contract.py
```diff
@@ -435,6 +447,7 @@ def _finalize_and_create_invoices(self, invoices_values):
             )
         invoices = self.env['account.invoice'].create(final_invoices_values)
         self._finalize_invoice_creation(invoices)
+        self._followers_invoice(invoices)
```
pedrobaeza: Better wording?
```suggestion
        self._invoice_followers(invoices)
```

pedrobaeza (changes requested):

pedrobaeza:
Can you check Travis status?

victoralmau:
> Can you check Travis status?

Sorry, solved now.

marcelsavegnago (approved):
LGTM

contract/data/mail_message_subtype.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" ?>
+<odoo noupdate="1">
+    <record id="mail_message_subtype_invoice_created" model="mail.message.subtype">
+        <field name="name">Invoice created</field>
+        <field name="res_model">contract.contract</field>
+    </record>
```
ernestotejeda, line 6: Why don't you specify here in the 'parent_id' field the existing sup-type 'Invoice created' (https://github.com/odoo/odoo/blob/14.0/addons/account/data/account_data.xml#L110)  so that each partner that has that subtype is automatically notified of the creation of an invoice from the contract ?

victoralmau: Although it's a good idea IMO isn't possible because need to related a contract field but in this case don't exists these field and would be one2many field.

ernestotejeda (approved):

pedrobaeza (approved):

PR #585, opened 2020-12-02
[12.0][IMP] contract: Add contract modifications tracking

victoralmau:
Add contract modification (portal + email + report).

@Tecnativa TT27029


pedrobaeza:
Please finish this

contract/models/contract.py
```diff
@@ -123,6 +123,41 @@ class ContractContract(models.Model):
...
+    modification_ids = fields.One2many(
+        comodel_name='contract.modification',
+        inverse_name='contract_id',
+        string='Modifications',
+    )
+
+    @api.multi
+    def write(self, vals):
+        res = super(ContractContract, self).write(vals)
+        if 'modification_ids' in vals:
```
pedrobaeza, line 139: You should do this instead on create/write of the subobject. If not, no notification will be sent in case of direct creation through code or specific UI. We have to take into account multi-creation on specific code for sending all the modification in one batch.

contract/models/contract.py
```diff
@@ -123,6 +123,41 @@ class ContractContract(models.Model):
         copy=False,
         track_visibility="onchange",
     )
+    modification_ids = fields.One2many(
+        comodel_name='contract.modification',
+        inverse_name='contract_id',
+        string='Modifications',
+    )
+
+    @api.multi
```
pedrobaeza: Don't add this in new code as it's not needed and forbidden in future versions.

contract/data/mail_template.xml
```diff
@@ -62,5 +62,26 @@
...
+        <field name="subject">${object.company_id.name} Contract (Ref ${object.name or 'n/a'})</field>
+        <field name="partner_to">${object.partner_id.id}</field>
+        <field name="model_id" ref="model_contract_contract"/>
+        <field name="report_template" ref="contract.report_contract"/>
+        <field name="report_name">Contract Modification</field>
+        <field name="lang">${object.partner_id.lang}</field>
+        <field name="body_html"><![CDATA[
+<div style="font-family: 'Lucida Grande', Ubuntu, Arial, Verdana, sans-serif; font-size: 12px; color: rgb(34, 34, 34); background-color: #FFF; ">
+    <p>Hello,</p>
+    <p>We have modifications on the contract <strong>${object.name}</strong></p>
```
pedrobaeza:
```suggestion
    <p>We have modifications on the contract <strong>${object.name}</strong> that we want to notify you.</p>
```

contract/models/contract.py
```diff
@@ -123,6 +123,41 @@ class ContractContract(models.Model):
...
+            lambda x: not x.mail_notify
+        )
+        if modification_ids_not_notify:
+            template = self.env.ref(
+                'contract.email_contract_modification_template', False
+            )
+            contract_modification_subtype = self.sudo().env.ref(
+                'contract.mail_message_subtype_contract_modification'
+            )
+            self.message_post_with_template(
```
pedrobaeza: Don't post messages if no followers.

contract/models/contract_modification.py
```diff
@@ -0,0 +1,31 @@
...
+        string='Description'
+    )
+    contract_id = fields.Many2one(
+        string='Contract',
+        comodel_name='contract.contract',
+        required=True,
+        ondelete='cascade',
+        index=True
+    )
+    mail_notify = fields.Boolean(
```
pedrobaeza: Be more precise in the field name. This may mean the contrary: that the notification should be done. Maybe you wanted to say `mail_notified`. Get use to the participle form of the verbs in these cases. Anyway, I think it's better to use `sent` as field name.

contract/security/ir.model.access.csv
```diff
@@ -9,3 +9,5 @@
 "contract_line_portal","Recurring portal","model_contract_line","base.group_portal",1,0,0,0
 "contract_template_line_manager","Recurring manager","model_contract_template_line","account.group_account_manager",1,1,1,1
 "contract_template_line_user","Recurring user","model_contract_template_line","account.group_account_invoice",1,0,0,0
+"contract_modification_user","Recurring user","model_contract_modification","account.group_account_invoice",1,1,1,1
```
pedrobaeza: The name is not very descriptive (although it has been used constantly before). It's better to put:
```suggestion
"contract_modification_user","Contract modifications - User","model_contract_modification","account.group_account_invoice",1,1,1,1
```

contract/models/contract.py
```diff
@@ -123,6 +123,41 @@ class ContractContract(models.Model):
...
+
+    @api.model
+    def _modification_mail_send(self):
+        self.ensure_one()
+        modification_ids_not_notify = self.modification_ids.filtered(
+            lambda x: not x.mail_notify
+        )
+        if modification_ids_not_notify:
+            template = self.env.ref(
+                'contract.email_contract_modification_template', False
```
pedrobaeza: You can remove the `False` for not silently fail on this.

contract/models/contract_modification.py
```diff
@@ -0,0 +1,31 @@
...
+
+    _name = 'contract.modification'
+    _description = 'Contract Modification'
+    _order = 'date asc'
+
+    date = fields.Date(
+        required=True,
+        string='Date'
+    )
+    description = fields.Text(
```
pedrobaeza, line 17: Doing this multi-line, will be ugly when rendering the mail template. Options:

1. Not being multi-line.
2. More elegant: render a table in the mail for displaying this.

contract/security/ir.model.access.csv
```diff
@@ -9,3 +9,5 @@
...
+"contract_modification_portal","Recurring portal","model_contract_modification","base.group_portal",1,0,0,0
```
pedrobaeza: Same for this:
```suggestion
"contract_modification_portal","Contract modifications - Portal","model_contract_modification","base.group_portal",1,0,0,0
```

contract/views/contract_portal_templates.xml
```diff
@@ -165,6 +165,28 @@
                                 </tbody>
                             </table>
                         </div>
+                        <div t-if="contract.modification_ids" id="item_modifications">
+                            <table class="table table-sm" id="sales_order_table">
```
pedrobaeza: Can this be shown as timeline?

contract/models/contract.py
pedrobaeza: I'm thinking of using QWeb template instead of this for several reasons:

- Being more hidden from modifications.
- Allow inheritance.

Anyway, the message should be put in the thread. There's an option for using QWeb from mail templates as an envelope, but maybe we can trick it for being the full template and empty inner body.

What do you think?

contract/tests/test_contract.py
```diff
@@ -133,6 +134,36 @@ def _add_template_line(self, overrides=None):
...
+                    0,
+                    0,
+                    {
+                        'date': '2020-02-01',
+                        'description': 'Modification 2',
+                    }
+                )
+            ]
+        })
+        self.assertEqual(len(self.contract.modification_ids.ids), 2)
```
pedrobaeza: This assert doesn't add value,  as it's testing the framework, not the feature. You can search for mail.message linked to the contract ID if you want to check for the sent mail.

contract/views/contract.xml
```diff
@@ -170,6 +170,14 @@
                             </field>
                             <field name="note"/>
                         </page>
+                        <page name="modications" string="Modifications">
+                            <field name="modification_ids" attrs="{'readonly': [('is_terminated','=',True)]}">
+                                <tree editable="bottom">
+                                    <field name="date"/>
+                                    <field name="description"/>
+                                </tree>
```
pedrobaeza, line 179: Show also the sent flag.

pedrobaeza (changes requested):

contract/models/contract.py
victoralmau: IMO QWeb template it's a good idea. I will change it now.

contract/models/contract_modification.py
victoralmau: Option 2 it's a good idea.

contract/views/contract_portal_templates.xml
victoralmau: Please, copy some little example about it.

https://github.com/odoo/odoo/blob/14.0/addons/website/views/snippets/s_timeline.xml

victoralmau:
I will add the modifications in the report (PDF).

contract/models/contract.py
```diff
@@ -123,6 +123,49 @@ class ContractContract(models.Model):
...
     )
+    modification_ids = fields.One2many(
+        comodel_name='contract.modification',
+        inverse_name='contract_id',
+        string='Modifications',
+    )
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        record = super(ContractContract, self).create(vals_list)
```
Tardo:
```suggestion
        record = super().create(vals_list)
```

contract/models/contract_modification.py
```diff
@@ -0,0 +1,47 @@
...
+        index=True
+    )
+    sent = fields.Boolean(
+        string='Mail notify',
+        default=False
+    )
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        record = super(ContractModification, self).create(vals_list)
```
Tardo:
```suggestion
        record = super().create(vals_list)
```

contract/models/contract_modification.py
```diff
@@ -0,0 +1,47 @@
...
+        record.check_modification_ids_need_sent()
+        return record
+
+    def write(self, vals):
+        res = super(ContractModification, self).write(vals)
```
Tardo:
```suggestion
        res = super().write(vals)
```

contract/models/contract_modification.py
```diff
@@ -0,0 +1,47 @@
...
+        self.check_modification_ids_need_sent()
+        return res
+
+    def check_modification_ids_need_sent(self):
+        records_not_sent = self.filtered(lambda x: not x.sent)
+        if records_not_sent:
+            records_not_sent.mapped('contract_id')._modification_mail_send()
```
Tardo: idk if it's possible, but mapped can return repeated items.

contract/data/mail_template.xml
```diff
@@ -62,5 +62,40 @@
...
+                    <tr>
+                        <th>State</th>
+                        <th>Description</th>
+                    </tr>
+                </thead>
+                <tbody>
+                    <t t-foreach="object.modification_ids" t-as="notification_id">
+                        <tr t-if="not notification_id.sent">
+                            <td><span t-esc="notification_id.date" /></td>
+                            <td><t t-esc="notification_id.description" /></td>
```
carlosdauden: Why t-esc instead of t-field?

contract/models/contract_modification.py
carlosdauden: Multiple records can be returned
```suggestion
        records = super(ContractModification, self).create(vals_list)
```

contract/report/report_contract.xml
```diff
@@ -99,6 +99,31 @@
                                 </table>
                             </div>
                         </div>
+                        <div class="row" id="modification_info" t-if="o.modification_ids">
+                            <div class="col-12">
+                                <p id="modification_info_p"><strong>Modifications</strong></p>
+                                <table class="table table-sm">
+                                    <thead>
+                                        <tr>
+                                            <th><strong>Date</strong></th>
```
Tardo: "TH" should apply bold style ... why need use "strong" here?

contract/report/report_contract.xml
```diff
@@ -99,6 +99,31 @@
...
+                                        </tr>
+                                    </thead>
+                                    <tbody>
+                                        <t t-foreach="o.modification_ids" t-as="notification_id">
+                                            <tr>
+                                                <td>
+                                                    <span t-esc="notification_id.date"/>
+                                                </td>
+                                                <td class="text-right">
+                                                    <t t-esc="notification_id.description"/>
```
carlosdauden: Why t-esc instead of t-field?

contract/views/assets.xml
```diff
@@ -7,4 +7,9 @@
             <script type="text/javascript" src="/contract/static/src/js/section_and_note_fields_backend.js"></script>
         </xpath>
     </template>
+    <template id="assets_frontend" inherit_id="web.assets_frontend" name="contract assets">
+        <xpath expr="." position="inside">
+            <link rel="stylesheet" href="/contract/static/src/scss/frontend.scss" type="text/scss"/>
```
Tardo: The client gets a CSS file
```suggestion
            <link rel="stylesheet" href="/contract/static/src/scss/frontend.scss" type="text/css"/>
```

carlosdauden (approved):

Tardo (changes requested):
Some comments are non-blocking ...

contract/models/contract_modification.py
victoralmau: @carlosdauden @pedrobaeza It's possible and need to improve `_modification_mail_send` function to use multi records?

carlosdauden: Char field mapped can return repeated items, but Many2one field mapped retuns a browse with unique items

contract/__manifest__.py
```diff
@@ -17,7 +17,13 @@
               "LasLabs, "
               "Odoo Community Association (OCA)",
     'website': 'https://github.com/oca/contract',
-    'depends': ['base', 'account', 'product', 'portal'],
+    'depends': [
+        'base',
+        'account',
+        'product',
+        'portal',
+        'email_template_qweb'
```
pedrobaeza: This dependency is too hard and can't be added now. When I told you about QWeb, I meant to use core feature of enveloping a QWeb

contract/data/mail_template.xml
```diff
@@ -62,5 +62,40 @@
...
     </record>
-
+    <template id="template_contract_modification" name="Contract Modification">
+        <t t-call="email_template_qweb.view_email_template_corporate_identity">
+            <p>Hello</p>
+            <p>We have modifications on the contract <strong t-esc="object.name" /> that we want to notify you.</p>
+            <table border="1">
+                <thead>
+                    <tr>
+                        <th>State</th>
```
pedrobaeza:
```suggestion
                        <th>Date</th>
```

pedrobaeza (changes requested):
Please include translations

contract/__manifest__.py
victoralmau: OK, change now to `message_post_with_view` function

pedrobaeza (approved):

victoralmau:
Sorry Pedro, fix tests and remove `social` in oca dependencies. 
When Travis tests is green, can you launch command again?

PR #602, opened 2021-01-04
[12.0][FIX] contract: Modification mail + Auto-create follower in contract creation related to partner_id

victoralmau:
- [FIX] Modification mail to correct partner_ids (add `subtype_id` in function).
- [IMP] Improve tests
- [IMP] Auto-create follower related to `partner_id`
- [FIX] I have changed `_invoice_followers` function to other place to prevent add all followers from all contracts to all invoices created.

Please, @pedrobaeza and @chienandalu  review it.
@Tecnativa TT27605

### DIFF
--- a/contract/models/contract.py
+++ b/contract/models/contract.py
@@ -147,11 +147,16 @@ class ContractContract(models.Model):
 
     @api.model
     def _set_start_contract_modification(self):
+        subtype_id = self.env.ref('contract.mail_message_subtype_contract_modification')
         for record in self:
             if record.contract_line_ids:
                 date_start = min(record.contract_line_ids.mapped('date_start'))
             else:
                 date_start = record.create_date
+            record.message_subscribe(
+                partner_ids=[record.partner_id.id],
+                subtype_ids=[subtype_id.id]
+            )
             record.write({
                 'modification_ids': [
                     (0, 0, {'date': date_start, 'description': _('Contract start')})
@@ -165,20 +170,15 @@ class ContractContract(models.Model):
                 lambda x: not x.sent
             )
             if modification_ids_not_sent:
-                contract_modification_subtype = self.env.ref(
-                    'contract.mail_message_subtype_contract_modification'
+                record.message_post_with_template(
+                    self.env.ref(
+                        "contract.mail_template_contract_modification"
+                    ).id,
+                    notif_layout="contract.template_contract_modification",
+                    subtype_id=self.env.ref(
+                        'contract.mail_message_subtype_contract_modification'
+                    ).id
                 )
-                notified_partners = record.message_follower_ids.filtered(
-                    lambda x: contract_modification_subtype in x.subtype_ids
-                ).mapped('partner_id')
-                if notified_partners:
-                    record.message_post_with_template(
-                        self.env.ref(
-                            "contract.mail_template_contract_modification"
-                        ).id,
-                        partner_ids=[(4, x.id) for x in notified_partners],
-                        notif_layout="contract.template_contract_modification",
-                    )
                 modification_ids_not_sent.write({'sent': True})
 
     @api.multi
@@ -485,7 +485,9 @@ class ContractContract(models.Model):
                 lambda x: invoice_create_subtype in x.subtype_ids
             ).mapped('partner_id')
             if partner_ids:
-                invoices.message_subscribe(partner_ids=partner_ids.ids)
+                (
+                    invoices & item._get_related_invoices()
+                ).message_subscribe(partner_ids=partner_ids.ids)
 
     @api.model
     def _finalize_and_create_invoices(self, invoices_values):
--- a/contract/tests/test_contract.py
+++ b/contract/tests/test_contract.py
@@ -135,11 +135,11 @@ class TestContract(TestContractBase):
         return self.env['contract.template.line'].create(vals)
 
     def test_add_modifications(self):
+        partner2 = self.partner.copy()
+        subtype = self.env.ref('contract.mail_message_subtype_contract_modification')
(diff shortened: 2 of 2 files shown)


pedrobaeza (commented):
Can we remove then all the previous code checking if there are recipients to send?

contract/models/contract.py
```diff
@@ -428,6 +428,7 @@ def _prepare_invoice(self, date_invoice, journal=None):
             'journal_id': journal.id,
             'origin': self.name,
             'user_id': self.user_id.id,
+            'old_contract_id': self.id
```
pedrobaeza: Please don't use this deprecated field. You need to inject the followers on the created invoices guessing them from the computed field, the same as the smart-button does.

contract/tests/test_contract.py
```diff
@@ -248,15 +252,15 @@ def test_contract_invoice_followers(self):
         self.acct_line.recurring_rule_type = 'daily'
         self.contract.pricelist_id = False
         self.contract.message_subscribe(
-            partner_ids=self.contract.partner_id.ids,
-            subtype_ids=self.env.ref(
-                'contract.mail_message_subtype_invoice_created'
-            ).ids
+            partner_ids=[self.contract.partner_id.id],
+            subtype_ids=[
+                self.env.ref('contract.mail_message_subtype_invoice_created').id
```
pedrobaeza: Why are you doing this change that doesn't bring anything?

contract/tests/test_contract.py
```diff
@@ -248,15 +252,15 @@ def test_contract_invoice_followers(self):
...
+            ]
         )
         self.contract.recurring_create_invoice()
         invoice_daily = self.contract._get_related_invoices()
         self.assertTrue(invoice_daily)
-        self.assertEqual(len(invoice_daily.message_partner_ids.ids), 2)
+        self.assertEqual(len(invoice_daily.message_partner_ids.ids), 1)
```
pedrobaeza, line 263: Why this change?

contract/tests/test_contract.py
```diff
@@ -135,11 +135,11 @@ def _add_template_line(self, overrides=None):
...
 
     def test_add_modifications(self):
+        partner2 = self.partner.copy()
+        subtype = self.env.ref('contract.mail_message_subtype_contract_modification')
         self.contract.message_subscribe(
-            partner_ids=self.contract.partner_id.ids,
-            subtype_ids=self.env.ref(
-                'contract.mail_message_subtype_contract_modification'
-            ).ids
+            partner_ids=[partner2.id],
```
pedrobaeza: Why are you preferring now this form instead of the more readable `partner_ids=partner2.ids`?

contract/tests/test_contract.py
victoralmau: IMO because before, more followers were added than they should.

pedrobaeza (approved):

contract/models/contract.py
```diff
@@ -485,7 +485,10 @@ def _invoice_followers(self, invoices):
                 lambda x: invoice_create_subtype in x.subtype_ids
             ).mapped('partner_id')
             if partner_ids:
-                invoices.message_subscribe(partner_ids=partner_ids.ids)
+                item_invoices = item._get_related_invoices()
+                invoices.filtered(
+                    lambda x: item_invoices in x
```
pedrobaeza: Wait, this is not correct, as `get_related_invoices` can return several old invoices:
```suggestion
                    lambda x: bool(item_invoices & x)
```

pedrobaeza (changes requested):

contract/models/contract.py
pedrobaeza: I have fixed it in a more elegant way:
```python
(invoices & item._get_related_invoices()).message_subscribe(...
```

pedrobaeza (approved):

pedrobaeza:
Fast-tracking due to the severity of the bug:

/ocabot merge patch

pedrobaeza:
Cherry-picked for v13 at ecacc4e8f249dd881fdda627972ae0f8a7fda87f

StefanRijnhart:
This change seems to break my tests

```
2021-01-28 10:19:20,604 951171 ERROR 12contract odoo.addons.contract.tests.test_contract: FAIL: test_contract_invoice_followers (odoo.addons.contract.tests.test_contract.TestContract) 
2021-01-28 10:19:20,605 951171 ERROR 12contract odoo.addons.contract.tests.test_contract: Traceback (most recent call last): 
2021-01-28 10:19:20,605 951171 ERROR 12contract odoo.addons.contract.tests.test_contract: `   File "contract/contract/tests/test_contract.py", line 264, in test_contract_invoice_followers 
2021-01-28 10:19:20,605 951171 ERROR 12contract odoo.addons.contract.tests.test_contract: `     self.assertTrue(invoice_daily) 
2021-01-28 10:19:20,605 951171 ERROR 12contract odoo.addons.contract.tests.test_contract: ` AssertionError: 2 != 1 
```

Subscribed users are 

```
(Pdb) invoice_daily.message_partner_ids.mapped('name')
['OdooBot', 'partner test contract']
```

Is this what you would expect?

OdooBot is the creator of the invoice and as such is created here: https://github.com/odoo/odoo/blob/12.0/addons/mail/models/mail_thread.py#L269.

This set of message partners is the same in the test before this PR was merged. I see this PR was merged when Travis was disfunctional and tests were not always run properly, but it looks like there was at least one successful travis run with this code (https://travis-ci.com/github/OCA/contract/jobs/472018591#L2078).

I'm just going to add a PR with this test change reverted and see what happens.

PR #577, opened 2020-11-14
[MIG] contract: Migration to 14.0

ozono:
contract migration

PR #606, opened 2021-01-08
14.0 mig contract_sale

dsolanki-initos:
Depends on PR #577


dsolanki-initos:
Hi @joao-p-marques Can you please review this PR?

contract_sale/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 
 {
     "name": "Contract from Sale",
-    "version": "13.0.1.0.0",
+    "version": "14.0.1.0.0",
```
joao-p-marques, line 6: This change shouldn't go here, but in the `[MIG]` commit.

joao-p-marques (commented):
@dsolanki-initos can you take a look at the comments on the commits?

Perhaps @victoralmau can help reviewing this, as he migrated it to v13.

contract_sale/__manifest__.py
```diff
@@ -0,0 +1,24 @@
+# Copyright 2017 Tecnativa - Luis M. Ontalba
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+{
+    "name": "Contract from Sale",
+    "version": "14.0.1.0.0",
+    "category": "Sales",
+    "author": "Tecnativa, " "Odoo Community Association (OCA)",
```
victoralmau:
```suggestion
    "author": "Tecnativa, Odoo Community Association (OCA)",
```

victoralmau (approved):

contract_sale/__manifest__.py
dsolanki-initos: In the pre-commit it shows error if we change it in migration commit.

joao-p-marques: Yes, that is normal, that is why we do the first `[IMP]` with `--no-verify`, to avoid those. When you change it latter in the `[MIG]` commit it will not show the error anymore.

joao-p-marques:
The commits are still the same AFAICS @dsolanki-initos

dsolanki-initos:
@joao-p-marques  I have adjusted the commits.

joao-p-marques (approved):
Thanks! Looks good now :+1: :smiley:

Issue #610, opened 2021-01-10
[13.0][contract]: portal showing "back to edit mode" for everyone


ygol:
### Reproduce:
- add a portal user as a follower of a contract
- log as that portal user and select the contract: a top bar is shown inviting to "back to edit mode"
![image](https://user-images.githubusercontent.com/2945174/104120567-78861e80-5340-11eb-9f6d-3676616a2b29.png)

### Desired:
Only show the top bar when clicking on "review" button in backend and not when accessing a contract from the portal or not for portal users

PR #611, opened 2021-01-18
[13.0][FIX] contract: portal showing "back to edit mode" for everyone

victoralmau:
[FIX] Portal showing "back to edit mode" for everyone
Fixes https://github.com/OCA/contract/issues/610

Please @joao-p-marques and @pedrobaeza review it.

@Tecnativa TT27781

### DIFF
--- a/contract/models/contract.py
+++ b/contract/models/contract.py
@@ -113,6 +113,12 @@ class ContractContract(models.Model):
         string="Modifications",
     )
 
+    def get_formview_id(self, access_uid=None):
+        if self.contract_type == "sale":
+            return self.env.ref("contract.contract_contract_customer_form_view").id
+        else:
+            return self.env.ref("contract.contract_contract_supplier_form_view").id
+
     @api.model_create_multi
     def create(self, vals_list):
         records = super().create(vals_list)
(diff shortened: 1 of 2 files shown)


pedrobaeza:
About this, the form view shouldn't be the same if you return to a customer contract or supplier one. Can you check the method that return one or another depending on this?

contract/models/contract.py
```diff
@@ -112,6 +112,20 @@ class ContractContract(models.Model):
         inverse_name="contract_id",
         string="Modifications",
     )
+    backend_view_id = fields.Integer(
+        store=False, compute="_compute_backend_view_id", string="Backend view"
+    )
+
+    def _compute_backend_view_id(self):
```
pedrobaeza: This is not the idea I had in mind. There's a standard method that returns the view to obtain. Can't this be used?

pedrobaeza (approved):

pedrobaeza:
Fast-tracking for fixing the problem:

/ocabot merge minor

is this happening in v12 as well?

victoralmau:
Yes, in 12.0 it's equal, i will create PR to change it.

PR #613, opened 2021-01-20
[12.0][FIX] contract: Changed modification_mail auto to appear like email and not internal comment

victoralmau:
Changed modification_mail auto to appear like email and not internal comment

It's only my propose, probably exists other way to do it better.

Please, @pedrobaeza review it.
@Tecnativa TT27825

### DIFF
--- a/contract/tests/test_contract.py
+++ b/contract/tests/test_contract.py
@@ -1,5 +1,6 @@
 # Copyright 2018 Tecnativa - Carlos Dauden
 # Copyright 2018 Tecnativa - Pedro M. Baeza
+# Copyright 2021 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from collections import namedtuple
@@ -134,15 +135,35 @@ class TestContract(TestContractBase):
         vals.update(overrides)
         return self.env['contract.template.line'].create(vals)
 
+    def _get_mail_messages_prev(self, contract, subtype):
+        return self.env["mail.message"].search([
+            ("model", "=", "contract.contract"),
+            ("res_id", "=", contract.id),
+            ("subtype_id", "=", subtype.id),
+        ]).ids
+
+    def _get_mail_messages(self, exclude_ids, contract, subtype):
+        return self.env["mail.message"].search([
+            ("model", "=", "contract.contract"),
+            ("res_id", "=", contract.id),
+            ("subtype_id", "=", subtype.id),
+            ('id', 'not in', exclude_ids)
+        ])
+
     def test_add_modifications(self):
         partner2 = self.partner.copy()
-        subtype = self.env.ref('contract.mail_message_subtype_contract_modification')
         self.contract.message_subscribe(
             partner_ids=partner2.ids,
-            subtype_ids=subtype.ids
+            subtype_ids=self.env.ref('mail.mt_comment').ids
         )
+        subtype = self.env.ref('contract.mail_message_subtype_contract_modification')
+        partner_ids = self.contract.message_follower_ids.filtered(
+            lambda x: subtype in x.subtype_ids
+        ).mapped('partner_id')
+        self.assertGreaterEqual(len(partner_ids), 1)
         # Check initial modification auto-creation
         self.assertEqual(len(self.contract.modification_ids), 1)
+        exclude_ids = self._get_mail_messages_prev(self.contract, subtype)
         self.contract.write({
             'modification_ids': [
                 (
@@ -163,16 +184,12 @@ class TestContract(TestContractBase):
                 )
             ]
         })
-        partner_ids = self.contract.message_follower_ids.filtered(
-            lambda x: subtype in x.subtype_ids
-        ).mapped('partner_id')
-        self.assertGreaterEqual(len(partner_ids), 2)
-        total_mail_messages = self.env["mail.message"].search_count([
-            ("model", "=", "contract.contract"),
-            ("res_id", "=", self.contract.id),
-            ("subtype_id", "=", subtype.id)
-        ])
-        self.assertGreaterEqual(total_mail_messages, 1)
+        mail_messages = self._get_mail_messages(exclude_ids, self.contract, subtype)
+        self.assertGreaterEqual(len(mail_messages), 1)
+        self.assertEqual(
+            mail_messages[0].notification_ids.mapped('res_partner_id').ids,
+            self.contract.partner_id.ids
+        )
 
     def test_check_discount(self):
         with self.assertRaises(ValidationError):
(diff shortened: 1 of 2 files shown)


carlosdauden (approved):

contract/wizards/mail_compose_message.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class MailComposer(models.TransientModel):
+    _inherit = 'mail.compose.message'
+
+    @api.model
+    def create(self, values):
+        result = super(MailComposer, self).create(values)
+        if self.env.context.get('override_subtype_id'):
+            result.subtype_id = self.env.context.get('override_subtype_id')
```
pedrobaeza: Why do you need to override the subtype?

victoralmau: Because it's incompatible define subtype A (_contract-modification_) in `message_post_with_template` function and simulate subtype B (_message_).
It's not possible too force `partner_ids` in `message_post_with_template` function because is redefined according to subtype used in wizard creation.

pedrobaeza: Can we revert then to what we have previously?

victoralmau: IMO is not possible to revert, because in "_previous commit_" https://github.com/OCA/contract/commit/25baee81f9998bd2ff7af79d6fe9e505b3939477 fix error to send mail only **contract_modification** followers.

pedrobaeza: OK, let's go for it.

contract/wizards/mail_compose_message.py
```diff
@@ -0,0 +1,16 @@
...
+from odoo import api, models
+
+
+class MailComposer(models.TransientModel):
+    _inherit = 'mail.compose.message'
+
+    @api.model
+    def create(self, values):
+        result = super(MailComposer, self).create(values)
+        if self.env.context.get('override_subtype_id'):
```
pedrobaeza: You have to do this before calling super, not after. If not, the mails will still be sent as internal.

pedrobaeza (changes requested):

pedrobaeza:
@Yajo you had a similar problem in account_credit_control. How did you get rid off of this?

yajo:
I don't understand too much what's happening here, but I'd say this module reimplements Odoo's tracking system, which shouldn't be done in the 1st place.

However, under that reimplementation, it's true that [Odoo will only render as comment those messages with the `mail.mt_comment` subtype](https://github.com/odoo/odoo/blob/9d6d25f983e633025ce140dd244705f26ed69ebf/addons/mail/models/mail_message.py#L525).

The case with `account_credit_control` was a bit different because I had to convert it from `TransientModel` to `Model` to be able to keep a thread.

pedrobaeza:
@Yajo I meant this thing, that is the exact one that is happening here:

https://github.com/OCA/credit-control/pull/86#issuecomment-689408353

@victoralmau please double check that this way the message only gets the proper recipients (adding test that have some follower with not that subtype, and checking that is not sent to him/her).

yajo:
Well, then the solution is the same as in https://github.com/OCA/credit-control/pull/86/commits/ee5c863e383d6f8b15a053e001d9c817967c6e8d#diff-ec792622ff462a31754c4bf1df43249adaab5855153e46ab04f0e9c764683d5f

However I think that's a big workaround so as to keep maintaining that on every other module. I think that either:

1. The problem of seeing a comment rendered as a note is not a big one ➡ no need for the workaround.
2. The problem is considered big enough ➡ we should create a server-ux module that adds field `forced_render_style = fields.Selection([('comment', 'Comment'), ('note', 'Note')])` to model `mail.message.subtype` and overrides the needed methods to actually fix that problem, and then these other modules inherit from that one.

I have checked that Odoo 12-14 have this problem. We could also consider proposing the required patch to master, since the need for such a module to exist is a bit absurd in itself.

pedrobaeza:
Appearing as internal note is a big problem because:

- Aesthetically, it's sent with the internal note header.
- Functionally, any possible answer is not sent to the rest of external recipients.

Anyway, creating a new module and making this module to depend on it it's not a good thing on stable, and less for a couple of lines. A solution in core for master can be proposed though.

victoralmau:
Apply similar to `account_credit_control` until we decide better solution (IMO try to add it in odoo core).

contract/tests/test_contract.py
```diff
@@ -136,11 +136,15 @@ def _add_template_line(self, overrides=None):
 
     def test_add_modifications(self):
         partner2 = self.partner.copy()
-        subtype = self.env.ref('contract.mail_message_subtype_contract_modification')
         self.contract.message_subscribe(
             partner_ids=partner2.ids,
-            subtype_ids=subtype.ids
+            subtype_ids=self.env.ref('mail.mt_comment').ids
```
pedrobaeza, line 157: Why changing this?

contract/models/mail_message.py
```diff
@@ -0,0 +1,21 @@
...
+class MailMessage(models.Model):
+    _inherit = 'mail.message'
+
+    def message_format(self):
+        result = super().message_format()
+        contract_modification = self.env['ir.model.data'].xmlid_to_res_id(
+            'contract.mail_message_subtype_contract_modification'
+        )
+        for message in result:
+            message.update({
```
pedrobaeza: Does this makes that the outgoing mail is seen as external?

contract/tests/test_contract.py
victoralmau: Because try to add new followers without modification subtype to test after.

pedrobaeza: But I don't see that you subscribe to this type after, so you are relying on default values.

victoralmau: It's related to: https://github.com/OCA/contract/pull/613#issuecomment-766145055

contract/models/mail_message.py
victoralmau: Appear as internal yet.
I have only applied commit related to https://github.com/OCA/contract/pull/613#issuecomment-766145055 but probably isn't final solution.

StefanRijnhart:
But mail_bot is autoinstalled with mail?

contract/models/mail_message.py
pedrobaeza: Can you avoid to rewrite in the message and do this change at other level?

contract/tests/test_contract.py
pedrobaeza: Please do the test in a more resilient way.

pedrobaeza:
@victoralmau in https://github.com/OCA/rma/pull/205, we have subtypes, but messages are not sent as internal. Can you please compare to see if we don't need this override?

victoralmau:
I have apply some changes but message continue appear like internal comment. In RMA is exactly equal (appear like internal comment).

pedrobaeza:
One thing is to appear in grey in the chatter, and another that is sent as internal comment, and I think that doesn't happen, isn't it?

victoralmau:
Exactly, now appear as email.

contract-modification-mail

pedrobaeza:
Great then:

/ocabot merge patch

PR #618, opened 2021-01-28
[12.0][FIX] contract, test incorrect number of message partners

StefanRijnhart:
See https://github.com/OCA/contract/pull/602#issuecomment-768963729

### DIFF
--- a/contract/tests/test_contract.py
+++ b/contract/tests/test_contract.py
@@ -260,7 +260,7 @@ class TestContract(TestContractBase):
         self.contract.recurring_create_invoice()
         invoice_daily = self.contract._get_related_invoices()
         self.assertTrue(invoice_daily)
-        self.assertEqual(len(invoice_daily.message_partner_ids.ids), 1)
+        self.assertEqual(len(invoice_daily.message_partner_ids.ids), 2)
 
     def test_contract_weekly_post_paid(self):
         recurring_next_date = to_date('2018-03-01')


pedrobaeza:
I would say this requires `assertGreaterEqual`, as not always `mail_bot` is installed.

pedrobaeza:
See also #613 for other fine-tuning

StefanRijnhart:
Ah that explains.

StefanRijnhart:
But mail_bot is autoinstalled with mail?

pedrobaeza:
Maybe it's another one that does the auto-subscription. I don't have the details right now, as it was my colleague who did it, and is on holidays now.

StefanRijnhart:
It's green, so I think it's valid in all cases. As explained in the comment on the other PR, it's in the mail module that the user is subscribed. I don't think that mail_bot changes anything in this test except that user_root's name is updated to OdooBot in this module.

pedrobaeza:
Although I need the confirmation from @victoralmau to confirm the reasons, I'm fast-tracking this for not breaking your flows, as it's green:

/ocabot merge patch

victoralmau:
Reviewing in detail I see that this change is correct, it was actually changed here https://github.com/OCA/contract/pull/602/files#diff-da995126f9f1cfc9e7074016896e39baf645c3c59d1e5364bdb098a4fb7248a9L259 incorrectly.
It is correct that there are 2 followers (admin + the client of the contract to which we have previously subscribed to the type of invoices created).

StefanRijnhart:
Thanks for confirming @victoralmau!

victoralmau:
Reviewing 13.0 in https://github.com/OCA/contract/blob/13.0/contract/tests/test_contract.py#L275 something is incorrect because appear only 1 follower in test. What do you think @pedrobaeza ?

pedrobaeza:
You should check why the difference and try to get more resilient tests checking the exact thing you want to get. Maybe `assertGreaterEqual` is the resilient solution.

PR #619, opened 2021-01-31
[13.0][FIX] contract: invoices from section contract lines

rousseldenis:
It was impossible to generate invoices from section contract lines.

image

### DIFF
--- a/contract/models/contract_line.py
+++ b/contract/models/contract_line.py
@@ -543,6 +543,9 @@ class ContractLine(models.Model):
         name = self._insert_markers(dates[0], dates[1])
         invoice_line_vals.update(
             {
+                "account_id": invoice_line_vals["account_id"]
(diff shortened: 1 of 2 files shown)


rousseldenis:
@pedrobaeza

pedrobaeza:
Can you please check pre-commit? Not sure why the account is needed to be populated this way though.

rousseldenis:
> Can you please check pre-commit? Not sure why the account is needed to be populated this way though.

Account is populated through Form() even if section or note. That triggers Odoo constraint in that case. That's because the behaviour is not fully aligned on sale behaviour for example.

pedrobaeza (approved):

pedrobaeza:
Let's fast-track it for solving the problem:

/ocabot merge patch

PR #622, opened 2021-02-02
[13.0][FIX] contract: Error in test_contract_invoice_followers

victoralmau:
Error in `test_contract_invoice_followers`
Related to https://github.com/OCA/contract/pull/618

[**SOLVED**] In contract creation auto-create 2 followers (user + customer):

In 12.0 these code work fine:
```
    partner_ids=self.contract.partner_id.ids,
    subtype_ids=self.env.ref(
        "contract.mail_message_subtype_invoice_created"
    ).ids,
))
```

@Tecnativa TT28102

### DIFF
--- a/contract/models/contract.py
+++ b/contract/models/contract.py
@@ -539,7 +539,7 @@ class ContractContract(models.Model):
 
     @api.model
     def _invoice_followers(self, invoices):
-        invoice_create_subtype = self.sudo().env.ref(
+        invoice_create_subtype = self.env.ref(
             "contract.mail_message_subtype_invoice_created"
         )
         for item in self:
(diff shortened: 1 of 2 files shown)


contract/models/contract.py
```diff
@@ -547,8 +547,9 @@ def _invoice_followers(self, invoices):
                 lambda x: invoice_create_subtype in x.subtype_ids
             ).mapped("partner_id")
             if partner_ids:
-                (invoices & item._get_related_invoices()).message_subscribe(
-                    partner_ids=partner_ids.ids
+                subtype_comment = self.sudo().env.ref("mail.mt_comment")
```
pedrobaeza: Why do you need `sudo`?

contract/models/contract.py
```diff
@@ -547,8 +547,9 @@ def _invoice_followers(self, invoices):
...
+                (invoices & item._get_related_invoices()).sudo().message_subscribe(
```
pedrobaeza: And why using `sudo` here as well?

len-foss (approved):

pedrobaeza (approved):

PR #624, opened 2021-02-04
[FIX] agreement: fixed the agreement creation from template

deu-dev:


len-foss (commented):
The changes look legit, but it needs a proper commit message to explain why the old field definition was a problem.
Also, given the change, before agreement_id was not computed; so I would expect it to be set through the code, in onchange and/or create. But no such thing is part of the diff. Unless it was already an issue that you had to manually specify this field, in a way that could be inconsistent with the section.

PR #627, opened 2021-02-14
[13.0][MIG] contract_price_revision

rousseldenis:
Includes: 

- [ ] ~~#629~~

Improvements:

- [x] Add a fix price revision (through a Selection for revision type)


rousseldenis:
@pedrobaeza I think there are some issues due to https://github.com/OCA/contract/commit/cd086ddbb4a9b85ccfcac288e53fb93e716d9446

If you don't check recurrence at line level, some constraints are triggered like this when doing a price revision (successor/predecessor lines stuff):

https://github.com/OCA/contract/blob/13.0/contract/models/contract_line.py#L361

Moreover, I think contract should be cleaned as now date_end on contract.line is compute stored and some affectations are still present in code like https://github.com/OCA/contract/blob/13.0/contract/models/contract_line.py#L361.

pedrobaeza:
Well, all the successor thing is something that I didn't see to have a counterpart in standard module, but it seems that the features are outside the main module (in your extra `contract` v12 modules that you did for that customer), so I don't really see how this affects, but if you put some steps to reproduce the problem with standard OCA module, I can check. If it comes from other way, you may need to program that change on that part.

Are you proposing to remove that constraints? Without knowing the detail, maybe the constraint can be nullify if the recurrence is not at line level.

rousseldenis:
> Without knowing the detail, maybe the constraint can be nullify if the recurrence is not at line level.

That's the point

pedrobaeza:
OK, then if you have it clear, you can propose a PR with that changes. As said, all the successor thing is a total foreigner for me as no UI about that on standard module.

rousseldenis:
> OK, then if you have it clear, you can propose a PR with that changes. As said, all the successor thing is a total foreigner for me as no UI about that on standard module.

The code was there during migration... so, don't understand the question

pedrobaeza:
What I mean is that those fields AFAIK are not set in standard code, so the constraint doesn't raise there.

rousseldenis:
> What I mean is that those fields AFAIK are not set in standard code, so the constraint doesn't raise there.

It does in GUI, I think adding a flush() in the tests should help to highlight it. I will add a screenshot

pedrobaeza:
OK, please, I didn't know how to activate it. Some comments in README would be good.

rousseldenis:
@pedrobaeza Finally, I found the problem (described in last commits). As computed  stored values are not recomputed after a new(), the use of copy() is the real solution.

pedrobaeza:
Great that you find the problem! You can even pass the new values as arguments in `copy` call:

```python
new_line = line.copy(self._get_new_line_value(line))
```

PR #629, opened 2021-02-14
[13.0][IMP] contract: Set onchange on contract line date_start

rousseldenis:
Forwarding from 12.0

### DIFF
--- a/contract/tests/test_contract.py
+++ b/contract/tests/test_contract.py
@@ -552,6 +552,14 @@ class TestContract(TestContractBase):
         res = line._onchange_product_id()
         self.assertFalse(res["domain"]["uom_id"])
 
+    def test_contract_onchange_date_start(self):
+        """Recurring next date should be filled in"""
+        line = self.env["contract.line"].new({"date_start": "2021-01-01"})
+        line._onchange_date_start()
+        self.assertEquals(
+            fields.Date.from_string("2021-01-01"), line.recurring_next_date,
+        )
+
     def test_contract_onchange_product_id_domain(self):
         """It should return UoM category domain."""
         line = self._add_template_line()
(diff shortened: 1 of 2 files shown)


carlosdauden (approved):

contract/tests/test_contract.py
```diff
@@ -20,6 +20,7 @@ class TestContractBase(common.SavepointCase):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
+        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
```
pedrobaeza: Please don't do that, as what this lighten the tests, may cause other side effects (see for example #618 that tests this kind of things)

pedrobaeza (commented):
I'm not sure why this is needed, as now the fields are computed writable with proper depends. Can you put a user case that depicts the steps to reproduce the associated problem?

contract/tests/test_contract.py
rousseldenis: This is true only if you test messages, which is not the case here...

pedrobaeza: I have put you an example on something that is coming to 13.0, so please don't do it, and it's totally out of scope of this PR. What is the gain in performance with this anyway (out of curiosity)?

rousseldenis: Ok, I'll remove it.

For instance, without that we are at 24.76s, 20259 queries.
With: 20.40s, 18229 queries

As said, this shouldn't be used if we test messages for chatter. Instead, it should be enabled.

pedrobaeza:
Don't get me wrong, Denis: of course I love to have improvements on test times when possible, but in this case I'm not sure if it's going to introduce side effects on the upcoming PRs, and more important, this PR was for another thing, so introducing both changes here and unfocusing the discussion is not the best way to get this merged, and that's why I asked you to remove it. Let's do this: when fw-porting the other patch checking followers and messages, we will introduce the improvement to see if it affects.

About this PR, I still don't see why it's needed, and the test doesn't help, as it's not a real case, but mangling with virtual records.

PR #634, opened 2021-02-16
[FIX] contract: consistency between contract count and contrct action

len-foss:
On the commercial entity, the count shows all contracts of child partners.
However clicking on the action might show no contract, as the action only
restricted the domain to the partner itself.
We extract the domain into a function to share it in both cases.

### DIFF
--- a/contract/models/res_partner.py
+++ b/contract/models/res_partner.py
@@ -21,10 +21,14 @@ class ResPartner(models.Model):
         string="Contracts",
     )
 
+    def _get_partner_contract_domain(self):
+        self.ensure_one()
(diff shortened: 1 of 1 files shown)


sbejaoui (approved):

marcelsavegnago (approved):

carlosdauden (approved):

pedrobaeza:
Thanks for it!

/ocabot merge patch

Can you please check 13.0 as well?

len-foss:
No problem. It's directly fixing the line we changed yesterday, so forward-port is both necessary and trivial.

PR #635, opened 2021-02-16
[FIX] contract: consistency between contract count and contract action

len-foss:
On the commercial entity, the count shows all contracts of child partners.
However clicking on the action might show no contract, as the action only
restricted the domain to the partner itself.
We extract the domain into a function to share it in both cases.


len-foss:
Forward-port of https://github.com/OCA/contract/pull/634

pedrobaeza:
Fast-tracking as straight-forward:

/ocabot merge patch

PR #646, opened 2021-03-01
[12.0][FIX] agreement_legal: Add active field on Agreement Stage

newtratip:
This pr add active field on agreement stage for archive on agreement (No need to delete)


newtratip:
@OCA/project-service-maintainers, Could you check this please ?

dreispt (approved):

PR #649, opened 2021-03-03
[14.0][ADD] agreement_legal_contract

newtratip:
This module is connector between `agreement_legal` and `contract`.

cc: @aomorawanc

### DIFF
--- a/agreement_legal_contract/__init__.py
+++ b/agreement_legal_contract/__init__.py
@@ -1,0 +1,4 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html)
+
+from . import models
+from . import wizards
--- a/agreement_legal_contract/__manifest__.py
+++ b/agreement_legal_contract/__manifest__.py
@@ -1,0 +1,21 @@
+# Copyright 2021 Ecosoft Co., Ltd (http://ecosoft.co.th)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html)
+
+{
+    "name": "Agreements Legal Contract",
+    "summary": "Create contract from agreement",
+    "author": "Ecosoft, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/contract",
+    "category": "Agreement",
+    "license": "AGPL-3",
+    "version": "14.0.1.0.0",
+    "depends": ["agreement_legal", "contract"],
+    "data": [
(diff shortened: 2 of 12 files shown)


etobella:
@newtratip I have been checking your module but I have one question.

You are relating the agreement with contract, but shouldn't be related to contract line?

Let me explain. We usually manage maintenance contracts. Each renewal we will create a new agreement (a renewal), but the contract should be the same with several lines. WDYT?

If you like the idea. We can make a proposal keeping your commits.

kittiu:
@etobella I am interested to hear your use case. Do you mean by, i.e.,

* Agreement with 3 product lines -> create Contract with 3 lines
* then, renew Agreement (with no line, but refer to old Agreement) -> create Contract with 3 lines

OR

* Agreement with 0 product lines (don't care lines) -> create Contract with 3 lines
* then, renew Agreement -> create Contract with 3 lines.

> You are relating the agreement with contract, but shouldn't be related to contract line?

Are you saying, agreement OR agreement.line that related to contract.line ? I also not clear why you think current relation between agreement and contract is not enough for your case.

I am sorry, not sure you still remember the case, it was quite sometime.

rousseldenis:
@etobella Could you answer @kittiu questions ?

rousseldenis:
@kittiu My question: Why depending on agreement_legal and not just on agreement as far as I can see, no code depends on that one?

etobella:
> @etobella I am interested to hear your use case. Do you mean by, i.e.,
> 
> * Agreement with 3 product lines -> create Contract with 3 lines
> * then, renew Agreement (with no line, but refer to old Agreement) -> create Contract with 3 lines

It might be possible to add the renewal on the same contract :thinking: 

> 
> OR
> 
> * Agreement with 0 product lines (don't care lines) -> create Contract with 3 lines
> * then, renew Agreement -> create Contract with 3 lines.
> 
> > You are relating the agreement with contract, but shouldn't be related to contract line?
> 
> Are you saying, agreement OR agreement.line that related to contract.line ? I also not clear why you think current relation between agreement and contract is not enough for your case.

I think it is better to relate agreement line with contract line.

> 
> I am sorry, not sure you still remember the case, it was quite sometime.

nimarosa:
Hello, I want to reopen this discusión since I think a link into agreement and contract are more than necessary. I think the user expect contract and agreements linked.

PR #656, opened 2021-03-17
[13.0][FIX] contract: Changed modification_mail auto to appear like email and not internal comment

victoralmau:
Changed modification_mail auto to appear like email and not internal comment
Fix generate recurring invoice error (https://github.com/odoo/odoo/pull/67879)

Related to 12.0: https://github.com/OCA/contract/pull/613

Please, @pedrobaeza review it.
@Tecnativa TT27825

### DIFF
--- a/contract/models/contract.py
+++ b/contract/models/contract.py
@@ -519,6 +519,8 @@ class ContractContract(models.Model):
                     # nullifying line
                     invoice_vals["invoice_line_ids"].append((0, 0, invoice_line_vals))
             invoices_values.append(invoice_vals)
+            # Force the recomputation of journal items
+            del invoice_vals["line_ids"]
             contract_lines._update_recurring_next_date()
         return invoices_values
 
--- a/contract/models/contract.py
+++ b/contract/models/contract.py
@@ -1,9 +1,10 @@
 # Copyright 2004-2010 OpenERP SA
 # Copyright 2014 Angel Moya <angel.moya@domatix.com>
 # Copyright 2015-2020 Tecnativa - Pedro M. Baeza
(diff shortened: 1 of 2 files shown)


contract/models/contract.py
```diff
@@ -519,6 +519,7 @@ def _prepare_recurring_invoices_values(self, date_ref=False):
                     # nullifying line
                     invoice_vals["invoice_line_ids"].append((0, 0, invoice_line_vals))
             invoices_values.append(invoice_vals)
+            del invoice_vals["line_ids"]
```
pedrobaeza, line 523: Why this?

victoralmau: IMO it's very important in 13.0 and i have splited in other commit "[FIX] contract: Generate recurring invoice" because now recurring invoice creation process "fail" (create invoice empty, without lines). It's related to https://github.com/odoo/odoo/pull/67879 and these change is my suggested solution to solve it.

pedrobaeza: Then why you don't put any info about this in the commit message? If you don't add such information, we can't know why something is done. Anyway, not sure the solution is the correct. I will check later.

pedrobaeza: OK, I have seen the commit and indeed the best solution is to remove this, as we aren't able to reuse the Form logic for getting this part. The performance is suffering a lot though, which is a pity...

pedrobaeza: Please change the commit message and a code comment like `# Force the recomputation of journal items`, and we are good to go

pedrobaeza (approved):

ernestotejeda (approved):
code and functional review

PR #658, opened 2021-03-31
Style changed on contract form

Algoritmun:
From h3 to h1 it was changed the title of the contract to like like other model's forms.


pedrobaeza:
When using such big title in other places, the label is not shown. You should also squash your commits.

PR #663, opened 2021-04-16
migración agreement_legal

marionumza:


pedrobaeza:
Thanks for the contribution.

Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-13.0.

If the jump is between several versions, you have to modify the source branch in the main command to accommodate it to this circumstance.

marionumza:
Sorry, I wanted to do it privately and then if I was successful doing the PR in the OCA / contract repository, I had no idea that it would have an immediate effect, I'm going to read the procedure

PR #666, opened 2021-04-28
[MIG][14.0] agreement_legal (Migration from v12 to v14)

patrickrwilson:
Migrating agreement_legal from v12 to v14


patrickrwilson:
I will need to move this out of my base 14.0 clone and into it's own branch.

PR #667, opened 2021-04-30
[13.0][FIX] contract: Align contract line form view to inline behaviour

rousseldenis:
When editing contract line in inline mode, product is not mandatory (and should stay as is). Adapt form view as in that one, product is mandatory (and should not)

### DIFF
--- a/contract/views/abstract_contract_line.xml
+++ b/contract/views/abstract_contract_line.xml
@@ -16,11 +16,7 @@
                         col="4"
                         attrs="{'invisible': [('display_type', '!=', False)]}"
                     >
-                        <field
-                            colspan="4"
-                            name="product_id"
-                            attrs="{'required': [('display_type', '=', False)]}"
-                        />
+                        <field colspan="4" name="product_id" />
                         <label for="quantity" />
                         <div class="o_row">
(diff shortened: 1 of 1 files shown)


carlosdauden (approved):

bodedra (approved):

pedrobaeza:
I'm not sure about this, and maybe the alignment should be the other way, or do more things, due to the "automatic price" mechanism, that depends on having a product selected.

rousseldenis:
> I'm not sure about this, and maybe the alignment should be the other way, or do more things, due to the "automatic price" mechanism, that depends on having a product selected.

In another words, the original behaviour avoid having recurring line configuration without product(sometimes a simple description is sufficient)

pedrobaeza:
Yes, that's it. I forgot to include such conditions when I did the inline view.

pedrobaeza (changes requested):
Discussion in progress

rousseldenis:
> Yes, that's it. I forgot to include such conditions when I did the inline view.

That's the aim of this PR. IMHO, should be merged as is. @pedrobaeza

pedrobaeza:
No, it should be the contrary: put the required on the other view, and let this one.

PR #668, opened 2021-04-30
[13.0][FIX] contract: Set layout for section and note in portal template

rousseldenis:


contract/views/contract_portal_templates.xml
```diff
@@ -227,56 +227,87 @@
...
+                                                <td
+                                                    name="td_recurring_interval"
+                                                    class="text-right"
+                                                >
+                                                    <span
+                                                        t-field="line.recurring_interval"
+                                                    />
+                                                    <t
+                                                        t-if="line.recurring_rule_type=='daily'"
+                                                    >Day(s)</t>
```
pedrobaeza, line 261: I think this can be obtained programatically instead of having to manually re-map the values to their display name.

rousseldenis: > I think this can be obtained programatically instead of having to manually re-map the values to their display name.

These lines are not changed (indent due to ```<t>``` introduction)

rousseldenis: This is a copy paste of what is done on sale level

pedrobaeza: OK, but it can still apply, and simplify a bit the diff of the PR. Something like `line._fields['recurring_rule_type'].selection[line.recurring_rule_type]` or similar if I don't remember bad.

rousseldenis: > OK, but it can still apply, and simplify a bit the diff of the PR. Something like `line._fields['recurring_rule_type'].selection[line.recurring_rule_type]` or similar if I don't remember bad.

Ok, I got what you mean. This is not the aim of this PR as that code was there before.

The aim here is to avoid products and quantities description when you want to display a note or a section

pedrobaeza: I know, but it was for avoiding huge diff. Do you want me to propose the other imp in a separate PR and then rebase over it?

rousseldenis: Before :

image

After :

image

victoralmau (approved):

pedrobaeza:
Let's merge this and let the other improvement for future PR:

/ocabot merge minor

@victoralmau please check the suggested improvement and move both to v14.

PR #679, opened 2021-05-21
[14.0][ADD] contract_state [WIP]

kittiu:
Add states and action buttons to contracts.


kittiu:
We have the need for state fields in contracts. I wonder if states should be included in the base addon `contract`, or keep it separated? (also wonder why it has no states, it is a also a document, isn't it).

cc @pedrobaeza @MiquelRForgeFlow @ThomasBinsfeld 

WDYT?

pedrobaeza:
IMO there's no sense for states, as it's determined by the invoiceability of the lines.

kittiu:
> IMO there's no sense for states, as it's determined by the invoiceability of the lines.

I can see that the main feature of this module is about recurring invoice, for which the state of line level make sense.

For our case, If the contract is not confirmed yet, user / scheduled job shouldn't create invoice. We use for government, a bit bureaucratic, need l.e., signed agreement.

So, may I work as separated module here. If it finally make so sense here, I can move it out later.

pedrobaeza:
I think that you want more or less to convert contract into what `agreement_legal` serves. There were previous threads for merging both, but the work was not done, but now you can develop a link between both instead of overriding the contract with new things.

kittiu:
> I think that you want more or less to convert contract into what `agreement_legal` serves. There were previous threads for merging both, but the work was not done, but now you can develop a link between both instead of overriding the contract with new things.

Thank you @pedrobaeza I will consider your comment :)

PR #680, opened 2021-05-21
[13.0][FIX] contract: Force add `access_token` in **View contract** button from contract modification mail

victoralmau:
Force add `access_token` in **View contract** button from contract modification mail.

It's need to apply in 14.0 too.

@Tecnativa TT29820

### DIFF
--- a/contract/data/mail_template.xml
+++ b/contract/data/mail_template.xml
@@ -91,6 +91,16 @@
     >
         <xpath expr="//t[@t-raw='message.body']" position="after">
             <t t-raw="0" />
+            <t t-if="record._name == 'contract.contract'">
+                <t
+                    t-set="share_url"
+                    t-value="record._get_share_url(redirect=True, signup_partner=True, share_token=True)"
+                />
+                <t
+                    t-set="access_url"
+                    t-value="is_online and share_url and base_url + share_url or ''"
+                />
+            </t>
         </xpath>
     </template>
     <template id="template_contract_modification" name="Contract Modification">


contract/data/mail_template.xml
```diff
@@ -63,6 +63,9 @@
         <p></p>
     </div>
     <p></p>
+    <p>Base url: ${object.get_base_url()}</p>
```
pedrobaeza: Why putting these things visible in the template?

victoralmau: Ups, sorry.

chienandalu (approved):

pedrobaeza (changes requested):
Same topic as in DMS: regularize way of populating the URL and don't pass access token when the argument is False.

contract/models/contract.py
```diff
@@ -177,6 +177,18 @@ def _compute_access_url(self):
         for record in self:
             record.access_url = "/my/contracts/{}".format(record.id)
 
+    def _get_share_url(
+        self, redirect=False, signup_partner=False, pid=None, share_token=True
+    ):
+        if redirect:
```
pedrobaeza: I don't think both arguments are related, so this seems more a trick than the real solution.

victoralmau: I have removed `_get_share_url` function and i have used other trick to solve PR description.

contract/data/mail_template.xml
```diff
@@ -91,6 +91,16 @@
     >
         <xpath expr="//t[@t-raw='message.body']" position="after">
             <t t-raw="0" />
+            <t t-if="record._name == 'contract.contract'">
```
pedrobaeza, line 94: Shouldn't this be before t-raw? Have you checked it's working?

victoralmau: I don't remember motive to set `t-raw=0` previously but with these changes all work fine, i have tested with some contract modification and mails and access_token have added to link button.

pedrobaeza (approved):

pedrobaeza (approved):

PR #681, opened 2021-05-23
[12.0][IMP] product_contract: Implements parameter to define the quantity (one or sale_order_line.product_uom_qty)

marcelsavegnago:
Implements parameter to define the quantity that should be used in the creation of a recurring contract.

image


marcelsavegnago:
Another option would be to place this parameter on the product.

rousseldenis:
@marcelsavegnago

PR #684, opened 2021-06-04
[13.0][FIX] contract: Remove noupdate in mail template data

joao-p-marques:
@Tecnativa
TT29820

ping @pedrobaeza @carlosdauden

### DIFF
--- a/contract/__manifest__.py
+++ b/contract/__manifest__.py
@@ -29,6 +29,7 @@
         "data/contract_cron.xml",
         "data/contract_renew_cron.xml",
         "data/mail_template.xml",
+        "data/template_mail_notification.xml",
         "data/mail_message_subtype.xml",
         "data/ir_ui_menu.xml",
         "wizards/contract_line_wizard.xml",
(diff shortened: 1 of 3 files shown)


pedrobaeza:
You should remove it only for the qweb views, not for all.

joao-p-marques:
Ok, I had to separate it in 2 files, was that what you meant?

pedrobaeza:
Not really. You can put several `<data>` sections in the same file. The requirement is to have the common ancestor `<odoo>` in that case. Anyways, we usually split files by model, so it's not bad to move the others to `views/template_....xml`.

joao-p-marques:
Ok, thanks @pedrobaeza 

Changes made

pedrobaeza (approved):
Usual path is in `templates` or in `views` (see https://github.com/OCA/maintainer-tools/tree/master/template/module/templates), but let's continue as it's a minor detail.

chienandalu (approved):

PR #685, opened 2021-06-08
[IMP] contract_queue_job: jobify renew method for crons

len-foss:


len-foss:
@sbejaoui Was there a reason why this function was not also jobified?

contract_queue_job/models/contract_line.py
```diff
@@ -0,0 +1,21 @@
...
+class ContractLine(models.Model):
+
+    _inherit = "contract.line"
+
+    @job(default_channel=QUEUE_CHANNEL)
+    def renew(self):
+        if len(self) > 1:
+            for rec in self:
+                rec.with_delay().renew()
+            return self.env["account.invoice"]
```
sbejaoui:
```suggestion
            return self.env["contract.line"]
```

contract_queue_job/models/contract_line.py
```diff
@@ -0,0 +1,21 @@
+# Copyright 2021 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+from odoo.addons.queue_job.job import job
+
+QUEUE_CHANNEL = "root.CONTRACT_INVOICE"
```
sbejaoui: IMO, it's better to do it on a specific channel 

```suggestion
QUEUE_CHANNEL = "root.CONTRACT_LINE_RENEW"
```

sbejaoui (approved):
small suggestions

PR #692, opened 2021-06-27
[14.0][FIX] contract: Defensive patch when next date is in the past

flotho:
if the recurring next date is in the past then the last_date_invoiced is returned as False there :+1: https://github.com/OCA/contract/blob/14.0/contract/models/contract_recurrency_mixin.py#L185 so the strftime failed.

If in any way the cron couldn't vee executed the nx date could be in the past.

This trivial  patch prevent this kind of issue.

### DIFF
--- a/contract/models/contract_line.py
+++ b/contract/models/contract_line.py
@@ -606,8 +606,10 @@ class ContractLine(models.Model):
         lang = lang_obj.search([("code", "=", self.contract_id.partner_id.lang)])
         date_format = lang.date_format or "%m/%d/%Y"
         name = self.name
-        name = name.replace("#START#", first_date_invoiced.strftime(date_format))
-        name = name.replace("#END#", last_date_invoiced.strftime(date_format))
+        if first_date_invoiced:
+            name = name.replace("#START#", first_date_invoiced.strftime(date_format))
+        if last_date_invoiced:
+            name = name.replace("#END#", last_date_invoiced.strftime(date_format))
         return name
 
     def _update_recurring_next_date(self):


pedrobaeza:
I don't have clear the steps to reproduce the problem.

flotho:
Hi @pedrobaeza you're right, actually I found other issue.
Please add WIP, I'm digging

flotho:
Here is the situation for the customer contract :
image
I think this case shouldn't have been possible. Yet I'm not sure what was my customer process.
As I can imagine he would like to get "older" invoices to have all the due invoices so he was probably expecting to generate cron multiple times and get the corresponding invoices.

pedrobaeza:
I think that screenshot data is correct, but `next_period_date_end` in that case shouldn't be higher than 31/05/2021

flotho:
Hi @pedrobaeza you're totally right, 
> I think that screenshot data is correct, but `next_period_date_end` in that case shouldn't be higher than 31/05/2021

date_end and recurring_next_date are available for input so user could get in this strange sitaution.
when you change the date_end and set it up with a date upper that the recurring_next_date you'll have the issue.
What could be the process to prevent this ?

pedrobaeza:
But if they have input such thing, they can undo it to correct data, isn't it?

flotho:
For sure they could correct it.
Yet for the fun, let me expose what they are think about.
They imagine that they could generate invoices in the past because they are migrating to Odoo. They know that their contract is ending in May but want to get all the invoices generated from January to May.
IMHO, this shouldn't be "allowed" because it causes too much possible issues regarding accounting, yet I'm wondering what could be our defensive  "message"/patch that could prevent this case.

flotho:
The case is difficult to reproduce.
The customer recreate a proper contract.
I close this one

PR #694, opened 2021-07-01
[13.0][FIX] contract: Cleanup values on line when invoicing

pedrobaeza:
It seems that having several fields returned by `Form` as `False` that are related to one2many inverse field, makes Odoo ORM mad, and get to an unbalanced move when generating the invoice.

Cleaning these values assures to work without problems.

Not able to provide a regression test, as I don't get to reproduce the conditions to happen in test environment, but the patch is safe and harmless anyways.

@Tecnativa TT30842

### DIFF
--- a/contract/models/contract.py
+++ b/contract/models/contract.py
@@ -516,7 +516,9 @@ class ContractContract(models.Model):
                 invoice_line_vals = line._prepare_invoice_line(move_form=move_form)
                 if invoice_line_vals:
                     # Allow extension modules to return an empty dictionary for
-                    # nullifying line
+                    # nullifying line. We should then cleanup certain values.
+                    del invoice_line_vals["company_id"]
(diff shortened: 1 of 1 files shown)


carlosdauden (approved):
Some other community member would raise questions about this patch and demand that the original issue be fixed, but I'm going to be pragmatic and approve the PR :wink:

pedrobaeza:
Yeah, I tried hard to have a regression test to prove it, but unfortunately in this time I'm not able to prove it more than this, but doing a code review you can check that is very safe to do it.

carlosdauden:
I understand it perfectly

victoralmau (approved):

chienandalu (approved):

pedrobaeza:
Cherry-picked to v14 at c41a35b29a5a24492e66c5e110612a660d231a47

PR #696, opened 2021-07-02
[14.0][FIX] agreement_legal (Removed old demo file causing error)

patrickrwilson:
There was an old demo file in the data folder that caused conflicts during install when demo data is not needed. The install was successful if demo data was set however if demo data was not an error would occur during install.


newtratip (approved):
LGTM :+1:

PR #706, opened 2021-07-26
[12.0][ADD] contract_variable_qty_sale_order_line

marcelsavegnago:
This formula makes sense when using the product_contract module which changes the amount of contracted service to 1 when the contract is created.

However, there are cases where the quantity informed in the Sales Order is important. Examples: Recurrence Contract in hours or Recurrence Contract based on number of users.


rousseldenis (approved):
Code review

max3903 (approved):

PR #710, opened 2021-08-20
[12.0][IMP] product_contract: termination_notice_interval on sale_order form

marcelsavegnago:
This PR changes the module's behavior to allow that via sales order it is possible to change the default value of the termination_notice_interval and termination_notice_rule_type fields that come from the product.

Makes sense ?


rousseldenis:
@sbejaoui What do you think ?

max3903 (approved):

PR #713, opened 2021-08-30
[14.0][FIX] contract: fix an obvious crash when self is a multi-recordset

alexis-via:
_compute_contract_count() is an api.multi
It calls self._get_partner_contract_domain(), so you can't have self.ensure_one() in there !!!
Quite an obvious bug.


pedrobaeza (approved):

pedrobaeza:
Fast-tracking this obvious patch:

/ocabot merge patch

PR #714, opened 2021-09-10
[13.0][FIX] contract: Fix invoice creation and salesperson assignment

joao-p-marques:
In v13, the `user_id` field is a related field to `invoice_user_id` (https://github.com/odoo/odoo/blob/13.0/addons/account/models/account_move.py#L191), that defaults to the environment user (`self.env.user`).
Therefore, if we try to create an invoice just by passing `user_id`, it is overwritten by the default computation of `invoice_user_id`.
This fixes it by passing the correct field and data.

@Tecnativa
TT31715

ping @pedrobaeza @victoralmau

### DIFF
--- a/contract/models/contract.py
+++ b/contract/models/contract.py
@@ -411,7 +411,7 @@ class ContractContract(models.Model):
                 "invoice_date": date_invoice,
                 "journal_id": journal.id,
                 "invoice_origin": self.name,
-                "user_id": self.user_id.id,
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):

carlosdauden (approved):

victoralmau (approved):

pedrobaeza:
Move it to v14 as well

PR #715, opened 2021-09-10
[14.0][FIX] contract: Fix invoice creation and salesperson assignment

joao-p-marques:
FWP of https://github.com/OCA/contract/pull/714

In v14, the `user_id` field is a related field to `invoice_user_id`, that defaults to the environment user (`self.env.user`).
Therefore, if we try to create an invoice just by passing `user_id`, it would be overwritten by the default computation of `invoice_user_id`.
This fixes it by passing the correct field and data.

@Tecnativa
TT31715

ping @pedrobaeza @carlosdauden

### DIFF
--- a/contract/models/contract.py
+++ b/contract/models/contract.py
@@ -441,7 +441,7 @@ class ContractContract(models.Model):
                 "invoice_date": date_invoice,
                 "journal_id": journal.id,
                 "invoice_origin": self.name,
-                "user_id": self.user_id.id,
+                "invoice_user_id": self.user_id.id,
             }
         )
         return invoice_vals, move_form
--- a/contract/tests/test_contract.py
+++ b/contract/tests/test_contract.py
@@ -299,6 +299,18 @@ class TestContract(TestContractBase):
         self.assertTrue(invoice_daily)
         self.assertTrue(self.contract.partner_id in invoice_daily.message_partner_ids)
 
+    def test_contract_invoice_salesperson(self):
+        self.acct_line.recurring_next_date = "2018-02-23"
+        self.acct_line.recurring_rule_type = "daily"
+        new_salesperson = self.env["res.users"].create(
+            {"name": "Some Salesperson", "login": "salesperson_test"}
+        )
+        self.contract.user_id = new_salesperson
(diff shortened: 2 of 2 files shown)


victoralmau (approved):

joao-p-marques:
Now includes some fixes to `agreement_sale_condition_delivery_planning` because of pre-commit errors...

agreement_sale_condition_delivery_planning/__manifest__.py
```diff
@@ -8,10 +8,11 @@
     "development_status": "Alpha",
     "category": "Sale",
     "author": "Camptocamp, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/contract",
     "maintainers": ["gurneyalex"],
     "license": "AGPL-3",
     "application": False,
-    "installable": True,
+    "installable": False,
```
pedrobaeza: You are disabling a module. That can't be done without a strong reason.

joao-p-marques: Sorry for not explaining...

The module was added a few days ago coming from v12: https://github.com/OCA/contract/commit/1602e240aee8337788924836f45f467f9f620865#diff-691bfd59e98c623738e91124f0f2c1bb00e4f5219a2cecb0380c2cac7f5541dc

It seems not to be migrated, as the version still indicates v12, so I think it is better to just ignore it until it is migrated... That seems to be what is done in some other situations like these...

joao-p-marques: Seeing the commit, I can have even been an error, where the new module shouldn't be uploaded... Perhaps @gurneyalex can give some insight...

pedrobaeza: It seems he had some dirty files in his local repository. Better to revert such commit directly. Please do it in a separate PR.

joao-p-marques: Ok, done in https://github.com/OCA/contract/pull/716 :+1:

pedrobaeza:
Please remove the second commit

joao-p-marques:
Done :+1:

PR #721, opened 2021-10-07
[13.0][FIX] contract: security group_account_invoice

albariera:
In this PR, the permission given to the gruoup_account_invoice in security is changed from the "contract_sale" module to the "contract" module. If the two modules were installed, it worked perfectly. The problem came if you only had installed the contract module. In this case, the group_account_invoice could not view contracts. So I just changed the security to the more general "contract" module.

### DIFF
--- a/contract/security/contract_security.xml
+++ b/contract/security/contract_security.xml
@@ -40,4 +40,16 @@
             name="domain_force"
         >['|',('company_id','=',False),('company_id','in',company_ids)]</field>
     </record>
+    <!-- For the sake of the OCA ecosystem, please do not remove this rule, as
+       it is used in other modules as contract_sale and maintenance_equipment_contract-->
+    <record id="contract_contract_see_all" model="ir.rule">
+        <field name="name">See All Contracts</field>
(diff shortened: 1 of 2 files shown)


pedrobaeza:
I don't see this need. If no record rule, no restriction and no need to put a `[(1, =, 1)]` rule. The problem you have should be because you have manual record rules applying to this contract (or leftovers from migration).

etobella:
Every time we link a module with contract we need to add a rule and a rule for invoicing saying 1, '=', 1. So, wouldn't be easier to add that on the parent module?

It happens with contract_sale, and should happen with maintenance_equipment_contract (https://github.com/OCA/maintenance/pull/170)

pedrobaeza:
The risk is that someone deletes it as useless in any refactoring of the main module. You can put a explicit comment about that. Although it's not my main preference, we can tolerate this rule for the sake of the whole OCA ecosystem :smiley: 

@albariera please put both changes in the same commit, as they are totally related and needs to be understood as a whole, although they refer to 2 different modules.

pedrobaeza (approved):

etobella (approved):

PR #723, opened 2021-10-08
[13.0][FIX] agreement_legal: Change create_new_version() function in agreements

victoralmau:
Change `create_new_version()` function in agreements + Add tests.

Please @pedrobaeza and @joao-p-marques can you review it?

@Tecnativa TT32359

### DIFF
--- a/agreement_legal/models/agreement.py
+++ b/agreement_legal/models/agreement.py
@@ -256,7 +256,7 @@ class Agreement(models.Model):
         inverse_name="parent_agreement_id",
         string="Previous Versions",
         copy=False,
-        domain=[("active", "=", False)],
+        context={"active_test": False},
     )
     child_agreements_ids = fields.One2many(
         comodel_name="agreement",
@@ -356,7 +356,7 @@ class Agreement(models.Model):
     )
 
     # Create New Version Button
-    def create_new_version(self, vals):
+    def create_new_version(self):
         for rec in self:
             if not rec.state == "draft":
                 # Make sure status is draft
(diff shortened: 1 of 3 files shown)


agreement_legal/models/agreement.py
```diff
@@ -371,7 +371,7 @@ def create_new_version(self, vals):
             # Increment the Version
             rec.version = rec.version + 1
         # Reset revision to 0 since it's a new version
-        vals["revision"] = 0
+        vals = {"revision": 0}
```
pedrobaeza: I think the logic is incorrect, as you are writing `revision = 0` on the old agreement, not in the new. Is this how it's supposed to work?

victoralmau: Yes, the logic seems wrong. You should define `revision = 0` for new records (child records in these case).
Thinking about this issue, IMO increase the revision is not correct when performing this action. What do you think?

pedrobaeza: How does this work in v12?

victoralmau: In v12 it has the strange behavior.
https://github.com/OCA/contract/blob/9893ba7589c67451e7f2cc071b2ca211d2706f9b/agreement_legal/models/agreement.py#L319-L336

victoralmau: Changes done to standardize this behavior and show "archived" versions.

pedrobaeza (approved):

pedrobaeza:
Fast-tracking for fixing the bug:

/ocabot merge patch

PR #700, opened 2021-07-16
[14.0][ENH] agreement_legal

newtratip:
**Edited**

- Add terminated stage
- Remove duplicate `ir.module.category` (Agreement)
- Remove model not used
  - `agreement.renewaltype` 
  - `agreement.increasetype`
- Remove file not used 
  - `agreement_status.py`
  - `product_template.py`
- Remove field not used
  - Description field in `agreement.stage`
  - Term field in `agreement`
  -  Currency field in `agreement`
  - Product_ids field in `agreement`
- Add active field on submenu of configuration menu
  - Stages
  - Sub-Types
- Add search view
  - Stages
  - Sub-Types
- Add context `{'default_is_template': False}` on the agreement and `{'default_is_template': True}` on the template
- Invisible is_template field
- Add web_ribbon on template
- Hide agreement field on recital, section, clause, appendix on the agreement
- Remove domain widget, it's not used.
- Stage in agreement form view can not click, I add clickable=1 to it
- Add default agreement's stage is New
- Reset agreement revision to 0 since it's a new agreement version
- Fixed https://github.com/OCA/contract/issues/640
- Fixed missing save button https://github.com/OCA/contract/issues/500
- Fixed https://github.com/OCA/contract/issues/502 (switched created by and created date)
- Agreement user role can create agreement from template
Selection_363
- Agreement can not edit if setting readonly = True in stage
Selection_359

I have fix following above, Could you review this ? 
@patrickrwilson @ygol @kittiu

PR #669, opened 2021-05-04
[14.0][MIG] agreement_legal (Version 12.0 to 14.0)

patrickrwilson:

PR #726, opened 2021-10-11
[13.0][ENH] agreement_legal

etobella:
This PR applies several commits found on 14 that were not present on 13.0

#669 #696 #700 #725

TODO: proably it is needed to add a migration script for the change from product.template to agreement.line


etobella:
@newtratip @patrickrwilson @victoralmau

pedrobaeza:
#725 should be already in 13.0

etobella:
Sorry @pedrobaeza You are right, I had to remake commits in order to check that everything working fine. I think now it could be right

pedrobaeza:
About the `is_old_version` things, it's worse solution than active=False with simply a context `active_test=False` like we have here in 13.0, and also a lot of extra empty lines have been added with no value.

etobella:
@pedrobaeza I think your comments have been attended correctly (at least is what I hope :smile: )

pedrobaeza:
Well, I see very heavy changes here, and knowing that this has been merged in 14.0 doesn't comfort me, as it's removing some features as I can see reading the diff. I also see that the first commit added a demo file that is removed on the second commit, so they should be both merged and neutralized.

etobella:
Well, removed features were unused features. They couldn't be accessed from the agreement and had no sense if they couldn't be edited. However, I think we should need to create some extra features like how to manage renewals and so on. But this should go on a new PR, this intends to have both branches similar

pedrobaeza:
I hope that they are unused features. Please do at least the squash for avoiding useless diffs.

agreement_legal/data/agreement_stage.xml
```diff
@@ -1,4 +1,3 @@
-<?xml version="1.0" ?>
```
olgamarcocb, line 1: Why are you removing this? You deleted a lot of headers

olgamarcocb (approved):
:+1:

agreement_legal/data/agreement_stage.xml
pedrobaeza: Yes, and creating incorrect XML files. They also don't include the encoding, which is necessary.

etobella: Fixed, now the headers are correctly formed

pedrobaeza:
Let's not block this anymore, although there were no need to put `utf-8` in upper:

/ocabot merge major

The problem now is to align again 14.0 version reverting the is_old_version thing.

PR #727, opened 2021-10-12
[IMP] agreement_legal: Improve multicompany management

etobella:
While checking the module I found there was a problem as you cannot define an agreement for another company. This should solve the problem.

### DIFF
--- a/agreement_legal/views/agreement.xml
+++ b/agreement_legal/views/agreement.xml
@@ -155,10 +155,12 @@
                             </div>
                         </group>
                         <group name="company" string="Company">
-                            <div class="o_address_format">
+                            <div
+                                class="o_address_format"
+                                groups="base.group_multi_company"
+                            >
                                 <field
                                     name="company_id"
-                                    readonly="1"
                                     context="{'show_address': 1}"
                                     options="{'always_reload': True}"
                                 />


pedrobaeza:
I don't get the goal of the patch and the double field declaration. Why is it needed?

etobella:
Without the patch the company is not editable. I wanted to show it when it was not multicompany but not editable and editable when we are in a multicompany environment

pedrobaeza:
It's weird that adding `no_create: True` makes the field editable, but anyways, if it's the way, don't repeat the field and add it to the existing one with the multi-company group. Odoo doesn't show the company on the rest of models if you don't have such model. The reasoning behind it is that if you don't use the group, it's supposed you don't handle multiple companies, and thus, the company info is redundant.

etobella:
Well, the change came from the readonly, the 'no_create' was no necessary. Now should be a lesser diff

pedrobaeza (approved):

pedrobaeza:
Do you let the group showing no matter the multi company group?

etobella:
Yes, because without it the align of the information looses sense, there are other options, but it implies to change more the view

pedrobaeza:
Ok, let's not complicate it more. Do you know why Travis is red?

olgamarcocb (approved):
:+1:

etobella:
I don't know why it was red. I have rebased and now it seems right :smile:

PR #728, opened 2021-10-15
[14.0][FIX] agreement_legal: Change image field name in view.

victoralmau:
Change image field name in view.

Please @joao-p-marques and @pedrobaeza can you review it?


pedrobaeza (approved):
Shouldn't this be changed in 13.0 as well?

joao-p-marques (approved):

PR #730, opened 2021-10-15
[IMP] agreement_legal: Improve management of end dates

olgamarcocb:
Working with the module we found we can have agreements without end date. For example, when we make an agreement for a designation, no end date is defined.

With this change, we can specify if the field is required on the agreement type


etobella (approved):

albariera (approved):

AaronHForgeFlow (approved):
Code review :+1:

Issue #502, opened 2020-04-21
[12.0][BUG] agreement_legal: wrong revision date


ygol:
When creating a new version for an agreement, the creation date of the "OLD" version is set to today while the date of the "NEW" version is still the date when the 1st version was created.
These two dates should be switched.

max3903:
@murtuzasaleh Log your time on task 19158

PR #733, opened 2021-10-19
[13.0][IMP+OU-ADD] agreement_legal: Replace the created_by + date_created fields + Openupgrade convert_binary_field_to_attachment 

victoralmau:
Changes done:
- [IMP] The created_by + date_created fields are removed and will be displayed in the view create_uid + create_date or related fields (create_uid_parent + create_date_parent). https://github.com/OCA/contract/pull/726
- [OU-ADD] Openupgrade convert_binary_field_to_attachment 

Please @pedrobaeza and @joao-p-marques can you review it?

@Tecnativa TT32469

### DIFF
--- a/agreement_legal/migrations/13.0.1.0.0/post-migration.py
+++ b/agreement_legal/migrations/13.0.1.0.0/post-migration.py
@@ -1,0 +1,13 @@
+# Copyright 2021 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+from openupgradelib import openupgrade, openupgrade_90
+
+attachment_fields = {
+    "agreement": [("signed_contract", None)],
+}
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade_90.convert_binary_field_to_attachment(env, attachment_fields)
--- a/agreement_legal/views/agreement.xml
+++ b/agreement_legal/views/agreement.xml
@@ -365,8 +365,27 @@
                             name="revision"
                             readonly="True"
                         />
-                        |  Created By: <field name="created_by" readonly="True" />
-                        |  Created On: <field name="date_created" readonly="True" />
+                        |  Created By:
+                        <field
+                            name="create_uid"
+                            readonly="True"
+                            attrs="{'invisible': [('parent_agreement_id', '!=', False)]}"
+                        />
+                        <field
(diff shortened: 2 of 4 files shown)


joao-p-marques (commented):
Is it important to have these fields filled for new installs as well? If so, maybe it could be converted into an init hook, for example.

If not, LGTM :slightly_smiling_face:

victoralmau:
> Is it important to have these fields filled for new installs as well? If so, maybe it could be converted into an init hook, for example.
> 
> If not, LGTM slightly_smiling_face

Sorry, I've changed the commit completely.
Can you check it again?

agreement_legal/migrations/13.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2021 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+from openupgradelib import openupgrade, openupgrade_90
+
+attachment_fields = {
+    "agreement": [("signed_contract", None)],
```
pedrobaeza, line 7: You haven't renamed the column in pre.

agreement_legal/views/agreement.xml
```diff
@@ -365,8 +365,27 @@
...
                         />
-                        |  Created By: <field name="created_by" readonly="True" />
-                        |  Created On: <field name="date_created" readonly="True" />
+                        |  Created By:
+                        <field
+                            name="create_uid"
+                            readonly="True"
+                            attrs="{'invisible': [('parent_agreement_id', '!=', False)]}"
+                        />
+                        <field
```
pedrobaeza, line 374: Why 2 fields?

victoralmau: In the migration to v14 https://github.com/OCA/contract/pull/700 this issue was resolved https://github.com/OCA/contract/issues/502 and the `created_by` + `date_created` fields were added. In my opinion, those fields would not be necessary because when the record has `parent_agreement_id` it should show the date of that parent record. What do you think about it?

pedrobaeza (approved):

pedrobaeza:
@etobella do you agree?

etobella (approved):
Makes sense to me. I don't get why you need to rename the field, but I assume it has been deleted or something like that.

PR #737, opened 2021-10-22
[IMP] agreement_legal: Add res_partner security group

olgamarcocb:
Without this change, if a user accessed partner but had not access to agreement, an error was raised


albariera (approved):
LGTM :+1:

AaronHForgeFlow (approved):
LGTM

Better that giving read access to agreements to all groups with access to partners.

PR #661, opened 2021-04-07
[14.0][MIG]Migrated contract_sale_generation in v14

Freni-OSI:
Migrated Module contract_sale_generation from v12 to v14

PR #741, opened 2021-11-04
[14.0][IMP] contract

rousseldenis:
I will:

- [x] Update cron process to generate invoices for multi company
- [x] Add tests for multi company
- [x] Add a generation type field to avoid code duplication for sale generation See https://github.com/OCA/contract/pull/661


rousseldenis:
@manuelregidor @MiquelRForgeFlow This is a more correct solution of https://github.com/OCA/contract/pull/755

See commit : https://github.com/OCA/contract/pull/741/commits/f015ee5fdc0790fbd129c641ca73b7f245018c87

rousseldenis:
I will rebase this

flotho:
hi @rousseldenis , any review needed ?

rousseldenis:
@flotho Of course.

As said, this is more correct than https://github.com/OCA/contract/pull/755/files#diff-bd92412e26bfd77a1e9fb18d4a493b62b8a9c1ec64cbf57a807f244f12c198fcR608

As the search is done once and using with_company() instead of ugly 'allowed_company_ids'

contract/models/contract.py
```diff
@@ -595,19 +599,44 @@ def _recurring_create_invoice(self, date_ref=False):
...
-            contracts_to_invoice = self.search(row["__domain"]).with_context(
-                allowed_company_ids=[row["company_id"][0]]
-            )
-            invoices |= contracts_to_invoice._recurring_create_invoice(date_ref)
-        return invoices
+        # Use ids instead of recordset for perfomances reasons
+        for company in companies:
+            contracts_to_invoice = contracts.filtered(
+                lambda c: c.company_id == company
+            ).with_company(company)
```
rousseldenis, line 635: @MiquelRForgeFlow Could you please review this ?

I want to use with_company that is better than using allowed_company_ids in context (that part is in Odoo core).

carlosdauden (approved):
Resolve conflicts

MiquelRForgeFlow (approved):
LGTM

contract/models/abstract_contract.py
```diff
@@ -48,6 +48,20 @@ class ContractAbstractContract(models.AbstractModel):
...
     )
+    generation_type = fields.Selection(
+        string="Generation Type",
+        selection=lambda self: self._get_generation_type_selection(),
+        default=lambda self: self._get_default_generation_type(),
+        help="Choose the document that will be automatically generated by cron.",
+    )
+
+    @api.model
+    def _get_generation_type_selection(self):
```
pedrobaeza: Please use conventions for method naming (https://www.odoo.com/documentation/15.0/es/developer/misc/other/guidelines.html#symbols-and-conventions): `_selection_generation_type` and `_default_generation_type`.

contract/models/contract.py
```diff
@@ -595,19 +599,44 @@ def _recurring_create_invoice(self, date_ref=False):
...
+            getattr(contracts_to_invoice, _recurring_create_func)(date_ref)
+        return True
+
+    @api.model
+    def cron_recurring_create_invoice(self, date_ref=None):
+        return self._cron_recurring_create(date_ref)
```
pedrobaeza: I would suggest the explicit form for avoiding misunderstandings:

```suggestion
        return self._cron_recurring_create(date_ref, create_type="invoice")
```

contract/tests/test_contract.py
```diff
@@ -2412,3 +2412,7 @@ def test_contract_action_preview(self):
         action = self.contract.action_preview()
         self.assertIn("/my/contracts/", action["url"])
         self.assertIn("access_token=", action["url"])
+
+
+class TestContractCase(TestContract, common.SavepointCase):
```
pedrobaeza: Why adding this?

pedrobaeza (changes requested):

contract/models/abstract_contract.py
rousseldenis: Arf didn't know Odoo has done such requirements :smile: But great!

rousseldenis:
@MiquelRForgeFlow Rebased on your last work
@pedrobaeza Requested changes attended

rousseldenis:
@pedrobaeza This one is ready

contract/tests/test_contract.py
```diff
@@ -17,7 +17,7 @@ def to_date(date):
     return fields.Date.to_date(date)
 
 
-class TestContractBase(common.SavepointCase):
+class TestContractBase:
```
pedrobaeza: Please don't remove the base inheritance from here.

contract/models/contract.py
```diff
@@ -595,23 +599,45 @@ def _recurring_create_invoice(self, date_ref=False):
...
-            )
-            invoices |= contracts_to_invoice._recurring_create_invoice(date_ref)
-        return invoices
+        # Use ids instead of recordset for perfomances reasons
+        for company in companies:
+            contracts_to_invoice = contracts.filtered(
+                lambda c: c.company_id == company
+                and (not c.date_end or c.recurring_next_date <= c.date_end)
+            ).with_company(company)
+            getattr(contracts_to_invoice, _recurring_create_func)(date_ref)
```
pedrobaeza: I don't like too much this obscure way to execute methods. When you try to search for calling occurrences, you don't get anything, and that's very confusing. You can prepare a hook that explicitly call each method depending on the type.

pedrobaeza (commented):
Maybe you can squash a bit the commit history: 10 commits for 6 files seems too much.

contract/tests/test_contract.py
rousseldenis: That's done with purpose as tests for multi company import account base ones that inherits from SavePointCase.

contract/models/contract.py
rousseldenis: Solved

contract/tests/test_contract.py
pedrobaeza: But this forces all of the inherited tests to multi-inherit from SavepointCase and this one, being very prompt to be forgotten and not having real tests performed. If you keep it, is there any error?

rousseldenis: Of course, you cannot inherit from same one twice.

rousseldenis:
@pedrobaeza 

> Maybe you can squash a bit the commit history: 10 commits for 6 files seems too much.

No, it reflects development

contract/tests/test_contract.py
pedrobaeza: Well, I don't like to change this base structure, due to the reasons above. Can you please find another solution?

rousseldenis:
So this one is really ready :smile: 
@pedrobaeza

contract/models/contract.py
```diff
@@ -595,23 +599,47 @@ def _recurring_create_invoice(self, date_ref=False):
...
-            contracts_to_invoice = (
-                self.search(row["__domain"])
-                .with_context(allowed_company_ids=[row["company_id"][0]])
-                .filtered(
-                    lambda a: not a.date_end or a.recurring_next_date <= a.date_end
-                )
-            )
-            invoices |= contracts_to_invoice._recurring_create_invoice(date_ref)
-        return invoices
+        # Use ids instead of recordset for perfomances reasons
```
pedrobaeza: I think this comment doesn't apply anymore, isn't it?

rousseldenis: Done

pedrobaeza (approved):
Thanks for the efforts. Do you plan to fw-port it to v15?

MiquelRForgeFlow:
@rousseldenis will you forward port this?

rousseldenis:
> @rousseldenis will you forward port this?

I'll do oca-port yes!

rousseldenis:
> @rousseldenis will you forward port this?

Done: https://github.com/OCA/contract/pull/776

PR #652, opened 2021-03-09
14.0 mig product_contract

mourad-ehm:

PR #744, opened 2021-11-09
[14.0][MIG] product_contract

daerio:


pedrobaeza:
Thanks for the contribution.

Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-14.0.

If the jump is between several versions, you have to modify the source branch in the main command to accommodate it to this circumstance.

brendapaniagua (commented):
Thank you very much for your support. I think I have encountered an
error
 error when accessing a sales order.  See image Attache.

Thanks.

product_contract/models/sale_order.py
```diff
@@ -0,0 +1,147 @@
...
+                    lambda r: not r.contract_id and r.product_id.is_contract
+                )
+                line_to_update_contract = rec.order_line.filtered(
+                    lambda r: r.contract_id
+                    and r.product_id.is_contract
+                    and r
+                    not in r.contract_id.contract_line_ids.mapped("sale_order_line_id")
+                )
+                if line_to_create_contract or line_to_update_contract:
+                    rec.need_contract_creation = True
```
ajaniszewska-dev, line 46: if above conditions are not met, there wont be value assigned to this field, probably that's why you have an error

ygol:
what is the difference and does this PR deprecates #652 ? TIA

rousseldenis:
@daerio Could you please review that one instead as it was done before ?

JorgeQuinteros:
Hi @pedrobaeza , I think this PR should be closed, it is not active since last year and there is another similar [PR](https://github.com/OCA/contract/pull/652)

PR #745, opened 2021-11-16
[12.0][FIX] contract: correct relation of old_contract_id

MiquelRForgeFlow:
If you have an account.analytic.account without contract line, then the contract is not ported to new table contract_contract. And then, old_contract_id of account_invoice, as now will point to contract.contract, happens that will point to a non-existing contract. Odoo detects that and breaks.

### DIFF
--- a/contract/migrations/12.0.4.0.0/pre-migration.py
+++ b/contract/migrations/12.0.4.0.0/pre-migration.py
@@ -97,17 +97,24 @@ def create_contract_records(cr):
         INSERT INTO contract_contract
         SELECT * FROM account_analytic_account
         WHERE id IN (SELECT DISTINCT {} FROM contract_line)
+            OR id IN (SELECT DISTINCT contract_id FROM account_invoice)
         """).format(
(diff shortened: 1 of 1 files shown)


pedrobaeza:
I would prefer to merge the contract in that case as inactive. In fact, there was some code for that if I recall correctly.

LoisRForgeFlow (approved):

MiquelRForgeFlow:
@pedrobaeza 
image
You only make inactive if not recurring_invoices

pedrobaeza:
Then expand that to cover such situation

MiquelRForgeFlow:
But that would be fake inactive contracts.

pedrobaeza:
Why they are fake? If there's an invoice with such relation, it has been a contract in a moment in time.

MiquelRForgeFlow:
@pedrobaeza done

MiquelRForgeFlow:
and green

pedrobaeza (approved):

PR #750, opened 2021-11-24
[IMP] agreement_legal: extend button

olgamarcocb:
This button allows the user to create a copy.

This can be used in order to create an extension without losing the original one.

### DIFF
--- a/agreement_legal/views/agreement.xml
+++ b/agreement_legal/views/agreement.xml
@@ -34,6 +34,12 @@
                         attrs="{'invisible': [('state', '=', 'active')]}"
                     />
                     <button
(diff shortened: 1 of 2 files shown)


agreement_legal/views/agreement.xml
```diff
@@ -33,6 +33,12 @@
                         class="oe_highlight"
                         attrs="{'invisible': [('state', '=', 'active')]}"
                     />
+                    <button
+                        string="Extend"
```
rousseldenis, line 37: I don't think 'Extension' is a good term as it does nothing more than copying.

Why not simply 'Copy' ?

@olgamarcocb

PR #770, opened 2022-01-11
Update fr_FR.po

MarcDataEng:
"Contact lié au contrat" doesn't make sense in this case
"Contact facturé" seems more adequate


rousseldenis:
@MarcSIO Thanks for this but the good way to update translation is through https://translation.odoo-community.org/

rousseldenis:
@MarcSIO Did you find your way ?

PR #772, opened 2022-01-24
[14.0][IMP] contract: Remove deprecated domain in onchange

rousseldenis:


rousseldenis:
Fixes:

image

pedrobaeza:
But there should be a replacement through a static domain or something.

pedrobaeza:
Maybe through `=?` operator.

rousseldenis:
> But there should be a replacement through a static domain or something.

But correct me if I'm wrong, the current behaviour is ineffective. The domain return has no influence.

So, this can be merged and another PR should be done to get the domain

pedrobaeza:
Being deprecated doesn't mean the domain is not working. It should, as this message is a warning for being replacing it without still removing the support.

rousseldenis:
> Being deprecated doesn't mean the domain is not working. It should, as this message is a warning for being replacing it without still removing the support.

Indeed, I'll modify this accordingly

contract/tests/test_contract.py
```diff
@@ -483,22 +504,35 @@ def test_onchange_partner_id(self):
             self.contract.partner_id.property_product_pricelist,
         )
 
+    def test_onchange_partner_id_domain(self):
+        with Form(self.contract) as contract_form:
+            contract_form.invoice_partner_id = self.partner_2
+        self.assertNotEqual(
```
rousseldenis: @pedrobaeza Both changes done and functional in views.

The first test for uom is ok, but this one does not seem to work. Do you have an idea?

pedrobaeza (approved):

contract/tests/test_contract.py
pedrobaeza: Are you trying to test domains? I think that through direct assignation, there's no domain restriction applied.

rousseldenis: In uom the behaviour is working in test. If you try to assign a wrong uom, it bypasses it.

rousseldenis: @pedrobaeza Or do you know the property of a field that has the evaluated domain value ?

pedrobaeza: I have never tested such domains being frank. The only way I know right now is to call fields_view_get and check the domain attribute in the resulting xml. I can investigate later.

rousseldenis: Ok, I have found the way to test it. Set the invoice_partner_id to void before

rousseldenis: So, this one is ready

pedrobaeza: Great, TIL about this way of testing domains through Form

rousseldenis:
Don't find the way to test that (Form() does not take into account field domain not defined on views). But at least the domains are working

rousseldenis:
> Don't find the way to test that (Form() does not take into account field domain not defined on views). But at least the domains are working

Tested the domain result in view as suggested by @pedrobaeza

pedrobaeza:
OK, don't worry. Not sure if the `fields_view_get` thing is adding enough value to the tests, and they may be conflicting even if other customizations are done. I would remove them. They get meaning when they are dynamic according anything, and you test that they get the proper dynamic value. I will investigate if there's a way to test such domains.

hparfr (approved):

PR #783, opened 2022-02-11
ADJ: When generating invoices from the Wizard to create invoices …

jcdrubay:
…manually, reraise the UserErrors with information about which contract triggered the exception to facilitate the investigation by accountants in case invoices are not generated automatically.

Initial request: https://github.com/OCA/contract/issues/781

PR #785, opened 2022-02-18
ADJ: When generating invoices from the Wizard to create invoices manu…

jcdrubay:
…ally, reraise any UserError with information about which contract triggered the exception to facilitate the investigation by accountants in case invoices are not generated automatically.

Forward port of https://github.com/OCA/contract/pull/783

### DIFF
--- a/contract/wizards/contract_manually_create_invoice.py
+++ b/contract/wizards/contract_manually_create_invoice.py
@@ -2,6 +2,13 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from odoo import _, api, fields, models
+from odoo.exceptions import (
+    AccessDenied,
+    AccessError,
+    MissingError,
+    UserError,
+    ValidationError,
+)
 
 
 class ContractManuallyCreateInvoice(models.TransientModel):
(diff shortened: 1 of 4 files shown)


jcdrubay:
@pedrobaeza Only runbot fails, but I guess this is ok since runboat is the new runbot.

contract/wizards/contract_manually_create_invoice.py
```diff
@@ -52,7 +59,24 @@ def create_invoice(self):
...
+                ValidationError,
+            ) as oe:
+                raise UserError(
+                    _(
+                        "Failed to process the contract %(name)s [id: %(id)s]:\n%(ue)s",
+                        name=contract.name,
+                        id=contract.id,
+                        ue=repr(oe),
+                    )
+                ) from oe
```
jcdrubay, line 78: From v14, only this line was changed to comply with pre-commit/flake8 requirement:

```
contract/wizards/contract_manually_create_invoice.py:71:17: B904 Within an `except` clause, raise exceptions with `raise ... from err` or `raise ... from None` to distinguish them from errors in exception handling.  See https://docs.python.org/3/tutorial/errors.html#exception-chaining for details.
```

pedrobaeza (approved):
runbot is not working with 15.0

hieulucky111 (approved):
LGTM.

PR #792, opened 2022-02-25
[14.0][FIX] contract: Don't update recurring_next_date as computed

rousseldenis:


rousseldenis:
@carlosdauden @pedrobaeza 

I don't have the exact explanation but I have recomputation problems on 'recurring_next_date' value when other compute methods depend on it.

I know this is breaking some tests but I have tested some flows on interface side and no problem.

pedrobaeza:
Well, you know my answer: we need the steps to reproduce it. With them, I can help you to find the solution.

rousseldenis:
> Well, you know my answer: we need the steps to reproduce it. With them, I can help you to find the solution.

@pedrobaeza In fact, when you changed recurring_next_date to computed field, you forgot the records that have next_start_date == False.

So, the write on recurring_next_date was a hack to set the correct value.

This commit fixes that : https://github.com/OCA/contract/pull/792/commits/b56236796b6fc2738c7d400a54f4c14f4d73858a

In addition, this breaks contract_sale_generation, highlighting a behaviour that should be copied from contract: https://github.com/OCA/contract/pull/792/commits/39046a6adee4786cffabbdaa120bf466afe14e81

I've added too a missing start date in contract_sale_generation tests when creating contract. And a bunch of test with another recurring rule type.

rousseldenis:
@pedrobaeza As get_next_invoice_date' function does take into account False values, filtered() is no more necessary.

Done in the last commit

sbejaoui (approved):
LGTM, code review

pedrobaeza (commented):
It seems weird that a method that is called "_update_recurring_next_date", doesn't update `recurring_next_date`. Maybe a refactoring for doing it other way is needed?

rousseldenis:
> It seems weird that a method that is called "_update_recurring_next_date", doesn't update `recurring_next_date`. Maybe a refactoring for doing it other way is needed?

We could do that later. For now, I'd prefer fixing the behaviour.

carlosdauden (approved):

pedrobaeza (changes requested):
I know what that later means, which is not doing it at all, or have a time where the code is ugly and not easy debuggable due to that inconsistencies. I prefer to resolve everything in one shot, as I still don't have steps to reproduce the problem in UI.

rousseldenis:
> I know what that later means, which is not doing it at all, or have a time where the code is ugly and not easy debuggable due to that inconsistencies. I prefer to resolve everything in one shot, as I still don't have steps to reproduce the problem in UI.

We can change method name but that means it will break all overrides people did (and goes against stable policy).

Do we want to do this in this version?

rousseldenis:
> as I still don't have steps to reproduce the problem in UI.

I can try to reproduce it **but**: code was wrong! A pure compute field should not be updated manually (if there is no inverse - we had cache values problems for records not updated, that's why it's complicated to reproduce), moreover, the use of filtered was wrong in compute method. Nothing not understandable...

rousseldenis:
@pedrobaeza Moreover, don't forget that's something you introduced here https://github.com/OCA/contract/commit/33743841012b29bf4064d16dcd2f91195bae93c7#diff-99d8269b34038533ecf9ccb6ca400b316e5cbd5b428647e0db172f94d9b5cbc8R94

So, say thank you and don't block or I'll find someone to merge this.

lmignon (approved):
Thank you for the fix @rousseldenis. LGTM and all the explanations highlight the reasoning behind your changes. It's true that it's strange to have a method named `_update_recurring_next_date` that don't update directly the `recurring_next_date` field. Nevertheless, the update of the `last_date_invoiced`will trigger the recompute of `recurring_next_date` therefore the final behavior is conform the method name.

pedrobaeza:
> @pedrobaeza In fact, when you changed recurring_next_date to computed field, you forgot the records that have next_start_date == False.
> 
> So, the write on recurring_next_date was a hack to set the correct value.
> 
> This commit fixes that : [b562367](https://github.com/OCA/contract/commit/b56236796b6fc2738c7d400a54f4c14f4d73858a)

That was done on purpose, as it's a computed writable field that should react like an onchange. I think your problem is that you are not initializing properly the lines, and try to move such bad initialization values to existing logic. If not, I don't get why aren't you able to reproduce the problem in UI.

sbejaoui:
Hi everyone,

The write on the computed field was from legacy code (before the 13.0 refactoring).
It has no added value now (after the write, and in all cases, the compute is recalled). Ideally, all the logic should be in the compute method.

It's true that it doesn't make sense to have a method named `_update_recurring_next_date` that don't update `recurring_next_date`. But, it is not safe to do such refactoring on a stable version. This must be done within the next migration and renamed to `_update_last_date_invoiced`

So I propose to add a _# FIXME_ and explain why this method must change its name so it will be considered later and not forgotten. 

Beside all this, the changes are understandable and make sense to keep the logic of a computed field exclusively in its compute method.

Best regards,

rousseldenis:
> So I propose to add a _# FIXME_ and explain why this method must change its name so it will be considered later and not forgotten.

@sbejaoui Done

pedrobaeza (approved):
OK, let's go this way and see if there's any side effect

pedrobaeza:
Can you fw-port it to v15?

PR #793, opened 2022-02-25
[FIX] [14.0] Removed external dependencies already defined in Odoo base

Shide:
Removed external dependencies already defined in Odoo base

https://github.com/odoo/odoo/blob/ae6f414ba41eee4e14235a1e82fd5ebd237dd97f/requirements.txt#L40

MT-282 @moduon

### DIFF
--- a/contract/__manifest__.py
+++ b/contract/__manifest__.py
@@ -17,7 +17,6 @@
     "author": "Tecnativa, ACSONE SA/NV, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/contract",
     "depends": ["base", "account", "product", "portal"],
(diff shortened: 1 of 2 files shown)


yajo (approved):

pedrobaeza:
This should be removed instead, as it's part of the Odoo standard requirements:

https://github.com/odoo/odoo/blob/ae6f414ba41eee4e14235a1e82fd5ebd237dd97f/requirements.txt#L40

pedrobaeza:
Merge aborted due to my last comment.

pedrobaeza (changes requested):
Remove external dependency instead

pedrobaeza (approved):

PR #800, opened 2022-03-09
Oca port pr 792 from 14.0 to 15.0

rousseldenis:


rousseldenis:
@pedrobaeza

rousseldenis:
Thanks to 
image

:sweat_smile:

pedrobaeza (approved):
I don't think this deserves 4 commits to be added on the commit history, some of them undoing part of what was done on the previous ones, but it seems I'm the only one with this feeling.

rousseldenis:
> I don't think this deserves 4 commits to be added on the commit history, some of them undoing part of what was done on the previous ones, but it seems I'm the only one with this feeling.

Each one has its own signification and has sense separately. I didn't do commits with just 'fix' comment. Coding is writing a story. :wink:

pedrobaeza:
For me (and your future you), that history is not adding too much value IMO. The important thing is the final decisions, and put in code the relevant whys, and in commit message, the expanded history (if needed). Both principles has saved me from screwing up things when coming back to the code after some time, not such commit history. But that's my opinion, of course.

flotho (approved):
easy port.
LGTM
Thanks @rousseldenis

PR #810, opened 2022-03-23
[14.0][ADD] contract_update_last_date_invoiced

rousseldenis:


scigghia (approved):
LGTM

carlosdauden:
Approved 😊

pedrobaeza:
Fixed CI in #892

/ocabot merge nobump

rafaelbn:
Hello @rousseldenis , sorry by asking 😄 I quite frustrated actually.

Why this module? I obvious the BUG in 13 and 14 with the `next_invoice_date` and `last_date_invoice`, why not to fix the main module and make a new one?

I'm testing know in runboat 13 and 14 the module doesn't work propertly. Basically if you create invoice and don't validate, just delete it because something was wrong you have to... delete de contract fully and create a new one because dates all time recalculated wrongly.

So why this module if it fix a bug? is any use case where this module is not needed?

Thanks in advance 🤗

rousseldenis:
> Hello @rousseldenis , sorry by asking 😄 I quite frustrated actually.
> 
> Why this module? I obvious the BUG in 13 and 14 with the `next_invoice_date` and `last_date_invoice`, why not to fix the main module and make a new one?
> 
> I'm testing know in runboat 13 and 14 the module doesn't work propertly. Basically if you create invoice and don't validate, just delete it because something was wrong you have to... delete de contract fully and create a new one because dates all time recalculated wrongly.
> 
> So why this module if it fix a bug? is any use case where this module is not needed?
> 
> Thanks in advance 🤗

This is not about a bug, but linked to a  base behaviour at purpose.

The initial bug has to be fixed (I'd prefer the way I've done). The last invoice date is defined as a read-only field as it is normally updated automatically.

This module allows to update it to recompute the next generation date. This is necessary when people deleted manually invoices for instance.

As I consider the base module is tough enough with full of code, I'd prefer putting logic in a new one.

I hope this explanation is clear enough 😅

rafaelbn:
Thank you for your explanation! 🤗

PR #813, opened 2022-03-29
[15.0][MIG] contract_queue_job

palkesh-aktivsoftware:
Standard Migration


pablontura (approved):
LGTM

juliap98 (approved):
LGTM

rousseldenis:
@palkesh-aktivsoftware Thanks for this. Could you rename your last commit with something like '[15.0][MIG] contract_queue_job' ?

palkesh-aktivsoftware:
> @palkesh-aktivsoftware Thanks for this. Could you rename your last commit with something like '[15.0][MIG] contract_queue_job' ?

sure, I made this change. Please review it.

contract_queue_job/__manifest__.py
```diff
@@ -6,7 +6,7 @@
     "summary": """
         This addon make contract invoicing cron plan each contract in a job
         instead of creating all invoices in one transaction""",
-    "version": "12.0.1.1.0",
+    "version": "15.0.1.1.0",
```
rousseldenis, line 9: You should reset minor version too

15.0.1.0.0

ygol:
How may I access a runboat instance for this PR? 
I would like to check if I can reproduce that invoices are created every time the cron is launched resulting in multiple time the same draft invoice with the invoice date = when the cron was launched:
image

rousseldenis (changes requested):

PR #821, opened 2022-05-09
[14.0][FIX] Add check_company=True on contracts and contract lines

alexis-via:
Add check_company=True in field definition and
[('company_id', '=', company_id)] in the domain (in view or in field definition)

### DIFF
--- a/contract/models/abstract_contract.py
+++ b/contract/models/abstract_contract.py
@@ -22,7 +22,12 @@ class ContractAbstractContract(models.AbstractModel):
     partner_id = fields.Many2one(
         comodel_name="res.partner", string="Partner", index=True
     )
-    pricelist_id = fields.Many2one(comodel_name="product.pricelist", string="Pricelist")
+    pricelist_id = fields.Many2one(
+        comodel_name="product.pricelist",
+        string="Pricelist",
+        check_company=True,
+        domain="[('company_id', '=', company_id)]",
+    )
     contract_type = fields.Selection(
         selection=[("sale", "Customer"), ("purchase", "Supplier")],
         default="sale",
@@ -36,6 +41,7 @@ class ContractAbstractContract(models.AbstractModel):
         store=True,
         readonly=False,
         index=True,
+        check_company=True,
     )
     company_id = fields.Many2one(
         "res.company",
--- a/contract/models/contract.py
+++ b/contract/models/contract.py
@@ -21,6 +21,7 @@ class ContractContract(models.Model):
     _name = "contract.contract"
     _description = "Contract"
     _order = "code, name asc"
+    _check_company_auto = True
     _inherit = [
         "mail.thread",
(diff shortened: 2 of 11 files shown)


mazairenouraly (approved):

leemannd (approved):

rousseldenis (approved):

contract_payment_mode/models/contract.py
```diff
@@ -7,8 +7,9 @@ class ContractContract(models.Model):
     payment_mode_id = fields.Many2one(
         comodel_name="account.payment.mode",
         string="Payment Mode",
-        domain=[("payment_type", "=", "inbound")],
+        domain="[('payment_type', '=', 'inbound'), ('company_id', '=', company_id)]",
```
pedrobaeza, line 10: This should have pre-commit passed

pedrobaeza (changes requested):

alexis-via:
Should be green and ready to merge

carlosdauden (approved):

contract_payment_mode/models/contract.py
pedrobaeza: Please pass pre-commit, or this can't be merged.

leemannd:
The pre-commit has been fixed and @pedrobaeza 's comment has been answered.

pedrobaeza:
Please fw-port it to upper versions.

contract/models/abstract_contract.py
```diff
@@ -22,7 +22,12 @@ class ContractAbstractContract(models.AbstractModel):
     partner_id = fields.Many2one(
         comodel_name="res.partner", string="Partner", index=True
     )
-    pricelist_id = fields.Many2one(comodel_name="product.pricelist", string="Pricelist")
+    pricelist_id = fields.Many2one(
+        comodel_name="product.pricelist",
+        string="Pricelist",
+        check_company=True,
```
MrTango, line 28: this leads to an issue with the pricelists in contract, when no multi companay is used.
I can no longer choose existing lists.
Before this change it was working fine.

contract/models/contract.py
```diff
@@ -81,13 +84,19 @@ class ContractContract(models.Model):
     )
     date_end = fields.Date(compute="_compute_date_end", store=True, readonly=False)
     payment_term_id = fields.Many2one(
-        comodel_name="account.payment.term", string="Payment Terms", index=True
+        comodel_name="account.payment.term",
+        string="Payment Terms",
+        index=True,
+        check_company=True,
+        domain="[('company_id', '=', company_id)]",
```
marcelsavegnago, line 91: @alexis-via do you believe it is necessary to define this domain? In this case, the objective is to force a company to be defined in each payment_term? In theory, check_company would already check whether the company is false or equal to the company in the record.

PR #826, opened 2022-05-11
[14.0][FIX][contract] Set uom as relevant parameter to calculate price unit in automatic price mode

MTantin:
Actually UOM is not used in calculation of price unit neither in automatic price mode.
I think this field is relevant to be used. If you change unit of measure, you want to show the price modified consequently.

### DIFF
--- a/contract/models/abstract_contract_line.py
+++ b/contract/models/abstract_contract_line.py
@@ -182,6 +182,7 @@ class ContractAbstractContractLine(models.AbstractModel):
         "quantity",
         "contract_id.pricelist_id",
         "contract_id.partner_id",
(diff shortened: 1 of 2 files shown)


rousseldenis (commented):
@MTantin Thanks for this. Could you add a test for this use case ?

MTantin:
> @MTantin Thanks for this. Could you add a test for this use case ?

Is it okay for you?

rousseldenis (approved):

MTantin:
@rousseldenis What's the next workprocess to insert this code to 14.0 branch?
Do you need anything more from me?

PR #831, opened 2022-05-15
[IMP] contract: Add contract lines entry point for reporting

rousseldenis:


carlosdauden (approved):

contract/views/contract_line.xml
```diff
@@ -193,4 +193,108 @@
...
+        <field name="domain">[('contract_id.contract_type', '=', 'purchase')]</field>
+        <field name="context">
+            {'search_default_group_by_contract': 1}
+        </field>
+        <field name="search_view_id" ref="contract_line_search_view" />
+        <field name="view_id" ref="contract_line_report_tree_view" />
+    </record>
+
+    <record id="action_customer_contract_lines" model="ir.actions.act_window">
+        <field name="name">Customer Contract Lines</field>
```
pedrobaeza, line 276: Why 2 separate actions instead of one like invoice analysis?

angelmoya (approved):

contract/views/contract_line.xml
rousseldenis: As there are two menus for supplier and customer contracts.

Some maybe want to give access to one or other.

pedrobaeza: OK

PR #832, opened 2022-05-16
[14.0][FIX] contract: fix compute_recurring_next_date for recurring_r…

mourad-ehm:
contract: fix compute_recurring_next_date for recurring_rule_type = "monthlylastday"
before fix  : 
When you change value of recurring_rule_type (ex you pass from "monthly" to  "monthlylastday" or "weekly") or you change recurring_invoicing_type (ex: from "post-paid" to "pre-paid") or recurring_interval (every 1 to every 2 "month") ==> the recurring_next_date remain unchanged.
ex:
recurring_next_date = 2022-07-01 and recurring_rule_type (from "monthly" to  "monthlylastday" ) ==> recurring_next_date = 2022-07-01
 After fix:
When you change value of recurring_rule_type (ex pass from "monthly" to  "monthlylastday" or "weekly") or you change recurring_invoicing_type (ex: from "post-paid" to "pre-paid") or recurring_interval (from every 1 to every 2 "month") ==> the recurring_next_date is updated based on this new parameters.
ex:
recurring_next_date = 2022-07-01 and recurring_rule_type (from "monthly" to  "monthlylastday" ) ==> recurring_next_date = 2022-06-30

### DIFF
--- a/contract/tests/test_contract.py
+++ b/contract/tests/test_contract.py
@@ -1814,7 +1814,7 @@ class TestContract(TestContractBase):
         self.assertEqual(last, to_date("2018-01-31"))
         self.assertEqual(recurring_next_date, to_date("2018-01-31"))
         self.assertEqual(self.acct_line.recurring_next_date, to_date("2018-01-31"))
-        # monthlylastday with match with post-paid
+        # monthlylastday must match with post-paid
         # because in pre-paid the invoice date must be de fist of month
         self.assertEqual(self.acct_line.recurring_invoicing_type, "post-paid")
         self.contract.recurring_create_invoice()
--- a/contract/tests/test_contract.py
+++ b/contract/tests/test_contract.py
@@ -293,8 +293,8 @@ class TestContract(TestContractBase):
     def test_contract_daily(self):
         recurring_next_date = to_date("2018-02-23")
         last_date_invoiced = to_date("2018-02-22")
-        self.acct_line.recurring_next_date = "2018-02-22"
         self.acct_line.recurring_rule_type = "daily"
+        self.acct_line.recurring_next_date = "2018-02-22"
         self.contract.pricelist_id = False
         self.contract.recurring_create_invoice()
         invoice_daily = self.contract._get_related_invoices()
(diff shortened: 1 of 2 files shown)


mourad-ehm:
ping @pedrobaeza @hparfr

contract/tests/test_contract.py
```diff
@@ -303,8 +303,8 @@ def test_contract_daily(self):
         self.assertEqual(self.acct_line.last_date_invoiced, last_date_invoiced)
 
     def test_contract_invoice_followers(self):
-        self.acct_line.recurring_next_date = "2018-02-23"
         self.acct_line.recurring_rule_type = "daily"
+        self.acct_line.recurring_next_date = "2018-02-23"
```
pedrobaeza, line 307: Why changing this order?

pedrobaeza (commented):
Can you depict the steps to reproduce the problem?

contract/tests/test_contract.py
mourad-ehm: The fix is to recompute  recurring_next_date when we change recurring_rule_type. So to force recurring_next_date we must set it after recurring_rule_type otherwise the date is not keeped.

pedrobaeza: Then this seems a problem in that code

mourad-ehm: I don't think so. recurring_next_date depends on several parameters and specially  date_start which is set to today by default. when we set recurring_rule_type to monthly or daily the recurring_next_date is recomputed based on all parameters. If we want to force this date we must set it last. The equivalent case (example) is the price list in sale order. If we choose a specific (special discount) price list in sale order, then we change client, price list is updated by the on-change. The order of data input is important if we want force field updated by on change.

pedrobaeza: But why now you need to change the order and not before? Something scratches here.

mourad-ehm: Because my fix concern the on change method

pedrobaeza: But this shouldn't have nothing to do with what you put on the title of the PR. Keep the diff minimum.

mourad-ehm: The order is important otherwise tests are not OK

pedrobaeza: But they were OK before your PR, so then what is modifying seems not OK, or at least not with an explanation of the steps to reproduce the problem. And the ideal situation is to not depend in any input order.

mourad-ehm: I don't know how to explain. Because my PR fix the onchange method. So it's logic that the value computed by this method is impacted !

mourad-ehm: @pedrobaeza could you validate this pr ?

mourad-ehm:
ping @hparfr

PR #834, opened 2022-05-23
[14.0][FIX] contract: add default_move_type in context

mariadforgeflow:
If `default_move_type` isn't in the context some views lack some information or are not correct.

For example the field` invoice_date` is not shown in the tree view  when accessing to the contract invoices' from the Contact view or from Invoicing/Customers/Customer Contracts. 

Also, if a new invoice line wants to be added in one of these invoices, only products that "Can be Purchased" appear because the product_id domain is:
```
context.get('default_move_type') in ('out_invoice', 'out_refund', 'out_receipt')
and [('sale_ok', '=', True), '|', ('company_id', '=', False), ('company_id', '=', parent.company_id)]
or [('purchase_ok', '=', True), '|', ('company_id', '=', False), ('company_id', '=', parent.company_id)]
```

### DIFF
--- a/contract/models/contract.py
+++ b/contract/models/contract.py
@@ -277,12 +277,20 @@ class ContractContract(models.Model):
         self.ensure_one()
         tree_view = self.env.ref("account.view_invoice_tree", raise_if_not_found=False)
         form_view = self.env.ref("account.view_move_form", raise_if_not_found=False)
+        ctx = dict(self.env.context)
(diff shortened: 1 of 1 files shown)


MiquelRForgeFlow (approved):

LoisRForgeFlow (approved):
Makes sense

MiquelRForgeFlow:
Forward port in https://github.com/OCA/contract/pull/835.

PR #838, opened 2022-06-03
[FIX] Set last date invoiced not readonly

angelmoya:
If you do wrong invoice on contract, you can not re invoice with right data...

For example, set recurring invoice type as post-paid, create first invoice, set next invoice to previous date to re invoice and then data is not right.

Any problem to unset readonly? Finally this field is visible only to technical user so I think no problem


rousseldenis:
@angelmoya Please review this one instead. https://github.com/OCA/contract/pull/810

As this is a critical field, that approach is better.

mourad-ehm (approved):

rousseldenis:
@mourad-ehm Could you look at #810 ?

pedrobaeza:
Closing, as #810 is being merged.

PR #839, opened 2022-06-05
[14.0] [IMP] Free invoice date

angelmoya:
Adds "Free Invoice Date" (boolean field) on contract.

If you set "Free Invoice Date" on contract, generated invoices will create without invoice date.

As invoice date is not required on invoice until you confirm the invoice you can set null, and it will autocomplete when you confirm invoice.

This way yo can, for example, generate invoice on contract monthly (from 1 to last day of the month), and validate few days later, after check it or complete with other date.

Other example, I invoice timesheets monthly (from 1 to last day of the month) with contract, but I need to check timesheet before invoice, so I disable cron, and launch manualy after check all timesheets. Before this modification, invoice was created with day 1.

### DIFF
--- a/contract/models/contract.py
+++ b/contract/models/contract.py
@@ -135,6 +135,9 @@ class ContractContract(models.Model):
         inverse_name="contract_id",
         string="Modifications",
     )
+    free_invoice_date = fields.Boolean(
+        string="Free Invoice Date",
+    )
 
     def get_formview_id(self, access_uid=None):
         if self.contract_type == "sale":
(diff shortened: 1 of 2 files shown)


rousseldenis:
@angelmoya Thanks for this.

In order to not overcharge contract module anymore, I propose you to create a new module for that functionality.

Best way to identify clearly what it does, minimize the tests on it (By the way could you add a test for generation ?)

Thanks

ikapasi-initos (commented):
Code review 
LGTM

ikapasi-initos (approved):

pegonzalezs:
@angelmoya you can close this as it has been moved to #855 following the suggestions from @rousseldenis

PR #840, opened 2022-06-12
[14.0][FIX] agreement_legal_sale - ignore sale order lines with display_type

ghost:
When confirming a sale order with an agreement template and the order has some lines that are not product lines, you have an error message about missing uom because it tries to create lines for the note/section lines

S00005 - Odoo

### DIFF
--- a/agreement_legal_sale/models/sale_order.py
+++ b/agreement_legal_sale/models/sale_order.py
@@ -27,7 +27,7 @@ class SaleOrder(models.Model):
                         or False,
                     }
                 )
-                for line in order.order_line:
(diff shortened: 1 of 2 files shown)


rousseldenis:
@jdidderen-be Thanks for this. Could you add lines with display_type in action_confirm test in order to avoid regression ?

ghost:
@rousseldenis Should be good now

rousseldenis (approved):
Code review.

@jdidderen-be

ikapasi-initos (approved):
Code review and functional
LGTM

PR #844, opened 2022-06-18
[13.0][ADD] copier

rousseldenis:


pedrobaeza:
We can't see the results until the files are on main branch, so that's why I usually commit directly and then solve problems in a later PR. Merging this directly.

rousseldenis:
Too fast !

Wrong directory

pedrobaeza:
Oh, I see! Sorry.

PR #846, opened 2022-06-18
[13.0][ADD] copier

rousseldenis:


pedrobaeza:
GH actions are being executed even in the PR. That didn't work before. I think what doesn't work is to merge with ocabot.

rousseldenis:
> I think what doesn't work is to merge with ocabot.

Yes, wait a bit. I'll merge when it's ok

rousseldenis:
@pedrobaeza I have promoted two modules but that one depends on a Beta.

In order to merge this we need to set it back to Beta. Are you ok with that ?

image

pedrobaeza:
OK. We'll promote later the main one to `stable`, being there for several versions and with tests.

PR #841, opened 2022-06-16
[14.0][FIX] contract/contract_sale_generation: Compute correctly the recurring next date after stop()

rousseldenis:

PR #776, opened 2022-01-26
[15.0][FWD] contract : PR #741 #789 #790 from 14.0 to 15.0

rousseldenis:

PR #847, opened 2022-06-21
[MIG] contract_sale_generation: Migration to 15.0

ntsirintanis:


contract_sale_generation/tests/test_contract_sale.py
```diff
@@ -0,0 +1,114 @@
...
+        self.assertEqual(set(action["domain"][0][2]), set(orders.ids))
+
+    def test_cron_recurring_create_sale(self):
+        self.contract_line.date_start = "2020-01-01"
+        self.contract_line.recurring_invoicing_type = "post-paid"
+        self.contract_line.date_end = "2020-03-15"
+        self.contract_line._onchange_is_auto_renew()
+        contracts = self.contract2
+        for _i in range(10):
+            contracts |= self.contract.copy({"contract_type": "sale"})
```
ntsirintanis: Currently tests fail at this one line. Error:

`contract/models/contract_line.py", line 452, in _check_recurring_next_date_start_date
    raise ValidationError(
odoo.exceptions.ValidationError: You can't have a date of next invoice anterior to the start of the contract line 'Test Contract Template'
`
Will need to check this more. @NL66278 @thomaspaulb if you have any ideas, please shoot.

thomaspaulb: @ntsirintanis I tried to follow what happens, and somehow it's a problem with the order of the evaluation of computations.

1. The test contract line is set to certain values, among which `date_start = 2020-01-01`
2. The contract including this line is copied 10 times
3. During copying of the contract, `recurring_next_date` of the lines gets recomputed correctly (2020-01-15)
4. Something happens and `recurring_next_date` is now `2020-01-01`, which is lower than 2020-01-15 (maybe recomputation, or goes back to cached value)
5. The computed values are checked and it finds 2020-01-01 < 2020-01-15 and borks out

I find that this stops happening when you access `self.contract_line.recurring_next_date` (by printing, or just like that) before copying, so between line 98 and 99.

Somehow equivalently, @rousseldenis found in [this PR](https://github.com/OCA/contract/pull/841/files#diff-9b7fa04a3606eef1c4322f66d703a6092d2b4bbd783b6b738c9b779dff22731dR106) that he had to call `_compute_recurring_next_date()`, however that specifically does not work for me.

thomaspaulb: There's another failing test on `import odoo.odoo.tests` in another module, I don't know how that got through CI, but it's clearly from and should be `odoo.tests`

rousseldenis:
@ntsirintanis Thanks for this.

Don't forget to make your PR's title a little bit understandable (e.g.: like your last commit)

contract_sale_generation/tests/test_contract_sale.py
rousseldenis: @thomaspaulb We need first to fix the tests import errors, then, check which test(s) are failing. @ntsirintanis

contract_sale_generation/models/contract.py
```diff
@@ -123,13 +123,10 @@ def _recurring_create_sale(self, date_ref=False):
         self._compute_recurring_next_date()
         return sale_orders
 
-    @api.model
-    def _get_recurring_create_func(self, create_type="invoice"):
```
rousseldenis: @ntsirintanis Why removing this ?

contract_sale_generation/models/abstract_contract.py
```diff
@@ -2,16 +2,10 @@
...
+from odoo import fields, models
 
 
 class ContractAbstractContract(models.AbstractModel):
     _inherit = "contract.abstract.contract"
 
-    sale_autoconfirm = fields.Boolean(string="Sale Autoconfirm")
-
-    @api.model
-    def _selection_generation_type(self):
```
rousseldenis: Why removing this ?

contract_sale_generation/models/contract.py
ntsirintanis: Because it was  removed from contract module in version 15.0, along with the generation_type field and all related methods.

contract_sale_generation/models/abstract_contract.py
ntsirintanis: Same as my previous comment.

contract_sale_generation/models/contract.py
rousseldenis: Mmmh, that's because migration PR for 15 was created before those changes.

I'll do forwardport of changes in contract since. Wait for it

rousseldenis: Arf, already done.

Could you review https://github.com/OCA/contract/pull/776 ?

contract_sale_generation/tests/test_contract_sale.py
```diff
@@ -96,6 +96,11 @@ def test_cron_recurring_create_sale(self):
         self.contract_line.recurring_invoicing_type = "post-paid"
         self.contract_line.date_end = "2020-03-15"
         self.contract_line._onchange_is_auto_renew()
+        # If we do not recompute recurring_next_date
+        # then it maintains it's 'old' value.
+        # TODO: Research that
+        recurring_next_date = self.contract_line.recurring_next_date
+        self.assertGreaterEqual(recurring_next_date, self.contract_line.date_start)
```
ntsirintanis, line 98: Added these 2 lines in order to make unit tests work, and pre-commit to not complain. Need to see why this is happening though.

contract_sale_generation/tests/test_contract_sale.py
```diff
@@ -96,6 +96,11 @@ def test_cron_recurring_create_sale(self):
         self.contract_line.recurring_invoicing_type = "post-paid"
         self.contract_line.date_end = "2020-03-15"
         self.contract_line._onchange_is_auto_renew()
+        # If we do not recompute recurring_next_date
```
rousseldenis, line 94: Yes, IMHO there are still some cases that should be solved after theses changes : https://github.com/OCA/contract/commit/cd086ddbb4a9b85ccfcac288e53fb93e716d9446

NL66278:
@ntsirintanis I think if a test is added for a contract with payment terms and a fiscal position, the code will be nearly completely covered and you will get rid of the last two red crosses.

contract_sale_generation/models/contract_line.py
```diff
@@ -0,0 +1,55 @@
+# Copyright (C) 2020 Open Source Integrators
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class ContractLine(models.Model):
+    _inherit = "contract.line"
+
+    def _prepare_sale_line_vals(self, dates, order_id=False):
```
NL66278, line 10: What is the use of the `order_id` parameter and logic? It seems to be called nowhere with `order_id=True`.

contract_sale_generation/models/contract_line.py
```diff
@@ -0,0 +1,55 @@
...
+        if order_id:
+            sale_line_vals["order_id"] = order_id.id
+        return sale_line_vals
+
+    def _prepare_sale_line(self, order_id=False, sale_values=False):
+        self.ensure_one()
+        dates = self._get_period_to_invoice(
+            self.last_date_invoiced, self.recurring_next_date
+        )
+        sale_line_vals = self._prepare_sale_line_vals(dates, order_id)
```
NL66278: Call keyword parm with keyword: `order_id=order_id`

NL66278: Actually `order_id` is nowhere passed to `_prepare_sale_line` as far as I can see. So logic and argument can maybe removed from this function and from `_prepare_sale_line_vals`. Can anyoint point out a use for thus argument?

contract_sale_generation/tests/common.py
```diff
@@ -0,0 +1,132 @@
...
+# Copyright 2017 Pesol (<http://pesol.es>)
+# Copyright 2017 Angel Moya <angel.moya@pesol.es>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from freezegun import freeze_time
+
+from odoo import fields
+from odoo.tests import Form
+
+
+def to_date(date):
```
NL66278, line 11: This one is not uses and can be removed

rousseldenis: No, but should be in each test line that uses fields.Date.to_date(

contract_sale_generation/models/contract_line.py
rousseldenis: I think this is intended in order to override and force an existing order

NL66278: But if you add sale lines to an existing sale order, through a write on that order, order_id will be filled automagically. Or if somehow some code wants to use the prepared values to create the order lines, the order_id can be inserted in that code. On the other hand this is existing code and we should be carefull with changing any method signature. I was just wondering about the usecase...

NL66278 (approved):
:+1: LGTM

ntsirintanis:
This PR is as green as it gets. I added some superfluous unit tests to satisfy codecov,  but can't do much beyond that. Please review.

rousseldenis (approved):

NL66278:
@ntsirintanis I already approved the PR. In the CodeCov there are still three red lines:
https://app.codecov.io/gh/OCA/contract/compare/847/diff
You might get rid of them (if you want) by testing:
- That a contract without lines does not generate a sale order;
- That if you pass an existing sale order the prepare line vals function, it fills the order_id;
- By testing the main cron function.

I do not know how the yellow lines are counted in the percentage. It might be very difficult to get rid of those.

ntsirintanis:
> @ntsirintanis I already approved the PR. In the CodeCov there are still three red lines: https://app.codecov.io/gh/OCA/contract/compare/847/diff You might get rid of them (if you want) by testing:
> 
>     * That a contract without lines does not generate a sale order;
> 
>     * That if you pass an existing sale order the prepare line vals function, it fills the order_id;
> 
>     * By testing the main cron function.
> 
> 
> I do not know how the yellow lines are counted in the percentage. It might be very difficult to get rid of those.

Agreed. I've been already working on points one and two. Will finish them, push, and hope for absolute greenness.

rousseldenis:
@ntsirintanis Is this ready ?

ntsirintanis:
> @ntsirintanis Is this ready ?

yes, finally, we are green!

rousseldenis:
@ntsirintanis Could you check at this ? https://github.com/OCA/contract/pull/841

I think this is related to your problem with recurring_next_date value.

ntsirintanis:
> @ntsirintanis Could you check at this ? #841
> 
> I think this is related to your problem with recurring_next_date value.
thanks @rousseldenis , will check it out

PR #851, opened 2022-06-23
[15.0][IMP] contract: add tree/form views for contract.modification

Olageibol:
Now the modifications page in contract form appears like that: Captura de pantalla de 2022-06-23 17-40-25
Editable="bottom" Attribute is not working https://github.com/OCA/contract/blob/15.0/contract/views/contract.xml#L422
After this PR, the 'modifications' model fields appear in the form:
Captura de pantalla de 2022-06-23 17-57-16

### DIFF
--- a/contract/__manifest__.py
+++ b/contract/__manifest__.py
@@ -39,6 +39,7 @@
         "views/contract_tag.xml",
         "views/abstract_contract_line.xml",
         "views/contract.xml",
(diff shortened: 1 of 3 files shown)


pedrobaeza:
Why is not working? Better to just fix the editable feature

rousseldenis:
> Why is not working? Better to just fix the editable feature

I would have said same

PR #852, opened 2022-06-28
[14.0][FIX] contract: Get correct domain for In Progress contract lines

rousseldenis:
As termination_notice_date field is void on not finished contract lines,
domain to search for In Progress ones should take into account void values


rousseldenis:
Take also into account section lines.

pedrobaeza:
I think this is fixed in a simpler way in #853

max3903 (approved):

PR #853, opened 2022-07-01
[13.0][FIX] contract: Reset next invoicing date when terminating a line

pedrobaeza:
**Steps to reproduce the problem:**

- Create a new contract
- Set "Recurrence at line level?"
- Add a new line, with following data:

  * Invoice Every: 1 Month(s)
  * Date Start: 2022-06-01
  * Invoicing type: Pre-paid
- Save and click on "Create invoices" button
- Now edit again and put "Date End": 2022-06-30

**Current behavior:**

The line is still invoiceable (appears in blue), and if you click on "Create invoices" button, you get the traceback:

```
  ...
  File "/mnt/data/odoo-addons-dir/contract/models/contract.py", line 534, in recurring_create_invoice
    invoice = self._recurring_create_invoice()
  File "/mnt/data/odoo-addons-dir/contract/models/contract.py", line 561, in _recurring_create_invoice
    invoices_values = self._prepare_recurring_invoices_values(date_ref)
  File "/mnt/data/odoo-addons-dir/contract/models/contract.py", line 516, in _prepare_recurring_invoices_values
    invoice_line_vals = line._prepare_invoice_line(move_form=move_form)
  File "/mnt/data/odoo-addons-dir/contract_layout_category_hide_detail/models/contract_line.py", line 12, in _prepare_invoice_line
    vals = super()._prepare_invoice_line(move_form)
  File "/mnt/data/odoo-addons-dir/product_contract/models/contract_line.py", line 22, in _prepare_invoice_line
    res = super(ContractLine, self)._prepare_invoice_line(move_form)
  File "/mnt/data/odoo-addons-dir/contract_variable_quantity/models/contract_line.py", line 44, in _prepare_invoice_line
    vals = super()._prepare_invoice_line(move_form)
  File "/mnt/data/odoo-addons-dir/contract/models/contract_line.py", line 545, in _prepare_invoice_line
    name = self._insert_markers(dates[0], dates[1])
  File "/mnt/data/odoo-addons-dir/contract/models/contract_line.py", line 595, in _insert_markers
    name = name.replace("#END#", last_date_invoiced.strftime(date_format))
AttributeError: 'bool' object has no attribute 'strftime'
```

**Expected behavior:**

Line is not invoiceable anymore, and "Create invoices" button disappears (as no invoiceable line - but anyways, calling the method the same doesn't trigger the invoice creation).

@Tecnativa TT37880

PR #854, opened 2022-07-04
[14.0][FIX] contract: Reset next invoicing date when terminating a line

pedrobaeza:
Forward-port of #853 

Steps to reproduce the problem:

- Create a new contract
- Set "Recurrence at line level?"
- Add a new line, with following data:

  * Invoice Every: 1 Month(s)
  * Date Start: 2022-06-01
  * Invoicing type: Pre-paid
- Save and click on "Create invoices" button
- Now edit again and put "Date End": 2022-06-30

Current behavior:

The line is still invoiceable (appears in blue), and if you click on "Create invoices" button, you get the traceback:

```
  ...
  File "/mnt/data/odoo-addons-dir/contract/models/contract.py", line 534, in recurring_create_invoice
    invoice = self._recurring_create_invoice()
  File "/mnt/data/odoo-addons-dir/contract/models/contract.py", line 561, in _recurring_create_invoice
    invoices_values = self._prepare_recurring_invoices_values(date_ref)
  File "/mnt/data/odoo-addons-dir/contract/models/contract.py", line 516, in _prepare_recurring_invoices_values
    invoice_line_vals = line._prepare_invoice_line(move_form=move_form)
  File "/mnt/data/odoo-addons-dir/contract_layout_category_hide_detail/models/contract_line.py", line 12, in _prepare_invoice_line
    vals = super()._prepare_invoice_line(move_form)
  File "/mnt/data/odoo-addons-dir/product_contract/models/contract_line.py", line 22, in _prepare_invoice_line
    res = super(ContractLine, self)._prepare_invoice_line(move_form)
  File "/mnt/data/odoo-addons-dir/contract_variable_quantity/models/contract_line.py", line 44, in _prepare_invoice_line
    vals = super()._prepare_invoice_line(move_form)
  File "/mnt/data/odoo-addons-dir/contract/models/contract_line.py", line 545, in _prepare_invoice_line
  File "/mnt/data/odoo-addons-dir/contract/models/contract_line.py", line 595, in _insert_markers
AttributeError: 'bool' object has no attribute 'strftime'
```

Expected behavior:

Line is not invoiceable anymore, and "Create invoices" button disappears (as no invoiceable line - but anyways, calling the method the same doesn't trigger the invoice creation).

This reverts part of the fixes done by @rousseldenis to apply this simpler one and for not deforming the aim of the method, so no pending fixme.

@Tecnativa TT37880

### DIFF
--- a/contract/models/contract_line.py
+++ b/contract/models/contract_line.py
@@ -611,12 +611,19 @@ class ContractLine(models.Model):
         return name
 
     def _update_recurring_next_date(self):
-        # FIXME: Change method name according to real updated field
-        # e.g.: _update_last_date_invoiced()
         for rec in self:
             last_date_invoiced = rec.next_period_date_end
+            recurring_next_date = rec.get_next_invoice_date(
+                last_date_invoiced + relativedelta(days=1),
+                rec.recurring_invoicing_type,
+                rec.recurring_invoicing_offset,
+                rec.recurring_rule_type,
+                rec.recurring_interval,
+                max_date_end=rec.date_end,
+            )
             rec.write(
                 {
+                    "recurring_next_date": recurring_next_date,
                     "last_date_invoiced": last_date_invoiced,
                 }
             )
--- a/contract/models/contract_recurrency_mixin.py
+++ b/contract/models/contract_recurrency_mixin.py
@@ -95,7 +95,8 @@ class ContractRecurrencyMixin(models.AbstractModel):
 
     @api.depends("next_period_date_start")
     def _compute_recurring_next_date(self):
-        for rec in self:
+        self.recurring_next_date = False
+        for rec in self.filtered("next_period_date_start"):
             rec.recurring_next_date = self.get_next_invoice_date(
                 rec.next_period_date_start,
                 rec.recurring_invoicing_type,
--- a/contract/tests/test_contract.py
+++ b/contract/tests/test_contract.py
@@ -2401,3 +2401,11 @@ class TestContract(TestContractBase):
         action = self.contract.action_preview()
         self.assertIn("/my/contracts/", action["url"])
         self.assertIn("access_token=", action["url"])
+
+    def test_contract_line_termination(self):
+        """Don't fail when the line receives an end date."""
+        self.contract.recurring_create_invoice()
+        self.acct_line.date_end = "2018-02-14"
+        self.assertFalse(self.acct_line.recurring_next_date)
+        # This doesn't give any error
+        self.contract.recurring_create_invoice()


contract/models/contract_line.py
```diff
@@ -611,12 +611,19 @@ def _insert_markers(self, first_date_invoiced, last_date_invoiced):
...
+                last_date_invoiced + relativedelta(days=1),
+                rec.recurring_invoicing_type,
+                rec.recurring_invoicing_offset,
+                rec.recurring_rule_type,
+                rec.recurring_interval,
+                max_date_end=rec.date_end,
+            )
             rec.write(
                 {
+                    "recurring_next_date": recurring_next_date,
```
rousseldenis, line 626: IMHO This should never happen as the field is computed...

It's quite insane...

pedrobaeza: You're right. It should be `update`, not `write`, but the aim is to update such data.

pedrobaeza: Oh, wait, let me see, as this is not a computed method.

pedrobaeza: This is only called here, when preparing the next invoice:

https://github.com/OCA/contract/blob/15209ddf952625534d6464671d9cf8da785cfd6f/contract/models/contract.py#L529

so no computed method involved, and I consider it correct.

rousseldenis: That's exactly what I wanted to say.

This is a method called outside of computes.

rousseldenis: This was an historic method (the naming) before the change to computed field.

The only field that needs update is the last_invoice_date

pedrobaeza: Then what's the problem calling `write`? Remember that this uses `compute` as onchanges through the computed writable fields, so it's not a traditional compute field.

rousseldenis: > Remember that this uses `compute` as onchanges through the computed writable fields, so it's not a traditional compute field

In any case, the field should only be updated through its compute method

pedrobaeza: @rousseldenis the switch to `compute` was for removing onchanges, but they are traditional regular fields with such trigger for updating the field in certain cases, but that doesn't prevent that they can update it via other means, and these ways are when `depends` syntax can't reflect this change, like the one when creating the invoices. Do you understand?

pedrobaeza: > In any case, the field should only be updated through its compute method

We are crossing our answers, but I have replied to this on the previous one. Please check it and tell me any doubt.

rousseldenis: I understand, but IMHO that's an entry point that should never be used.

You can (and we had several cases here) have some troubles with cache.

If a field is computed, it should never be update elsewhere than its compute method.

pedrobaeza: Then I'm afraid you are not understanding what a computed writable field is. The compute method is just auxiliary, and replace the onchanges. You can (and in a lot of cases, you must) update the fields directly in other methods.

rousseldenis: Do you have a pointer where Odoo officially explain that ?

I have only this link : https://www.odoo.com/documentation/14.0/fr/developer/reference/addons/orm.html#reference-fields-compute

rousseldenis: That consideration apart, I think updating only the last_invoice_date should be sufficient to have the good recurring_next_date value.

pedrobaeza: It's explained here: https://youtu.be/OMZbbWrSsfw?t=1158

I'm not sure such date is updated through the `depends` chain, but we can try adding specific tests for such cases. Anyways, we can put update and the update is delayed until the end, avoiding double writing if you prefer.

rousseldenis: You have it there : https://github.com/OCA/contract/blob/14.0/contract/models/contract_line.py#L116

rousseldenis: > It's explained here: https://youtu.be/OMZbbWrSsfw?t=1158

As skipped v13, missed the presentation :sweat_smile:  Ok, I'll try to do a PR to update documentation about that.

rousseldenis: @pedrobaeza I think we have all the needed chain to let the computation works by itself.

The entry point to change it should be the last_invoice_date IMHO.

As for example here if we want to do it manually : https://github.com/OCA/contract/pull/810

victoralmau (approved):

CRogos:
I was testing this fix, but still get the same error message. After archiving all contracts with "Date End" < now (Date of next invoice)  I was able to generate the invoices.

rousseldenis:
> I was testing this fix, but still get the same error message. After archiving all contracts with "Date End" < now (Date of next invoice) I was able to generate the invoices.

@CRogos Could you try this ? https://github.com/OCA/contract/pull/841

CRogos:
Still the same error message. I assume there is a problem when the "Date End" is in the past, but the contract is still selected for invoice creation?
image

CRogos:
I've now retested this PR and also #841 with a new database. While #841 worked, this PR did not.
I tested as following:
Today is the 15.9.2022

- Create a new contract
  - Date Start 1.6.2022
  - Date End 31.7.2022
  - Invoice type: pre-paid
  - invoice every: 1 Month
- Add a new line, with a product, no End date.

When I now press "Create Invoices" 3 times, than the first invoice is created successfully, also the second, but can see that the Date of next Invoice is changed to the Start Date again:
image

When pressing the "Create Invoice" the third time we get an error.
image

CRogos:
@pedrobaeza are you still working on this?

rafaelbn:
Please @pedrobaeza could you rebase?

rafaelbn:
/ocabot rebase

🤞

rafaelbn (approved):
My logic is:

This PR was merged 4th July 2022 for 13 https://github.com/OCA/contract/pull/853

Why not for 14?

pedrobaeza:
Closing in favor of #841

PR #862, opened 2022-07-27
[15.0][IMP] contract : show tags with colors in the contract

chandni299:


marylla (approved):
I made a functional review. LGTM. :+1: 

I can see the colored tags in form and tree view.

yweng8111 (approved):

gregory-moka:
Tags with no color (hidden in kanban view) are always displayed in the list view (just to be consistent with the whole solution)

PR #863, opened 2022-08-05
[13.0][MIG] contract_sale_generation

marionumza:


rousseldenis:
Thanks for this. Could you preserve commit history as explained here : 
https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-13.0

You can follow the full process (inversing source and target branches in git format-patch command)

PR #866, opened 2022-08-10
[IMP] contract: Add contract date

rousseldenis:
In some cases, the contract date (e.g.: signature date) can be different from the
start date (e.g.: signature on 05/15/2022 and start date on 06/01/2022).

It can have importance in case of prices revisions from a legal point of view as
this is the contract date that has to be taken into account.


rafaelbn (approved):
👍

carlosdauden (approved):

PR #870, opened 2022-08-28
Update contract_portal_templates.xml

ghost:
[FIX] contract: Proper anonymous avatar portal path


pedrobaeza:
Thanks for the fix. Can you please change the commit message to: `[FIX] contract: Proper anonymous avatar portal path`?

PR #871, opened 2022-08-31
Improve report presentation

gregory-moka:
- Remove table presentation (weird display depending on default report style)
- Add the products and services section if set


rvalyi:
would be great if you could attach 2 prints: before and after...

gregory-moka:
> would be great if you could attach 2 prints: before and after...

- [Before.pdf](https://github.com/OCA/contract/files/9465632/Before.pdf)
- [After.pdf](https://github.com/OCA/contract/files/9465631/After.pdf)

PR #877, opened 2022-09-21
[14.0][FIX] agreement_legal: fix action open agreement

ps-tubtim:
This PR fixes **Access Error** when click smart button **Agreements** on Contact Form View.

Selection_121

### DIFF
--- a/agreement_legal/models/res_partner.py
+++ b/agreement_legal/models/res_partner.py
@@ -22,10 +22,13 @@ class Partner(models.Model):
 
     def action_open_agreement(self):
         self.ensure_one()
-        action = self.env.ref("agreement.agreement_action")
-        result = action.read()[0]
-        result["domain"] = [("partner_id", "=", self.id)]
-        result["context"] = {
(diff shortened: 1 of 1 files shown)


agreement_legal/models/res_partner.py
```diff
@@ -23,7 +23,7 @@ def _compute_agreements_count(self):
     def action_open_agreement(self):
         self.ensure_one()
         action = self.env.ref("agreement.agreement_action")
-        result = action.read()[0]
+        result = action.sudo().read()[0]
```
pedrobaeza: A shorter alternative:

```suggestion
        action = self.env["ir.actions.actions"]._for_xml_id("agreement.agreement_action")
```

ps-tubtim: Thank you. I already fixed it.

rousseldenis (approved):

rousseldenis:
@ps-tubtim Could you check pre-commit ?

PR #883, opened 2022-10-18
[14.0][IMP] contract: Set invoice_user_id field with Form() to apply the correct values of other fields.

victoralmau:
Set `invoice_user_id` field with `Form()` to apply the correct values of other fields (`team_id` for example).

@Tecnativa TT39934

### DIFF
--- a/contract/models/contract.py
+++ b/contract/models/contract.py
@@ -436,6 +436,8 @@ class ContractContract(models.Model):
             move_form.invoice_payment_term_id = self.payment_term_id
         if self.fiscal_position_id:
             move_form.fiscal_position_id = self.fiscal_position_id
+        if invoice_type == "out_invoice" and self.user_id:
+            move_form.invoice_user_id = self.user_id
         invoice_vals = move_form._values_to_save(all_fields=True)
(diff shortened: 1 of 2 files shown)


contract/models/contract.py
```diff
@@ -436,6 +436,8 @@ def _prepare_invoice(self, date_invoice, journal=None):
             move_form.invoice_payment_term_id = self.payment_term_id
         if self.fiscal_position_id:
             move_form.fiscal_position_id = self.fiscal_position_id
+        if self.user_id:
```
pedrobaeza: Do it only for sales invoices

rousseldenis (commented):
@victoralmau Thanks for this. Could you improve tests to highlight this change ?

victoralmau:
> @victoralmau Thanks for this. Could you improve tests to highlight this change ?

I have improved test a bit, but I can't validate `team_id` field because it is added with `sales` module.

rousseldenis (approved):

pedrobaeza (approved):

carlosdauden (approved):

Issue #880, opened 2022-10-03
Migration to version 16.0


sbidoul:
# Todo

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-16.0

# Modules to migrate

- [x] agreement_rebate_partner_company_group - By @stefan-tecnativa - #980
- [x] contract - By @wpichler - #887
- [ ] contract_invoice_start_end_dates - By @JesusValdez96 - #950
- [x] contract_mandate - By @adriapalleja - #1070
- [x] contract_payment_auto - By @9184Lucky - #1231
- [x] contract_payment_mode - By @carolinafernandez-tecnativa - #996
- [ ] contract_price_revision - By @houzefa-abba - #1038
- [x] contract_queue_job - By @luisDIXMIT - #1077
- [ ] contract_sale - By @CRogos - #949
- [ ] contract_sale_generation - By @JesusValdez96 - #955
- [x] contract_update_last_date_invoiced - By @edlopen - #1147
- [x] contract_variable_qty_timesheet - By @danypr92 - #1110
- [x] contract_variable_quantity - By @carolinafernandez-tecnativa - #1027
- [x] product_contract - By @Rad0van - #959
- [x] subscription_oca - By @carolina-domatix - #1029

Missing module? Check https://github.com/OCA/maintainer-tools/wiki/%5BFAQ%5D-Missing-modules-in-migration-issue-list

jado95:
Hi, it is possible update module contract_sale with #915?

JesusValdez96:
Please add the missing module contract_invoice_start_end_dates to the list.
PR here: https://github.com/OCA/contract/pull/950

CRogos:
Continuing contract_sale in #949

maisim:
Hi !
I tried to backport product_contract from 17 (https://github.com/OCA/contract/pull/1090) rather than migrate from 15 (https://github.com/OCA/contract/pull/959), thinking it would be more up to date. This is done quite easily. Does this suit you in principle or would you prefer that we continue on the migration from 15.0 (https://github.com/OCA/contract/pull/959) ?

cc @Rad0van what do you think?

Rad0van:
> Hi !
[...]

PR #887, opened 2022-10-24
[16.0][MIG] contract: Migration to 16.0

wpichler:
https://github.com/OCA/contract/issues/880


contract/views/res_partner_view.xml
```diff
@@ -5,14 +5,14 @@
     <record id="view_partner_form" model="ir.ui.view">
         <field name="inherit_id" ref="base.view_partner_form" />
         <field name="model">res.partner</field>
-        <field name="groups_id" eval="[(4, ref('account.group_account_invoice'))]" />
         <field type="xml" name="arch">
             <xpath expr="//div[@name='button_box']" position="inside">
-                <field name="customer_rank" invisible="1" />
+                <field name="customer_rank" invisible="1" groups="account.group_account_invoice"/>
```
rousseldenis: why adding a group on invisible field ?

contract/models/contract.py
```diff
@@ -560,10 +560,10 @@ def _prepare_recurring_invoices_values(self, date_ref=False):
                     # nullifying line. We should then cleanup certain values.
                     del invoice_line_vals["company_id"]
                     del invoice_line_vals["company_currency_id"]
-                    invoice_vals["invoice_line_ids"].append((0, 0, invoice_line_vals))
+                    invoice_vals["invoice_line_ids"].append(Command.create(invoice_line_vals))
             invoices_values.append(invoice_vals)
             # Force the recomputation of journal items
-            del invoice_vals["line_ids"]
+            # del invoice_vals["line_ids"] # TODO: Check why line_ids is missing here
```
rousseldenis: Please check this, before approving

wpichler:
One check is still not working - but i assume this is an error on odoo side - look at: https://github.com/odoo/odoo/issues/104009

I could need help a the pre-commit steps - i don't get the point what should be wrong here...

pedrobaeza:
Please cherry-pick eda6d54d63b055046b3044abe40e332511a02978

pedrobaeza:
You can integrate #894 as well. Can you please check the tests in the CI?

wpichler:
Done. There is only one test which fails - and this test does fail because of a bug in odoo core (imho). I can disable this test for now....

Here is my bug report for this odoo bug
https://github.com/odoo/odoo/issues/104009

wpichler:
I did take an other look on the failed test - and i don't get the point what is wrong here. In the test test_cron_recurring_create_invoice_multi_company it just creates the first account.move Form - this does try to access the member name - which is a compute field - so the compute functions triggers with a NewId Object - but in the compute function it does try to compute the name for all the NewId Objects from the past as it seems. And this does fail, because it can't compare agains't NewId.

Why are there that many NewId records in the compute function ? 

Can we add some sort of flush / invalidate before the test to get rid of the NewId Objects ? Why are they don't get closed on the other tests ?

contract/views/contract_template.xml
```diff
@@ -0,0 +1,133 @@
...
+                            options="{'no_create': True}"
+                            groups="base.group_multi_company"
+                        />
+                    </group>
+                </group>
+                <group name="group_invoice_lines" string="Contract Template Lines">
+                    <field
+                        name="contract_line_ids"
+                        widget="section_and_note_one2many"
+                        nolabel="1"
```
gkliska, line 28: colspan="2"
Looks much better taking whole form width.

victoralmau:
Please cherry-pick https://github.com/OCA/contract/pull/902

wpichler:
@pedrobaeza There is just one test failing - and i am not able to fix it - because i think that it is because of a bug in the odoo account module directly. See https://github.com/odoo/odoo/issues/104009

Should i disable this test so that we can merge this ?

contract/models/contract_line.py
```diff
@@ -35,10 +35,6 @@ class ContractLine(models.Model):
         string="Analytic account",
         comodel_name="account.analytic.account",
     )
-    analytic_tag_ids = fields.Many2many(
```
rousseldenis: If you remove this, you should inherit from analytic.mixin instead

contract/models/contract_line.py
```diff
@@ -548,24 +545,22 @@ def _prepare_invoice_line(self, move_form):
             self.last_date_invoiced, self.recurring_next_date
         )
         line_form = move_form.invoice_line_ids.new()
-        line_form.display_type = self.display_type
+        line_form.display_type = self.display_type or "product"
         line_form.product_id = self.product_id
         invoice_line_vals = line_form._values_to_save(all_fields=True)
         name = self._insert_markers(dates[0], dates[1])
+        # TODO: We lost the analytic_account_id on invoice line vals
```
rousseldenis: See above

rousseldenis (changes requested):

contract/models/contract.py
```diff
@@ -0,0 +1,716 @@
...
+                )
+                % {
+                    "contract_type": self.contract_type,
+                    "company": self.company_id.name or "",
+                }
+            )
+        invoice_type = "out_invoice"
+        if self.contract_type == "purchase":
+            invoice_type = "in_invoice"
+        move_form = Form(
```
rousseldenis: Not wanting to push a new debate on that, I'm convinced this is not a correct solution

pedrobaeza: I'm not as you know. With this, we assure maximum compatibility. Even with `onchange_helper`, you won't get the same compatibility. I know it's not the most efficient one, but now on v16, the performance gets better.

pedrobaeza:
> There is just one test failing - and i am not able to fix it - because i think that it is because of a bug in the odoo account module directly. See [odoo/odoo#104009](https://github.com/odoo/odoo/issues/104009)
> 
> Should i disable this test so that we can merge this ?

I think we should push for the solution. It has been already tagged, so I expect a response from Odoo soon. You can also push through an enterprise ticket.

About the analytic thing, you should convert it to the new format.

contract/models/contract_line.py
wpichler: Ok - will inherit from analytic.mixin

contract/models/contract.py
wpichler: Did checked this - its ok - they are not needed here

contract/views/res_partner_view.xml
wpichler: Can't remember why i did added this - but seems senseless - will remove it

contract/models/contract_line.py
wpichler: done

flotho:
hi @rousseldenis any chance to have a final review ?

contract/models/contract_modification.py
```diff
@@ -0,0 +1,38 @@
...
+    date = fields.Date(required=True)
+    description = fields.Text(required=True)
+    contract_id = fields.Many2one(
+        string="Contract",
+        comodel_name="contract.contract",
+        required=True,
+        ondelete="cascade",
+        index=True,
+    )
+    sent = fields.Boolean(default=False)
```
DucTruongKomit: we don't need to set default=False for boo, because it will be False by default.

pedrobaeza:
#894 should be cherry-picked as well

pedrobaeza:
Another investigation line is that now invoices can't be created through `Form`, but through pure code, like it has been done in https://github.com/OCA/commission/pull/395/commits/2cd32a669ec6f2912bb0d26da5f3da1b00c9b8b9

tate11:
WARNING
odoo.fields: Field contract.line.price_subtotal: unknown parameter 'digits', if this is an actual parameter you may want to override the method _valid_field_parameter on the relevant model in order to allow it 

digits="Account",

wpichler:
@pedrobaeza It does look good now - is something missing for the final merge ? Reviews ?

wpichler:
What is missing on this PR to get approived and merged ?

pedrobaeza:
@rousseldenis is it everything OK for you?

contract/models/contract_line.py
```diff
@@ -0,0 +1,1086 @@
...
+    )
+    active = fields.Boolean(
+        string="Active",
+        related="contract_id.active",
+        store=True,
+        readonly=True,
+    )
+
+    @api.depends(
+        "last_date_invoiced", "date_start", "date_end", "contract_id.last_date_invoiced"
```
DucTruongKomit: Should we add 1 more depends contract_id.contract_line_ids?

contract/models/contract_line.py
```diff
@@ -0,0 +1,1086 @@
...
+        for rec in self:
+            if rec.date_end:
+                rec.termination_notice_date = rec.date_end - self.get_relative_delta(
+                    rec.termination_notice_rule_type,
+                    rec.termination_notice_interval,
+                )
+            else:
+                rec.termination_notice_date = False
+
+    @api.depends("is_canceled", "date_start", "date_end", "is_auto_renew")
```
DucTruongKomit: maybe it's missing manual_renew_needed and termination_notice_date.
also manual_renew_needed and successor_contract_line_id.

I imagine what if the field date_from, date_to are kept, only the fields, which I mention above, are changed the value?

contract/models/contract_modification.py
wpichler: Ok - will remove the default=False

contract/models/contract_line.py
wpichler: i will add a depends on  contract_id.contract_line_ids.last_date_invoiced

contract/models/contract_line.py
wpichler: i agree with you - added the depends...

wpichler:
@rousseldenis please review my last changes

manuelcalerosolis (approved):
Functional review ok!

pedrobaeza:
@rousseldenis is it OK for you?

rousseldenis (approved):

pedrobaeza:
Wait, aborting, as the commit history is a bit messed... I'll push a new branch with proper squashings.

pedrobaeza:
Squash done, so merging:

/ocabot merge nobump

PR #894, opened 2022-11-04
[15.0][FIX] contract: Switch string > title in buttons

pedrobaeza:
Forward-port of #893 

Since v14, the `string` attribute of the buttons is shown directly in UI. The tooltips are now put through `title` attribute. As that was the original intention, let's switch `string` to `title`.

@Tecnativa


victoralmau (approved):

rafaelbn (approved):
👍

PR #897, opened 2022-11-23
[14.0][MIG] contract_queue_job

BurkhalterY:
Standard migration


rousseldenis:
@BurkhalterY Thanks for this. Could you fix tests and pre-commit ?

rousseldenis:
@BurkhalterY Could you also follow migration guide https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-14.0#how-to

As the version change should appear in migration commit, not in pre-commit stuff. Thanks

BurkhalterY:
Hello, 

Thanks for the feedback

I notice that pre-commit is falling because Gitlab requires login.

https://gitlab.com/PyCQA/flake8
It works if we change to https://github.com/PyCQA/flake8

I'm working on the tests and I'll change version change to migration commit.

rousseldenis:
> Hello,
> 
> Thanks for the feedback
> 
> I notice that pre-commit is falling because Gitlab requires login.
> 
> https://gitlab.com/PyCQA/flake8 It works if we change to https://github.com/PyCQA/flake8
> 
> I'm working on the tests and I'll change version change to migration commit.

That's a known issue. I will update pre-commit config

BurkhalterY:
I moved the version change in the migration commit.

The test of contract module is falling. It's because the test checks only the directly returned account moves. But the inherit function of contract_queue_job module can't return it because there aren't yet created due to jobs use.

Do you have any idea how to fix it ?

Thank you !

rousseldenis:
> I moved the version change in the migration commit.
> 
> The test of contract module is falling. It's because the test checks only the directly returned account moves. But the inherit function of contract_queue_job module can't return it because there aren't yet created due to jobs use.
> 
> Do you have any idea how to fix it ?
> 
> Thank you !

There is an option to put in context to have no delayed jobs.

Check in queue_job module tests

contract_queue_job/__init__.py
```diff
@@ -0,0 +1,3 @@
+from . import models
+from . import wizards
+from . import tests
```
jguenat: You don't need to import tests

contract_queue_job/data/ir_config_parameter.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo noupdate="1">
+    <record id="config_param_contract_queue_job_as_job" model="ir.config_parameter">
+        <field name="key">contract.queue.job.as.job</field>
```
jguenat: I think "contract.queue.job.as.job" is a little bit confusing, "contract.queue.job" should be clear enough?

contract_queue_job/readme/USAGE.rst
```diff
@@ -0,0 +1,2 @@
+The feature can be enabled by setting the ir.config_parameter
+"acccontract.queue.job.as.job" to True.
```
jguenat: typo

jguenat (commented):
Hello @BurkhalterY ,
Thanks for the migration it looks good. I left some comments.

BurkhalterY:
Thank you @jguenat, I fixed it.

jguenat (approved):
LGTM, thanks

BurkhalterY:
@rousseldenis do you think that it's ok ?
Thank you

ygol:
hello,
not sure this should be reported here but testing this module I am getting an error caused by 
https://github.com/OCA/contract/blob/f449aab4cfdcc1274eb260c838a637ab286bff11/contract/models/contract_line.py#L622
when date is not valid but there's no problem when the invoice is generated manually.
```
File "/home/odoo/src/user/oca/contract/contract/models/contract_line.py", line 622, in _insert_markers
AttributeError: 'bool' object has no attribute 'strftime'
```
image

rousseldenis:
> hello, not sure this should be reported here but testing this module I am getting an error caused by
> 
> https://github.com/OCA/contract/blob/f449aab4cfdcc1274eb260c838a637ab286bff11/contract/models/contract_line.py#L622
> 
> 
> when date is not valid but there's no problem when the invoice is generated manually.
> ```
> File "/home/odoo/src/user/oca/contract/contract/models/contract_line.py", line 622, in _insert_markers
>     name = name.replace("#END#", last_date_invoiced.strftime(date_format))
> AttributeError: 'bool' object has no attribute 'strftime'
> ```
> 
> image

Some improvements should be merged, but not found yet an agreement on the method to do it.

See https://github.com/OCA/contract/pull/841

etobella (approved):
Tested on runbot LGTM

PR #899, opened 2022-12-01
[14.0][FIX] contract: Set fixed start dates in test contracts

victoralmau:
Set fixed start dates in test contracts.

### DIFF
--- a/contract/tests/test_contract.py
+++ b/contract/tests/test_contract.py
@@ -1723,7 +1723,7 @@ class TestContract(TestContractBase):
         self.assertFalse(line_4.successor_contract_line_id)
 
     def test_renew_create_new_line(self):
-        date_start = self.today - relativedelta(months=9)
+        date_start = fields.Date.from_string("2022-01-01")
         date_end = date_start + relativedelta(months=12) - relativedelta(days=1)
         self.acct_line.write(
             {
@@ -1743,7 +1743,7 @@ class TestContract(TestContractBase):
 
     def test_renew_extend_original_line(self):
         self.contract.company_id.create_new_line_at_contract_line_renew = False
-        date_start = self.today - relativedelta(months=9)
+        date_start = fields.Date.from_string("2022-01-01")
         date_end = date_start + relativedelta(months=12) - relativedelta(days=1)
         self.acct_line.write(
             {


chienandalu (approved):
:+1:

contract/tests/test_contract.py
```diff
@@ -669,14 +669,17 @@ def test_act_show_contract(self):
         show_contract = self.partner.with_context(
             contract_type="sale"
         ).act_show_contract()
-        self.assertDictContainsSubset(
```
pedrobaeza: This is done on purpose and you shouldn't change it.

contract/tests/test_contract.py
```diff
@@ -1739,7 +1742,11 @@ def test_renew_create_new_line(self):
         self.assertFalse(self.acct_line.is_auto_renew)
         self.assertTrue(new_line.is_auto_renew)
         self.assertEqual(new_line.date_start, date_start + relativedelta(months=12))
-        self.assertEqual(new_line.date_end, date_end + relativedelta(months=12))
+        # Be sure that the date is correct (leap years)
+        self.assertEqual(
+            new_line.date_end,
+            date_start + relativedelta(months=24) - relativedelta(days=1),
```
pedrobaeza: Why 2 years ?

contract/tests/test_contract.py
victoralmau: This change is to avoid warning log (deprecated). Change is the same as the one currently in 15.0 (https://github.com/OCA/contract/blob/15.0/contract/tests/test_contract.py#L637-L649)

contract/tests/test_contract.py
victoralmau: Before, 1 year was added because `date_end` was used, now, 2 years are added on `date_start` to be sure that the date is correct.

contract/tests/test_contract.py
pedrobaeza: But this is changing the meaning of the test. This can be done if you are sure that the dictionary is always going to return the same, but imaging a module that adds a new field in the act_window model, and if you do `read`, that value will be included in your return, not matching exactly the one expected, but being still correct.

contract/tests/test_contract.py
pedrobaeza: And why they aren't fixed dates (example, '2022-01-01')? It's one of the good test practices.

contract/tests/test_contract.py
victoralmau: Ok, this change is removed.

pedrobaeza: But isn't there any alternative not deprecated with the same result? And if in v15 it has been removed, it will happen eventually. This was put because an integration test failed.

victoralmau: What do you think about using `assertLessEqual`?

leemannd (commented):
Seems good. I would have used freezegun to set the dates

leemannd (approved):

pedrobaeza (commented):
The `assertEqual` sentences are not fixed date as well. We want fixed input + output data.

Issue #712, opened 2021-08-25
[14.0][BUG] - Contract with an End date raise an error


mkxmawilix:
Hello there,

I'm sorry let me know if am I on the wrong way but I'm not sure if this is a bad understanding of the module or if this is a real problem.

I have a problem if I want to specify the Date End of the Contract.

For instance this is Contract of 3 months invoiced 1 time for each month.

**Steps to reproduce the problem :** 

With an empty database and the Contracts (14.0) module freshly updated from github.
With and Odoo 14 with standard addons in a new project.

* Create a contract with a start_date in the past to make the test easier.
> date_start = 06/01/2021
> recurring interval = 1 
> recurring_rule_type = monthly
> recurring_invoice_type = pre-paid
> line_recurrence = False
* Add at least one line to invoice here
> name = Test
> quantity = 1
> price_unit = 10
* Save the contract
* Edit the contract and set the Date end and check the start/end dates are correctly set on Contract and on lines
> date_end = 08/01/2021
> We need to set the dates in two steps here because otherwise the Date end is not set when saving the first time.
* Before the first invoicing the dates are : 
> Date Start	06/01/2021
> Date End	08/01/2021
> Date of Next Invoice	06/01/2021
* Click on the button "Create Invoice" to simulate the call of the CRON
> 1st Click => OK : 
>> Date Start	06/01/2021
>> Date End	08/01/2021
>> Date of Next Invoice	07/01/2021
>
> 2nd Click => OK : 
>> Date Start	06/01/2021
>> Date End	08/01/2021
>> Date of Next Invoice	08/01/2021
>
> 3rd Click => OK / NOK ? : 
>> Date Start	06/01/2021
>> Date End	08/01/2021
>> Date of Next Invoice	06/01/2021
>
> Here the date of the next invoice is reset to the start date so at the next call of the CRON or click on the button this Contract will be invoiced because the "recurring_next_date" still set and lower than the "current date" of execution
> I guess the "recurring_next_date" should be empty here because this Contract dosn't need to be invoiced one more time because the Date end was 08/01/2021.
>
> 4th call the CRON => NOK
> An Odoo error is raise
>> ```
>> Error:
>> Odoo Server Error
>> 
>> Traceback (most recent call last):
>>   File "/opt/odoo/odoo/odoo/addons/base/models/ir_http.py", line 237, in _dispatch
>>     result = request.dispatch()
>>   File "/opt/odoo/odoo/odoo/http.py", line 683, in dispatch
[...]

PR #906, opened 2022-12-19
[IMP] contract: Not fail when missing some dates

hildickethan:
Avoid failing to create the invoices when missing some dates, since they are created in bulk in the cron

### DIFF
--- a/contract/models/contract_line.py
+++ b/contract/models/contract_line.py
@@ -639,16 +639,20 @@ class ContractLine(models.Model):
         lang = lang_obj.search([("code", "=", self.contract_id.partner_id.lang)])
         date_format = lang.date_format or "%m/%d/%Y"
         name = self.name
-        name = name.replace("#START#", first_date_invoiced.strftime(date_format))
-        name = name.replace("#END#", last_date_invoiced.strftime(date_format))
-        name = name.replace("#INVOICEMONTHNUMBER#", first_date_invoiced.strftime("%m"))
-        name = name.replace("#INVOICEYEAR#", first_date_invoiced.strftime("%Y"))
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Steps to reproduce the problem? Without them, this fix may solve anything that if not included in a regression test, will be broken again with any change. So this regression test is also very advisable.

hildickethan:
> Steps to reproduce the problem?

Ok I investigated the problem a bit more than this surface-level patch (though I still think it would be a good change to avoid blocking the cron for the rest of the invoices).
Some exact steps to reproduce the problem can be found in this issue https://github.com/OCA/contract/issues/712 which essentially faces the same problem. That has been discussed over time and fixed in multiple ways in both https://github.com/OCA/contract/pull/841 and https://github.com/OCA/contract/pull/854.

Those PRs intend to fix the root cause, which is some moments where it computes weirdly and ends up with incorrect date values in the contract, these incorrect dates are what cause the error since [these lines](https://github.com/OCA/contract/blob/14.0/contract/models/contract_recurrency_mixin.py#L185-L189) can potentially return False, while still being invoiced if it has a `recurring_next_date`.

This PR simply avoids the error by not erroring in the format, letting the invoice create itself even though in reality it shouldn't due to the dates being wrong, which sounds bad when typed out but in the context of the cron, which does all the invoices in bulk, IMO it is preferable to create the invoices and some extra over not creating any due to the format error in 1 of them.

> So this regression test is also very advisable.

Agree with this, but having seen the other PR(s), their tests should suffice as theyre covering the same issue, though I could do a small isolated one for this method specifically.

anajuaristi (approved):

PR #909, opened 2023-01-02
[14.0][FIX] contract_mode: fix payment_mode_id domain on supplier form

marcelsavegnago:
The payment_mode_id field is inserted in the contract_contract_form_view form and also in the contract_contract_supplier_form_view form. As the form contract_contract_supplier_form_view inherits the form contract_contract_form_view the payment_mode_id field appears twice in the form and with the default domain (inbound) instead of the expected domain (outbound).

image

Now

image

### DIFF
--- a/contract_payment_mode/views/contract_view.xml
+++ b/contract_payment_mode/views/contract_view.xml
@@ -48,11 +48,8 @@
         <field name="priority">18</field>
         <field name="inherit_id" ref="contract.contract_contract_supplier_form_view" />
         <field name="arch" type="xml">
-            <field name="partner_id" position="after">
-                <field
-                    name="payment_mode_id"
-                    domain="[('payment_type', '=', 'outbound')]"
-                />
+            <field name="payment_mode_id" position="attributes">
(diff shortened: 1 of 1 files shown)


rvalyi (approved):

augustodinizl:
@marcelsavegnago , the domain in the field creation: https://github.com/OCA/contract/blob/0ed31a345b5648b5085865a99d38d9c830711657/contract_payment_mode/models/contract.py#L10
have some impact on that outcome?

marcelsavegnago:
> @marcelsavegnago , the domain in the field creation:
> 
> https://github.com/OCA/contract/blob/0ed31a345b5648b5085865a99d38d9c830711657/contract_payment_mode/models/contract.py#L10
> 
> 
> have some impact on that outcome?

The suggestion made through this PR is to change the default domain when the form called is the contract_contract_supplier_form_view. In the case of the view of sales-type contracts, the domain is the one in the model.

augustodinizl (approved):

pedrobaeza:
Thanks for this! Please fw-port it to v15:

/ocabot merge minor

PR #917, opened 2023-01-19
[FIX]Add ref views on create_invoice action depending on invoices type

JulianaPoudou:
The functionnality Invoice sales contracts manually opens the supplier views instead of customer views. 
Add control on the type of invoice, then call ref views sale for customer invoices, and supplier views for the other fix the issue.

### DIFF
--- a/contract/__manifest__.py
+++ b/contract/__manifest__.py
@@ -17,7 +17,7 @@
               "LasLabs, "
               "Odoo Community Association (OCA)",
     'website': 'https://github.com/OCA/contract',
-    'depends': ['base', 'account', 'product', 'portal'],
+    'depends': ['base', 'account', 'auth_signup', 'product', 'portal'],
     "external_dependencies": {"python": ["dateutil"]},
     'data': [
         'security/groups.xml',
--- a/contract/wizards/contract_manually_create_invoice.py
+++ b/contract/wizards/contract_manually_create_invoice.py
@@ -56,10 +56,17 @@ class ContractManuallyCreateInvoice(models.TransientModel):
         invoices = self.env['account.invoice']
         for contract in self.contract_to_invoice_ids:
             invoices |= contract.recurring_create_invoice()
+        if self.contract_type == "sale":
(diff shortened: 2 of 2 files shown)


remi-filament (approved):
Makes sense
Code review LGTM
Thanks @JulianaPoudou

rousseldenis (approved):
LGTM

rousseldenis:
@JulianaPoudou Could you rebase ?

JulianaPoudou:
Done @rousseldenis

rousseldenis:
@JulianaPoudou Could you fix pre-commit ?

nimarosa (approved):

rousseldenis:
@JulianaPoudou

remi-filament:
Hi @rousseldenis trying to take over these remaining actions.
I just rebased and fixed pre-commit, however there is an error on Odoo test with portal test which do not occurs when testing the same thing with OCB code, and I cannot reproduce locally so I am not sure if this is a problem with GitHub Action or if there is really something to fix in the code ?

rousseldenis:
Mmmmh, I see this : https://github.com/OCA/OCB/pull/872/files

So, this can explain the problem.

@pedrobaeza @chienandalu Any insight here ?

pedrobaeza:
If the module or the tests depend on `auth_signup`, you should explicitly depend on it if the sources are not trustable like in this case. We added such dependency for avoiding update errors on core modules.

rousseldenis:
> If the module or the tests depend on `auth_signup`, you should explicitly depend on it if the sources are not trustable like in this case. We added such dependency for avoiding update errors on core modules.

In fact, module depends on portal and tests are running portal flows, so @JulianaPoudou I recommend to depends on auth_signup too (that should not be mandatory on further versions).

remi-filament:
I added auth_signup in dependencies but I still get the same error on test with Odoo. Not sure what else should be done ?

remi-filament:
Hi @rousseldenis I just rebased, squashed commits and pushed again and now Odoo tests are green !
Can this one be merged ?
Thanks !

PR #928, opened 2023-02-06
[FIX] Do preserve the uom from the contract line 

wpichler:
contract.line is configured with a qty and a product uom. Will get set in the _prepare_sale_line_vals function. Then there is a call to the onchange function - there product_uom will get set from the product. So we have to set it here again.

### DIFF
--- a/contract_sale_generation/models/contract_line.py
+++ b/contract_sale_generation/models/contract_line.py
@@ -40,7 +40,10 @@ class ContractLine(models.Model):
             )
             order_line.order_id = sale
         # Get other order line values from product onchange
+        uom = order_line['product_uom']
         order_line.product_id_change()
+        # Do set again the product_uom - the product_id_change function does reset it before
+        order_line['product_uom'] = uom
         sale_line_vals = order_line._convert_to_write(order_line._cache)
         # Insert markers
         name = self._insert_markers(dates[0], dates[1])


contract_sale_generation/models/contract_line.py
```diff
@@ -40,7 +40,10 @@ def _prepare_sale_line(self, order_id=False, sale_values=False):
             )
             order_line.order_id = sale
         # Get other order line values from product onchange
+        uom = order_line['product_uom']
```
Saran440, line 43: Is it possible to send something in context and skip change uom?

```suggestion
        order_line.with_context(skip_change_uom=1).product_id_change()
        
        def product_id_change(self):
            ...
            if self._context.get('skip_change_uom'):
                ...do something...
```

wpichler: No - this would require to change a core function

contract_sale_generation/models/contract_line.py
```diff
@@ -40,7 +40,10 @@ def _prepare_sale_line(self, order_id=False, sale_values=False):
...
         order_line.product_id_change()
+        # Do set again the product_uom - the product_id_change function does reset it before
+        order_line['product_uom'] = uom
```
rousseldenis, line 46: @wpichler  I think you can put it in sale_line_vals.update() instead. What do you think ?

Issue #803, opened 2022-03-10
[12.0][Bug] The "recurring_next_date" of contract line is not set after renewing this line.


marylla:
## Module
* contract

## Describe the bug
The "recurring_next_date" of contract line is not set after renewing this line.

## To Reproduce
**Affected versions**: 12.0
(I just tested in 12.0, so I don't know if this also affects other versions.)

1. Create a contract for a customer / supplier.
2. Add a product and choose monthly invoicing by setting recurring_interval = 1 and recurring_rule_type = "monthly".
3. Set the start date to 2021-01-01 (or something else in the past for testing purpose).
4. Set is_auto_renew = True, and then auto_renew_interval = 3 and auto_renew_rule_type = "monthly".
5. Create the 3 invoices by using the cron job or by clicking on "create invoice" button (with developer mode).
6. Now, I need to renew my contract line because I created the invoice for january, february and march. So, I click on "renew" on the contract line or use the cronjob.
7. The new end date is correctly set to 2021-06-30, the next_period_date_start is correctly set to 2021-04-01 and the next_period_date_end is correctly set 2021-04-30. So far, everything is fine.
8. BUT: The recurring_next_date is not set (should be 2021-04-01), and therefore the field create_invoice_visibility is still False. 
And that is why my contract is not displayed as "in progress" and I cannot see the "create invoice" button anymore, even though I am in the developer mode. The cronjob to automatically create the invoice also does nothing.

**Expected behavior**
After clicking on "renew" on a contract line or using the cronjob the recurring_next_date should be set to next invoicing date.

eLBati:
Fixing this with https://github.com/OCA/contract/pull/930

@marylla could have a look?

PR #930, opened 2023-02-23
[12.0] FIX contract - setting recurring_next_date when renewing extending line

eLBati:
FIX https://github.com/OCA/contract/issues/803


mrcast (approved):
Functional test:OK!

PR #934, opened 2023-03-13
contract_variable_qty_prorated tr.po created.

ko-eska:


pedrobaeza:
Please use weblate for translations. You can reuse this work importing the file.

https://odoo-community.org/resources/translate

PR #935, opened 2023-03-13
Contract invoice start end dates

ko-eska:


pedrobaeza:
Please use weblate for translations. You can reuse this work importing the file.

https://odoo-community.org/resources/translate

PR #937, opened 2023-03-21
[14.0] [ADD] contract_price_increase

paradoxxxzero:
This module allows to define contract price increase at a given date for a given rate.

The contract line prices increase is done by a cron.


paradoxxxzero:
@hparfr

rousseldenis:
@paradoxxxzero There is already this one : https://github.com/OCA/contract/tree/13.0/contract_price_revision

Could you check that ?

PR #936, opened 2023-03-21
[14.0] [ADD] contract_price_history

paradoxxxzero:
This module allows historization of all the price changes in contract lines.
It also allows to revert to a previous price.

PR #938, opened 2023-03-21
[14.0] [ADD] contract_price_increase_history

paradoxxxzero:
This module adds "increase" as price history type for contract price increases.

It depends on #937 and #936


paradoxxxzero:
@hparfr

PR #940, opened 2023-03-21
[14.0][FIX] contract: subtotal miscalculated when price_unit displayed with fewer digits than actual value

eantones:
This commit addresses an issue in the "contract" module where the subtotal was incorrectly calculated when the `price_unit` field was displayed with fewer digits than its actual value, resulting in discrepancies of a few cents.

The change made in this commit is: Adding the `digits` attribute with the value "Product Price" to the `price_unit` field, ensuring the proper display of decimal places, leading to accurate subtotal calculations and eliminating cent discrepancies.

Before:
image

After:
image


nimarosa (approved):

rousseldenis (approved):
Code review

PR #943, opened 2023-04-05
[16.0] [New] Module : contract sale generation client order ref

manuelcalerosolis:
This module extends functionality of sale contracts to select client order ref.


rousseldenis:
@manuelcalerosolis It seems that your commits are weird...

PR #946, opened 2023-05-08
[14.0] [IMP] Add INVOICEMONTHNAME marker to line description

odooNextev:

### DIFF
--- a/contract/models/contract_line.py
+++ b/contract/models/contract_line.py
@@ -616,6 +616,23 @@ class ContractLine(models.Model):
         )
         return first_date_invoiced, last_date_invoiced, recurring_next_date
 
+    def _translate_marker_month_name(self, month_name):
+        months = {
+            "January": _("January"),
+            "February": _("February"),
+            "March": _("March"),
+            "April": _("April"),
+            "May": _("May"),
+            "June": _("June"),
+            "July": _("July"),
+            "August": _("August"),
+            "September": _("September"),
+            "October": _("October"),
+            "November": _("November"),
+            "December": _("December"),
+        }
+        return months[month_name]
+
     def _insert_markers(self, first_date_invoiced, last_date_invoiced):
         self.ensure_one()
         lang_obj = self.env["res.lang"]
@@ -624,6 +641,12 @@ class ContractLine(models.Model):
         name = self.name
         name = name.replace("#START#", first_date_invoiced.strftime(date_format))
         name = name.replace("#END#", last_date_invoiced.strftime(date_format))
+        name = name.replace(
+            "#INVOICEMONTHNAME#",
+            self.with_context(lang=lang.code)._translate_marker_month_name(
+                first_date_invoiced.strftime("%B")
+            ),
+        )
         return name
 
     def _update_recurring_next_date(self):
--- a/contract/tests/test_contract.py
+++ b/contract/tests/test_contract.py
@@ -150,7 +150,7 @@ class TestContractBase(common.SavepointCase):
                         0,
                         {
                             "product_id": False,
-                            "name": "Header for Services",
+                            "name": "Header for #INVOICEMONTHNAME# Services",
                             "display_type": "line_section",
                         },
                     ),
@@ -2453,3 +2453,7 @@ class TestContract(TestContractBase):
         self.acct_line.uom_id = uom_day.id
         self.acct_line.refresh()
         self.assertEqual(self.acct_line.price_unit, 30.75 * 8)
+
+    def test_check_month_name_marker(self):
+        invoice_id = self.contract3.recurring_create_invoice()
+        self.assertEqual(invoice_id.invoice_line_ids[0].name, "Header for May Services")
--- a/contract/readme/USAGE.rst
+++ b/contract/readme/USAGE.rst
@@ -10,8 +10,8 @@
    * the recurrence parameters: interval (days, weeks, months, months last day or years),
      start date, date of next invoice (automatically computed, can be modified) and end date (optional)
    * auto-price, for having a price automatically obtained from the price list
-   * #START# or #END# in the description field to display the start/end date of
-     the invoiced period in the invoice line description
+   * #START# - #END# or #INVOICEMONTHNAME# in the description field to display
+     the start/end date or the start month of the invoiced period in the invoice line description
    * pre-paid (invoice at period start) or post-paid (invoice at start of next period)
 
 #. The "Generate Recurring Invoices from Contracts" cron runs daily to generate the invoices.
--- a/contract/views/contract.xml
+++ b/contract/views/contract.xml
@@ -466,6 +466,13 @@
                                     the
                                     invoiced period
                                 </p>
+                                <p colspan="2">
+                                    <strong>#INVOICEMONTHNAME#</strong>
+                                    : Invoice month name
+                                    of
+                                    the
+                                    invoiced period
+                                </p>
                             </group>
                         </page>
                     </notebook>
--- a/contract/views/contract_template.xml
+++ b/contract/views/contract_template.xml
@@ -68,6 +68,8 @@
                 >
                     <p> <strong>#START#</strong>: Start date of the invoiced period</p>
                     <p> <strong>#END#</strong>: End date of the invoiced period</p>
+                    <p> <strong
+                        >#INVOICEMONTHNAME#</strong>: Invoice month name of the invoiced period</p>
                 </group>
             </form>
         </field>


contract/tests/test_contract.py
```diff
@@ -150,7 +150,7 @@ def setUpClass(cls):
                         0,
                         {
                             "product_id": False,
-                            "name": "Header for Services",
+                            "name": "Header for #INVOICEMONTHNAME# Services",
```
rousseldenis, line 153: Ok, so you should at least test the generated invoice line name.

odooNextev: Hi @rousseldenis,
thank you for your review.
Sorry, but I'm a newbie with tests, I guess you told to use the marker in invoice line instead of section.
So, could be this enough to resolve?

```
cls.contract3 = cls.env["contract.contract"].create(
            {
                "name": "Test Contract 3",
                "partner_id": cls.partner.id,
                "pricelist_id": cls.partner.property_product_pricelist.id,
                "line_recurrence": False,
                "contract_type": "sale",
                "recurring_interval": 1,
                "recurring_rule_type": "monthly",
                "date_start": "2018-02-15",
                "contract_line_ids": [
                    (
                        0,
                        0,
                        {
                            "product_id": False,
                            "name": "Header for Services",
                            "display_type": "line_section",
                        },
                    ),
                    (
                        0,
                        0,
                        {
                            "product_id": False,
                            "name": "Services from #START# to #END#",
                            "quantity": 1,
                            "price_unit": 100,
                        },
                    ),
                    (
                        0,
                        0,
                        {
                            "product_id": False,
                            "name": "Line",
                            "quantity": 1,
                            "price_unit": 120,
                        },
                    ),
                    (
                        0,
                        0,
                        {
                            "product_id": False,
                            "name": "Header for #INVOICEMONTHNAME# Services",
                            "display_type": "line_section",
                        },
                    ),
                    (
                        0,
                        0,
                        {
                            "product_id": cls.product_1.id,
                            "name": "Services for #INVOICEMONTHNAME#",
                            "quantity": 1,
                            "uom_id": cls.product_1.uom_id.id,
                            "price_unit": 100,
                            "discount": 50,
                        },
                    ),
                ],
            }
```

odooNextev: Hi @rousseldenis,
thank you for your review.
Sorry, but I'm a newbie with tests, could be this enough to resolve?

```
def test_check_month_name_marker(self):
    invoice_id = self.contract3.recurring_create_invoice()
    self.assertEqual(invoice_id.invoice_line_ids[0].name, "Header for May Services")
```

I'm not sure of a couple of things, first `self.contract3.recurring_create_invoice()` is called 3 times before my call: every method will increase the month so I would check "August" instead "May"?
Furthermore, I suppose test env is set to English, so can I directly set English term for the month name?

rousseldenis: Yes a test like should be sufficent.

But translation speaking, I'm not sure about the behaviour of the '_' function with that code:

image


As the translated term is not known at advance.

Either strftime provides translations (I doubt of that), either you should write a constant with a dict of every month strftime value with its translation value (which I think is the best).

odooNextev: Hi Denis,
I found that '\_' function translates strings using the language defined in `self.env.context`.
I don't know how to pass language explicitly, if it was a ORM function we could use `.with_context(lang=...)`.
Language variable is defined few row above: https://github.com/OCA/contract/blob/d3e488c8cecedee8596ed5b6baa8909703f662da/contract/models/contract_line.py#L622
I did many test in debug and changing forcibly `self.env.context`, it works, but it's not a good way.
Do you know how to force '_' function destination language?

contract/models/contract_line.py
```diff
@@ -616,6 +616,9 @@ def _get_period_to_invoice(
         )
         return first_date_invoiced, last_date_invoiced, recurring_next_date
 
+    def _translate_marker_month_name(self, month_name):
```
odooNextev, line 619: @rousseldenis I added this method
I don't like it too much, but it works

contract/i18n/contract.pot
```diff
@@ -2244,3 +2244,75 @@ msgstr ""
 #: model_terms:ir.ui.view,arch_db:contract.contract_contract_form_view
 msgid "on"
 msgstr ""
+
+#. module: contract
```
odooNextev: I had to add these lines in .pot cause without `account_reports ` module it doesn't load months translation cause only in this module are "code" type.
Without it, in `account` base module there're only `model` translation type

rousseldenis: That's why you should not do like that. pot files should be generated automatically.

As said before, add this to your code:

```python
months = {
    "January": _("January"),
    "February", _("February"),
    (...)
}

(...)

def _translate_marker_month_name(self, month_name):
   return months[month_name]

```

This will allow to generate pot file with all months values and allow to follow translation normal flow.

Could you check that ?

odooNextev: Thank you so much Denis!
You told me before, but I didn't quite understand what you meant.
How can I check if .pot is correctly generated?
Is a CI/CD operation?

rousseldenis: > Thank you so much Denis! You told me before, but I didn't quite understand what you meant. How can I check if .pot is correctly generated? Is a CI/CD operation?

You can do it manually:

Settings > Translations > Export > 'New language / PO file'

The generated one should be correct.

Here, it will be generated automatically after merge

odooNextev: Thank you again
I found them in .pot

```
#. module: contract
#: code:addons/contract/models/contract_line.py:0
#, python-format
msgid "January"
msgstr ""
```

odooNextev: image

matteotognini (commented):
I tried and it works.

francesco-ooops (approved):
Functional review ok!

@rousseldenis what do you say?

matteotognini (approved):
Functional review ok!

contract/models/contract_line.py
```diff
@@ -624,6 +641,12 @@ def _insert_markers(self, first_date_invoiced, last_date_invoiced):
         name = self.name
         name = name.replace("#START#", first_date_invoiced.strftime(date_format))
         name = name.replace("#END#", last_date_invoiced.strftime(date_format))
+        name = name.replace(
```
pedrobaeza, line 644: Why not using Python for getting the localized month instead of adding more translations?

https://stackoverflow.com/a/17902535

babatoko:
Would it be possible to use that marker into a section line?

contract/models/contract_line.py
odooNextev: Hi Pedro,
it works fine!
I'm always afraid to change locale, even if for an instant, but if it's tested and doesn't affect other behaviors, it's fine, thanks

odooNextev:
> Would it be possible to use that marker into a section line?

Yes, it's possible.

image

contract/models/contract_line.py
```diff
@@ -641,12 +625,12 @@ def _insert_markers(self, first_date_invoiced, last_date_invoiced):
         name = self.name
         name = name.replace("#START#", first_date_invoiced.strftime(date_format))
         name = name.replace("#END#", last_date_invoiced.strftime(date_format))
+        locale.setlocale(locale.LC_TIME, lang.code)
```
rousseldenis: @pedrobaeza Doesn't those lines deserve to be emphasized with try/except/finally ?

pedrobaeza: Not sure what can fail there... Do you know it?

odooNextev: I tried to change "LC_TIME" to "LC_ALL" to see if it was the problem in Odoo test fails, but it doesn't change...
I found this, but it seems an env issue: https://stackoverflow.com/questions/14547631/python-locale-error-unsupported-locale-setting

rousseldenis: Test fail because locale is not present in system. That's why I'd prefer the former solution

odooNextev: Should I come back to your solution @rousseldenis?

pedrobaeza: A fallback locale should exist, isn't it?

odooNextev: > A fallback locale should exist, isn't it?

Sorry Pedro, but I don't know any fallback solution if not `try except`.
I guess it will be only useful to pass tests, but if someone installs this module in a system with no `locale` available, translations won't work.
@pedrobaeza @rousseldenis Let me know what I can do and I'll do it

pedrobaeza: OK, at this point I can't enforce to use this technique, as I don't know if there's a minimum locale per system.

odooNextev:
@rousseldenis I reverted to the commit you approved

PR #951, opened 2023-05-17
Fix #INVOICEMONTHNAME# for all languages

sistecem:
It was working only in English
- Changed %B to %m to work on all languages
- Update month translations in es.po
- Add the possibility to replace  #INVOICEMONTHNUMER# and #INVOICEYEAR# in line descriptions


AsierAcedo:
LGTM, Fix pre-commit

PR #953, opened 2023-05-19
[14.0][FIX] contract: show uom only when multi uom is activated

niyasraphy:
before this commit, the unit of measure field is
always visible in different views, irrespective
of the unit of measure option in the settings.

after this commit, the unit of measure field will
be shown only if the multi uom feature is activated in the settings.


pedrobaeza (approved):

rvalyi (approved):

pedrobaeza:
You can fw-port it to upper versions:

/ocabot merge minor

PR #954, opened 2023-05-22
Fix #INVOICEMONTHNAME# for all languages

sistecem:
It was working only in English

- Changed %B to %m to work on all languages
- Update month translations in es.po


AsierAcedo:
LGTM

rousseldenis (approved):
That's indeed better

pedrobaeza (approved):
/ocabot merge patch

PR #959, opened 2023-06-08
[16.0] [MIG] product_contract

Rad0van:
Fairly standard, had only to update one compute method name.


product_contract/models/contract_line.py
```diff
@@ -0,0 +1,53 @@
...
+
+    sale_order_line_id = fields.Many2one(
+        comodel_name="sale.order.line",
+        string="Sale Order Line",
+        required=False,
+        copy=False,
+    )
+
+    def _prepare_invoice_line(self):
+        res = super(ContractLine, self)._prepare_invoice_line()
```
rousseldenis:
```suggestion
        res = super()._prepare_invoice_line()
```

product_contract/models/contract_line.py
```diff
@@ -0,0 +1,53 @@
...
+        if self.sale_order_line_id and res:
+            res["sale_line_ids"] = [(6, 0, [self.sale_order_line_id.id])]
```
rousseldenis: FYI, there is now a Command in odoo.fields in order to avoid magic numbers and improve readability

product_contract/models/contract_line.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def _get_auto_renew_rule_type(self):
+        """monthly last day don't make sense for auto_renew_rule_type"""
+        self.ensure_one()
+        if self.recurring_rule_type == "monthlylastday":
+            return "monthly"
+        return self.recurring_rule_type
+
+    @api.onchange("product_id")
+    def _onchange_product_id_recurring_info(self):
```
rousseldenis, line 34: It's maybe better to avoid onchanges as fields are now computed and override compute methods

product_contract/models/product_template.py
```diff
@@ -0,0 +1,79 @@
...
+                    {"property_contract_template_id": False}
+                )
+        return super().write(vals)
+
+    @api.constrains("is_contract", "type")
+    def _check_contract_product_type(self):
+        """
+        Contract product should be service type
+        """
+        if self.is_contract and self.type != "service":
```
rousseldenis: Careful, self can be a multi recordset.

product_contract/models/sale_order.py
```diff
@@ -0,0 +1,145 @@
...
+            for line in line_to_update_contract:
+                line.create_contract_line(line.contract_id)
+        return contract_model.browse(contracts)
+
+    def action_confirm(self):
+        """If we have a contract in the order, set it up"""
+        self.filtered(
+            lambda order: (order.company_id.create_contract_at_sale_order_confirmation)
+        ).action_create_contract()
+        return super(SaleOrder, self).action_confirm()
```
rousseldenis:
```suggestion
        return super().action_confirm()
```

product_contract/models/sale_order_line.py
```diff
@@ -0,0 +1,279 @@
...
+        for rec in self:
+            if (
+                rec.order_id.state not in ("sale", "done", "cancel")
+                and rec.contract_id.is_terminated
+            ):
+                raise ValidationError(
+                    _("You can't upsell or downsell a terminated contract")
+                )
+
+    @api.depends("product_id")
```
rousseldenis: IMHO, 'order_id.company_id' should be added here

rousseldenis (changes requested):
Code review

product_contract/models/contract_line.py
Rad0van: 3 of these fields have compute methods. If I wanted to override these, is it possible to extend their depends list? Like going from this:
`@api.depends("contract_id.recurring_rule_type", "contract_id.line_recurrence")`
to this:
`@api.depends("contract_id.recurring_rule_type", "contract_id.line_recurrence", "product_id")`

What about the rest of the fields? Should I leave them in the onchange method?

All suggestions are welcome @rousseldenis

rousseldenis: > 3 of these fields have compute methods. If I wanted to override these, is it possible to extend their depends list? Like going from this: `@api.depends("contract_id.recurring_rule_type", "contract_id.line_recurrence")` to this: `@api.depends("contract_id.recurring_rule_type", "contract_id.line_recurrence", "product_id")`


I asked Odoo for this question before and overrides could just add "product_id" as they keep previous ones.

> What about the rest of the fields? Should I leave them in the onchange method?

Yes, unfortunately :-)

Rad0van:
@rousseldenis there is a typo in sale_order.py:

```
    @api.constrains("state")
    def _check_contact_is_not_terminated(self):
```

It should have been _check_contRact_is_not_terminated Should that be fixed? If so, should it be fixed in migration or in separate PR?

product_contract/models/sale_order_line.py
```diff
@@ -0,0 +1,279 @@
...
+            self.date_start
+            + contract_line_model.get_relative_delta(
+                self._get_auto_renew_rule_type(),
+                int(self.product_uom_qty),
+            )
+            - relativedelta(days=1)
+        )
+        return date_end
+
+    @api.depends("product_id")
```
Rad0van, line 123: @rousseldenis, @sbejaoui, @pedrobaeza, @sbidoul  maybe you could help me here please. I am trying to use this module in e-shop. However when I purchase the product, Contract is created but all the relevant fields are empty. What I found out is that this method `_compute_auto_renew` is not called in such scenario. I'm scratching my head on why is it never called and how to solve this in some good and elegant way. Any help / advice is really appreciated.

espo-tony: Hi @Rad0van ,
I think the issue is that all the fields computed by this method have a default value.
What happens at record creation is that get assigned their default value so they don't need to be computed;
The computation method triggers-in, correctly, next time the observed field (`product_id`) changes.

This issue doesn't happen when a user manually creates a Sale Order because the computation method is immediately triggered when `product_id` changes (as the computation method is also part of the `onchange` workflow).

Besides causing this issue, a default value on a computed field doesn't make much sense in my opinion. I suggest you remove all defaults from field definition and maybe add them in the computation method itself.

Rad0van: > 

Hi @espo-tony ,

not really sure I follow what you suggest here because the issue is as you point out that when Sale Order is created directly, the computation method is not called. I have created a hacky solution by creating glue module between product_contract and website_sale where I basically reproduce the whole `_compute_auto_renew` method except for setting the quantity. It goes like this:

```
class SaleOrder(models.Model):
    _inherit = "sale.order"

    def _cart_update_order_line(self, product_id, quantity, order_line, **kwargs):
        """
        This ensures that when contract product is put into basket in e-commerce
        all necessary values are set
        """
        order_line = super()._cart_update_order_line(
            product_id, quantity, order_line, **kwargs
        )

        """
        This is (apart from commented out line) 1:1 copy of _compute_auto_renew method
        from product_contract module to ensure we get the defaults configured
        """
        if order_line.product_id.is_contract:
            # order_line.product_uom_qty = order_line.product_id.default_qty
            order_line.recurring_rule_type = order_line.product_id.recurring_rule_type
            order_line.recurring_invoicing_type = (
                order_line.product_id.recurring_invoicing_type
            )
            order_line.date_start = order_line.date_start or fields.Date.today()

            order_line.date_end = order_line._get_date_end()
            order_line.is_auto_renew = order_line.product_id.is_auto_renew
            if order_line.is_auto_renew:
                order_line.auto_renew_interval = (
                    order_line.product_id.auto_renew_interval
                )
                order_line.auto_renew_rule_type = (
                    order_line.product_id.auto_renew_rule_type
                )

        return order_line

```

espo-tony: Hi @Rad0van 
When a Sale Order is created directly, the computation method is not called **because of** the default values. If you remove them, the computation method will be called even when a Sale Order is created directly. I tested this solution locally and it worked for me.

Rad0van: > Hi @Rad0van When a Sale Order is created directly, the computation method is not called **because of** the default values. If you remove them, the computation method will be called even when a Sale Order is created directly. I tested this solution locally and it worked for me.

Oh I see now! But that is quite a big change which I think should be approved by more senior people.
Also I would be interested to get to know the inner mechanics of this. I.e. why the *@api.depends* is called when the default values are not present? It only depends on *product_id* Does anyone know the explanation?

product_contract/views/sale_order.xml
```diff
@@ -0,0 +1,124 @@
...
+    License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+-->
+<odoo>
+    <record id="view_order_form" model="ir.ui.view">
+        <field name="name">sale.order.form (in product_contract)</field>
+        <field name="model">sale.order</field>
+        <field name="inherit_id" ref="sale.view_order_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//header" position="inside">
+                <field name="need_contract_creation" attrs="{'invisible': []}" />
```
Abranes:
```suggestion
                <field name="need_contract_creation" invisible="1" />
```

product_contract/views/sale_order.xml
```diff
@@ -0,0 +1,124 @@
...
+                    type="object"
+                    icon="fa-book"
+                    class="oe_stat_button"
+                    attrs="{'invisible': ['|', '|', ('is_contract', '!=', True), ('state', 'not in', ['sale', 'done']), ('contract_count', '=', 0)]}"
+                >
+                    <field string="Contracts" name="contract_count" widget="statinfo" />
+                </button>
+            </xpath>
+            <xpath expr="//field[@name='order_line']" position="before">
+                <field name="is_contract" attrs="{'invisible': []}" />
```
Abranes:
```suggestion
                <field name="is_contract" invisible="1" />
```

product_contract/views/sale_order.xml
```diff
@@ -0,0 +1,124 @@
...
+            </xpath>
+            <xpath
+                expr="//field[@name='order_line']/form//field[@name='product_id']"
+                position="after"
+            >
+                <field name="contract_template_id" attrs="{'invisible': []}" />
```
Abranes:
```suggestion
                <field name="contract_template_id" invisible="1" />
```

product_contract/views/sale_order.xml
```diff
@@ -0,0 +1,124 @@
...
+                    name="contract_line_id"
+                    attrs="{'invisible': [('is_contract', '=', False)]}"
+                    domain="[('contract_id','=',contract_id)]"
+                />
+            </xpath>
+            <xpath
+                expr="//field[@name='order_line']/form//field[@name='tax_id']/parent::group"
+                position="after"
+            >
+                <field name="is_contract" attrs="{'invisible': []}" />
```
Abranes:
```suggestion
                <field name="is_contract" invisible="1" />
```

Abranes (changes requested):
Change this `attrs` to `invisible` to fix the view problems

image
Image from runboat

JorgeQuinteros (approved):
LGTM Functional Review

gregory-moka:
Hello,

tried to add a contract line on a Sale Order. As I want to confirm it, this error message is displayed : 
image
=> Maybe the contract type should be required as we configure the product

Abranes (approved):
Code review

LGTM!

DorianMAG:
Thx for this work !
Tested on runboat, LGTM

navarrorico (approved):
LGTM :+1:

product_contract/models/sale_order_line.py
```diff
@@ -0,0 +1,279 @@
...
+            "recurring_rule_type": self.recurring_rule_type,
+            "is_auto_renew": self.is_auto_renew,
+            "auto_renew_interval": self.auto_renew_interval,
+            "auto_renew_rule_type": self.auto_renew_rule_type,
+            "termination_notice_interval": termination_notice_interval,
+            "termination_notice_rule_type": termination_notice_rule_type,
+            "contract_id": contract.id,
+            "sale_order_line_id": self.id,
+            "predecessor_contract_line_id": predecessor_contract_line_id,
+            "analytic_account_id": self.order_id.analytic_account_id.id,
```
Abranes: There is no more `analytic_account_id`, now you need to use `analytic_distribution`.

remi-filament: Hi @Rad0van thank you for porting this module, I just tested it locally (cannot find that PR on Runboat...) and except for this missing analytic distribution, it works fine !

analytic_account_id still exists on sale order, although analytic_distribution from sale order line needs to be forwarded to contract line and then to invoice line.
You should add the following :
```suggestion
            "analytic_account_id": self.order_id.analytic_account_id.id,
            "analytic_distribution": self.analytic_distribution,
```

Rad0van: @remi-filament done.

Rad0van:
@remi-filament seems @Abranes is right. In the commit #1019 the _analytic_account_id_ was indeed removed...

remi-filament:
> @remi-filament seems @Abranes is right. In the commit #1019 the _analytic_account_id_ was indeed removed...

Ah yes, my bad I checked contract code on your branch and not on upstream when your branch is outdated for that topic...
It makes sense that it is removed, there is still a group_id on contract but it does not seem to be used anywhere so better to not use analytic account anymore.

You should then rebase this PR on OCA v16 branch and squash your 3 last commits in one before this PR can be merged.

product_contract/models/sale_order.py
```diff
@@ -0,0 +1,150 @@
...
+                order_lines = line_to_create_contract.filtered(
+                    lambda r, template=contract_template: r.product_id.with_company(
+                        r.order_id.company_id
+                    ).property_contract_template_id
+                    == template
+                )
+                contract = contract_model.create(
+                    rec._prepare_contract_value(contract_template)
+                )
+                contracts.append(contract)
```
jaenbox: When performing a create(), the created object 'contract.contract(xxx,)' is returned. When appending to the list named 'contracts', you should indicate performing a contract.id to create a list of contract identifiers. If this is not done, when performing a contract_model.browse(contracts), it returns contract.contract(contract.contract(xxx,),), which is incorrect. From what I see, it should return contract.contract(xxx,).

aliciagaarzo (approved):
LGTM!

aliciagaarzo (commented):
Review to cover all tests. Regards.

product_contract/models/sale_order.py
pedrobaeza: @Rad0van can you answer this comment?

product_contract/models/sale_order.py
```diff
@@ -0,0 +1,150 @@
...
+                template_name=contract_template.name, sale_name=self.name
+            ),
+            "partner_id": self.partner_id.id,
+            "company_id": self.company_id.id,
+            "contract_template_id": contract_template.id,
+            "user_id": self.user_id.id,
+            "payment_term_id": self.payment_term_id.id,
+            "fiscal_position_id": self.fiscal_position_id.id,
+            "invoice_partner_id": self.partner_invoice_id.id,
+            "line_recurrence": self.partner_invoice_id.id,
```
flalexg: Why the line_recurrence is the self.partner_invoice_id?

product_contract/models/sale_order.py
```diff
@@ -0,0 +1,150 @@
...
+        contract_model = self.env["contract.contract"]
+        contracts = []
+        for rec in self.filtered("is_contract"):
+            line_to_create_contract = rec.order_line.filtered(
+                lambda r: not r.contract_id and r.product_id.is_contract
+            )
+            line_to_update_contract = rec.order_line.filtered(
+                lambda r: r.contract_id
+                and r.product_id.is_contract
+                and r
```
flalexg, line 73: You should check that r exists at the beggining of the function, no? Otherwise, when performing r.contract_id it will fail.

product_contract/models/sale_order.py
```diff
@@ -0,0 +1,150 @@
...
+            "company_id": self.company_id.id,
+            "contract_template_id": contract_template.id,
+            "user_id": self.user_id.id,
+            "payment_term_id": self.payment_term_id.id,
+            "fiscal_position_id": self.fiscal_position_id.id,
+            "invoice_partner_id": self.partner_invoice_id.id,
+            "line_recurrence": self.partner_invoice_id.id,
+        }
+
+    def action_create_contract(self):
```
flalexg, line 63: If I change the product_uom_qty before creating the contract, when creating the contract the quantity is doesn't match.

image
image

flalexg (changes requested):

product_contract/models/sale_order.py
Rad0van: > Why the line_recurrence is the self.partner_invoice_id?

Funny, it's the same as in original module: https://github.com/OCA/contract/blob/aca1183f804685361a0c119f545d547506d2aa41/product_contract/models/sale_order.py#L76-L79

But probably needs fixing as well.

mrcast (approved):
Functional test ok!

jaenbox:
Hello @Rad0van, I just wanted to ask if you were planning to continue with the PR. I see that you have some unanswered comments, and I was wondering if I can proceed with the PR of the product_contract module. If it's not a problem!

Rad0van:
> Hello @Rad0van, I just wanted to ask if you were planning to continue with the PR. I see that you have some unanswered comments, and I was wondering if I can proceed with the PR of the product_contract module. If it's not a problem!

Was quite busy but planning to finish this one soon.

pedrobaeza:
@rousseldenis can you update your review?

product_contract/models/sale_order.py
Rad0van: > You should check that r exists at the beggining of the function, no? Otherwise, when performing r.contract_id it will fail.

It's just formatting that's misleading you. Check the following line ;-)

Rad0van:
I have carried over changes done in #1083 so now these are 1:1 but also fix a typo in constrain method name.

product_contract/views/product_template.xml
```diff
@@ -0,0 +1,80 @@
...
+        <field name="name">account.invoice.select.contract</field>
+        <field name="model">product.template</field>
+        <field name="inherit_id" ref="product.product_template_form_view" />
+        <field name="arch" type="xml">
+            <xpath expr="//div[@name='options']" position="inside">
+                <div attrs="{'invisible': [('type', '!=', 'service')],}">
+                    <field name="is_contract" />
+                    <label for="is_contract" />
+                </div>
+            </xpath>
```
sbidoul, line 18:
```suggestion
            <div name="options" position="inside">
                <span class="d-inline-block" attrs="{'invisible': [('type', '!=', 'service')]}">
                    <field name="is_contract"/>
                    <label for="is_contract"/>
                </span>
            </div>
```

This looks better

sbidoul: But this can be done in another PR.

product_contract/models/sale_order.py
```diff
@@ -84,8 +82,13 @@ def action_create_contract(self):
                     raise ValidationError(
                         _(
                             "You must specify a contract "
-                            "template for '{}' product in '{}' company."
-                        ).format(order_line.product_id.name, rec.company_id.name)
+                            "template for '%(product_name)s' product "
+                            "in '%(company_name)s' company."
+                        )
```
rousseldenis, line 87: Prefer using translation method attributes.

`_("You must specify (...) %(product_name)s (...)", product_name=(...), company_name=(...))`

rousseldenis (commented):
Code review. Changes seems legit

sbidoul (approved):
As this is PR is identical to the v17 branch, I think the rest can be done in followup PRs.

rousseldenis (approved):
We can move forward and do improvements in further PR's

PR #960, opened 2023-06-09
[14.0] add contract_split module

damdam-s:


damdam-s:
this PR includes a fix for base `contract` module tests. See [2cfbf0c123ef54555378b87e9acda54860d55a5b](https://github.com/OCA/contract/pull/960/commits/2cfbf0c123ef54555378b87e9acda54860d55a5b)

contract/tests/test_contract.py
```diff
@@ -2456,4 +2456,7 @@ def test_automatic_price_with_specific_uom(self):
 
     def test_check_month_name_marker(self):
         invoice_id = self.contract3.recurring_create_invoice()
-        self.assertEqual(invoice_id.invoice_line_ids[0].name, "Header for May Services")
```
rousseldenis, line 2459: @damdam-s Because we are in June :smile:

rousseldenis:
@damdam-s Could you just rewrite your commit messages in order to mention the module concerned (like [FIX] contract: ...)? Thanks

contract/tests/test_contract.py
damdam-s: yep fixed by https://github.com/OCA/contract/pull/960/files#diff-da995126f9f1cfc9e7074016896e39baf645c3c59d1e5364bdb098a4fb7248a9R2461 as long as tests remains in English :)

contract_split/models/contract.py
```diff
@@ -0,0 +1,26 @@
+# Copyright 2023 Damien Crier - Foodles
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class Contract(models.Model):
+    _inherit = "contract.contract"
+
+    original_contract_id = fields.Many2one(
```
hparfr: original_contract_ids ? computed from contract line.

It will be future proof for contract_merge

hparfr:
what do you think of : 
- 1) only choose lines to move instead of entering quantity
- 2) allow to choose between move and plan successor (+ date) 
- 3) move the logic in contract or contract_line and use the wizard only to select what to do

damdam-s:
> what do you think of :
> 
> * 1. only choose lines to move instead of entering quantity
> * 2. allow to choose between move and plan successor (+ date)
> * 3. move the logic in contract or contract_line and use the wizard only to select what to do

1. I can default the quantity to split to the original quantity of the contract line to have almost the same behaviour
2. I'll add it in the roadmap
3. I agree

damdam-s:
@hparfr refactoring done. Is it ok for you?

contract/tests/test_contract.py
mourad-ehm: I fix this in this PR https://github.com/OCA/contract/pull/961.
in fact date set on contract level are not taken into account https://github.com/OCA/contract/blob/14.0/contract/models/contract.py#L168

damdam-s:
"replaced" by #967 to use the fix of #961

PR #961, opened 2023-06-19
14.0 fix contract compute recurring next date

mourad-ehm:
The recurring_next_date is depend on several parameters. 
This PR fix the compute correctly the recurring_next_date depending on other fields.
For example when we choose   "monthlylastday" for recurring_rule_type, we expect that recurring_next_date must be set the end of month.

### DIFF
--- a/contract/models/contract_recurrency_mixin.py
+++ b/contract/models/contract_recurrency_mixin.py
@@ -95,9 +95,11 @@ class ContractRecurrencyMixin(models.AbstractModel):
 
     @api.depends(
         "next_period_date_start",
-        "recurring_invoicing_type",
+        "recurring_invoicing_type", 
+        "recurring_invoicing_offset",
         "recurring_rule_type",
         "recurring_interval",
+       "date_end",
     )
     def _compute_recurring_next_date(self):
         for rec in self:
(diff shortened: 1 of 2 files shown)


mourad-ehm:
Hi @damdam-s, @rousseldenis  can you review this PR ?

damdam-s (approved):

contract/models/contract_recurrency_mixin.py
```diff
@@ -93,9 +93,16 @@ class ContractRecurrencyMixin(models.AbstractModel):
         string="Last Date Invoiced", readonly=True, copy=False
     )
 
-    @api.depends("next_period_date_start")
+    @api.depends(
+        "next_period_date_start",
+        "recurring_invoicing_type",
+        "recurring_rule_type",
+        "recurring_interval",
+    )
```
petrus-v: I'm wondering why all params send to the `get_next_invoice_date` are not present in this list?

should we add those ones, but I miss a bit of context here maybe we don't expect those fields update  while this method is to be called:
```suggestion
        "next_period_date_start",
        "recurring_invoicing_type", 
        "recurring_invoicing_offset",
        "recurring_rule_type",
        "recurring_interval",
       "date_end",
    )
```

petrus-v (approved):
LGTM

damdam-s:
@mourad-ehm now you fixed a date in february, you may want to update the test to fix the test here https://github.com/OCA/contract/actions/runs/5357590412/jobs/9718658755#step:8:401

damdam-s:
https://github.com/OCA/contract/pull/966 have been merged. probably a rebase will be OK

PR #963, opened 2023-06-20
[16.0][FIX] contract_sale: Menu sequence and remove auto install

CRogos:
[FIX] Update sale menu sequence and remove auto install

In version v15, the menu order was:
image

With the change to v16 it is changed to this:
image

This patch restores the order as in v15.

I also see no reason why sale_contract needs to be installed when installing contract. Therefore removed auto_install.

### DIFF
--- a/contract_sale/__manifest__.py
+++ b/contract_sale/__manifest__.py
@@ -3,7 +3,7 @@
 
 {
     "name": "Contract from Sale",
-    "version": "16.0.1.0.0",
+    "version": "16.0.1.0.1",
     "category": "Sales",
     "author": "Tecnativa, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/contract",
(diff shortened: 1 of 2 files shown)


CRogos:
@pedrobaeza could you have a look?

pedrobaeza (approved):

astirpe (approved):

StefanRijnhart:
Isn't this a glue module between sale and contract that is only automatically installed when both of these are installed (and not just when only `contract` is installed? Maybe we want to keep the auto_install.

astirpe:
@StefanRijnhart is not strictly a glue module IMO. Is for making contracts accessible by users with sale permissions but not account permissions. It's possible that not everyone wants it.

CRogos:
> Isn't this a glue module between sale and contract that is only automatically installed when both of these are installed (and not just when only `contract` is installed? Maybe we want to keep the auto_install.

In our company we are only using supplier contracts, but no sale contracts, because we are using subscriptions from odoo enterprise.

StefanRijnhart:
Thanks for clarifying!

StefanRijnhart (approved):

PR #964, opened 2023-06-21
[FIX] dependency on 'dateutil' does not appear to be valid

ozoromo:
Fixes:
`odoo odoo.addons.base.models.ir_module: python external dependency on 'dateutil' does not appear to be a valid PyPI package. Using a PyPI package name is recommended.`

### DIFF
--- a/contract/__manifest__.py
+++ b/contract/__manifest__.py
@@ -18,7 +18,7 @@
     "website": "https://github.com/OCA/contract",
     "depends": ["base", "account", "product", "portal"],
     "development_status": "Production/Stable",
-    "external_dependencies": {"python": ["dateutil"]},
+    "external_dependencies": {"python": ["python-dateutil"]},
     "data": [
         "security/groups.xml",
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Please change the PR title and commit message to one complying with the OCA expected format: 

https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#71commit-message

About the content itself, as `dateutil` is part of Odoo requirements, we can simply remove the line:

https://github.com/odoo/odoo/blob/15.0/requirements.txt#L34

ozoromo:
@pedrobaeza It seems simply removing the line causes errors

pedrobaeza:
Which error?

The commit message is still incorrect.

ozoromo:
> Which error?
https://github.com/OCA/contract/actions/runs/5333205012/jobs/9663374705

I'll close this PR now and reopen it in a second from another branch to avoid changes to Nitrokey:15.0 being added to this PR

PR #965, opened 2023-06-21
[FIX] dependency on 'dateutil' does not appear to be valid

ozoromo:
Contract module lists `dateutil` as a python dependency which causes the error :
`python external dependency on 'dateutil' does not appear to be a valid PyPI package.`

To fix it remove the Dependency `dateutils`  in `contract/__manifest__.py` since it is already required by Odoo and does not have to be added here

### DIFF
--- a/contract/__manifest__.py
+++ b/contract/__manifest__.py
@@ -18,7 +18,6 @@
     "website": "https://github.com/OCA/contract",
     "depends": ["base", "account", "product", "portal"],
     "development_status": "Production/Stable",
-    "external_dependencies": {"python": ["dateutil"]},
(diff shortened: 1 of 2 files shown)


pedrobaeza:
The commit message is still incorrect. It should be something like:

```
[FIX] contract: 'dateutil' dependency pypi name
```

rvalyi (approved):

pedrobaeza:
Please squash commits into one and run pre-commit locally or remove the line in requirements.txt.

pedrobaeza (approved):
/ocabot merge patch

PR #966, opened 2023-06-23
[14.0][FIX] contract: Set fixed date to check test correctly.

victoralmau:
Set fixed date to check test correctly (related to https://github.com/OCA/contract/pull/946).

We need to set a fixed date for `teststest_check_month_name_marker` to check against the month of the date we are set (otherwise it would check against the month of today).

Please @pedrobaeza and @carlosdauden  can you review it?

@Tecnativa

### DIFF
--- a/contract/tests/test_contract.py
+++ b/contract/tests/test_contract.py
@@ -2454,6 +2454,10 @@ class TestContract(TestContractBase):
         self.acct_line.refresh()
         self.assertEqual(self.acct_line.price_unit, 30.75 * 8)
 
+    @freeze_time("2023-05-01")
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):
Is this for avoiding the problem around midnight? If so, please comment whole problem on the commit message.

victoralmau:
> Is this for avoiding the problem around midnight? If so, please comment whole problem on the commit message.

No, it is a problem specific to `teststest_check_month_name_marker` and the month being checked.

pedrobaeza:
Ok, thanks for the extra commit message.

Fast tracking:

/ocabot merge patch

PR #967, opened 2023-06-23
[14.0][ADD] contract_split module

damdam-s:
based on #961 for contract test fix
superseed #960

### DIFF
--- a/contract_split/models/contract.py
+++ b/contract_split/models/contract.py
@@ -1,16 +1,45 @@
 # Copyright 2023 Damien Crier - Foodles
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo import fields, models
+from odoo import api, fields, models
 
 
 class Contract(models.Model):
     _inherit = "contract.contract"
 
-    original_contract_id = fields.Many2one(
+    original_contract_ids = fields.Many2many(
         comodel_name="contract.contract",
+        relation="contract_split_contract_rel",
+        column1="contract_id",
+        column2="split_contract_id",
         readonly=True,
     )
+
+    @api.model
+    def _get_contract_split_name(self, split_wizard):
+        return split_wizard.main_contract_id.name
+
+    @api.model
+    def _get_values_create_split_contract(self, split_wizard):
+        return {
+            "name": self._get_contract_split_name(split_wizard),
+            "partner_id": split_wizard.partner_id.id,
+            "invoice_partner_id": split_wizard.invoice_partner_id.id,
+            "original_contract_ids": [split_wizard.main_contract_id.id],
+            "line_recurrence": True,
+        }
+
+    def _get_default_split_values(self) -> dict:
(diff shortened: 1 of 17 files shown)


contract_split/tests/__init__.py
petrus-v, line 1: I suppose you need import to get your test launch by CI and get proper coverage !

contract_split/models/contract.py
```diff
@@ -0,0 +1,78 @@
...
+
+class ContractLine(models.Model):
+    _inherit = "contract.line"
+
+    splitted_from_line_id = fields.Many2one(
+        comodel_name="contract.line",
+        readonly=True,
+    )
+    splitted_from_contract_id = fields.Many2one(
+        comodel_name="contract.contract",
```
petrus-v, line 53: We could probably use a related field here using `splitted_from_line_id.contract_id` ?

contract_split/models/contract.py
```diff
@@ -0,0 +1,78 @@
...
+
+    @api.model
+    def _get_values_create_split_contract(self, split_wizard):
+        return {
+            "name": self._get_contract_split_name(split_wizard),
+            "partner_id": split_wizard.partner_id.id,
+            "invoice_partner_id": split_wizard.invoice_partner_id.id,
+            "original_contract_ids": [split_wizard.main_contract_id.id],
+            "line_recurrence": True,
+        }
```
petrus-v, line 30: I don't really understand why thoses methods are defined on the contract.contract ?

Probably I miss something but I've the feeling that it get more interest on the wizard and would make code easier to read ?!

contract_split/wizard/wizard_split_contract.py
```diff
@@ -0,0 +1,124 @@
...
+                contract_obj._get_values_create_split_contract(self)
+            )
+            # TODO: play onchange on partner_id. use onchange_helper from OCA ?
+            for line in self.split_line_ids:
+                original_line = line.original_contract_line_id
+                if not line.quantity_to_split:
+                    continue
+                if (
+                    float_compare(
+                        line.quantity_to_split, line.original_qty, precision_digits=2
```
petrus-v: should we avoid to hard code the precision and use the one defined on the unit of the product linked to the line ?

petrus-v (changes requested):
I ask change mainly to ensure test are launched by CI otherwise mainly suggested improvement that I won't lock on it

mourad-ehm (approved):

contract_split/models/contract.py
petrus-v: ok I realize if tomorrow we implement a move we know that split came from such contract !

petrus-v (approved):
Code reviews, looks good to me ! Thanks

contract_split/models/contract.py
damdam-s: yes and if the line is completly moved from one contract to another, there is no origin contract line

PR #970, opened 2023-06-26
Bugfix/plantx 502

aswani-beo:
[FIX] contract: resolved error when trying to create invoice out of contract


mrsavage (approved):

rousseldenis:
@aswani-beo You cannot make a change and remove tests.

Moreover, I would say you need to write one that highlight the problem you have.

Don't put other changes (pre-commit/crowdin ...)

PR #975, opened 2023-07-13
[12.0] FIX contract: allow to view and modify last_date_invoiced

eLBati:
This is needed when invoice date is not coherent with contract line, so that user can adjust it manually

### DIFF
--- a/contract/views/contract_line.xml
+++ b/contract/views/contract_line.xml
@@ -23,8 +23,8 @@
                            attrs="{'required': [('is_auto_renew', '=', True)]}"/>
                     <field name="next_period_date_end"/>
                 </group>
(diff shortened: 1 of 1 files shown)


rousseldenis:
@eLBati This has already been discussed and attended through https://github.com/OCA/contract/tree/14.0/contract_update_last_date_invoiced

eLBati:
@rousseldenis thanks! So you suggest to port `contract_update_last_date_invoiced` to v12, right?

rousseldenis:
> @rousseldenis thanks! So you suggest to port `contract_update_last_date_invoiced` to v12, right?

Indeed 😄

eLBati:
Moving to https://github.com/OCA/contract/pull/976

PR #995, opened 2023-10-02
[14.0][FIX] contract: Apply some values directly in Form

pedrobaeza:
If not applied, they will not be taken into account in the lines operations.

We have needed to add an auxiliary form that adds the fields without groups.

@Tecnativa TT45329

### DIFF
--- a/contract/__manifest__.py
+++ b/contract/__manifest__.py
@@ -36,6 +36,7 @@
         "wizards/contract_manually_create_invoice.xml",
         "wizards/contract_contract_terminate.xml",
         "views/contract_tag.xml",
+        "views/account_move_views.xml",
         "views/assets.xml",
         "views/abstract_contract_line.xml",
         "views/contract.xml",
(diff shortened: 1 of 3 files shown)


pedrobaeza:
WIP, as we need to make sure the field is in the view for tests not to fail.

carlosdauden (approved):

carolinafernandez-tecnativa:
@pedrobaeza

1- I have created a Customer Contract with this information:

image

2- Press "Create invoices" button and I got this error message:

image

3- Customer "Azure Interior" by default has this pricelist which is different as established in Customer Contract

image


So, it is correct that you should add pricelist_id in the view as you said before.

pedrobaeza:
Well, `pricelist_id` is not a standard field, but added by `account_invoice_pricelist`, so this one is restricted now to standard fields, and the pricelist one will be added in a hook.

carolinafernandez-tecnativa (approved):
TT45329

PR #857, opened 2022-07-20
[14.0][IMP] contract: Set default name to '/' for newly created invoice from contract

MTantin:
No default name is defined on invoice values for creation from contract.
The result is invoice without name for the first of the month and all others with the same invoice number specific at the month "INV/2022/07/00001".

It's better to have default name '/' for all invoices, the name will be recompute at the invoice validation.

PR #997, opened 2023-10-04
[15.0][FIX] contract: Apply some values directly in Form + default name

pedrobaeza:
Forward-port of #995 and #857

----

[FIX] contract: Apply some values directly in Form

If not applied, they will not be taken into account in the lines operations.

We have needed to add an auxiliary form that adds the fields without groups.

----

[IMP] contract: Set default name to '/' for newly created invoice from contract

No default name is defined on invoice values for creation from contract.
The result is invoice without name for the first of the month and all others with the same invoice number specific at the month "INV/2022/07/00001".

It's better to have default name '/' for all invoices, the name will be recompute at the invoice validation.

----

@Tecnativa Tecnativa-Task #45329

### DIFF
--- a/contract/__manifest__.py
+++ b/contract/__manifest__.py
@@ -37,6 +37,7 @@
         "wizards/contract_contract_terminate.xml",
         "views/contract_tag.xml",
         "views/abstract_contract_line.xml",
+        "views/account_move_views.xml",
         "views/contract.xml",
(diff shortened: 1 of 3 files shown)


carlosdauden (approved):

carolinafernandez-tecnativa (approved):
TT45329

PR #998, opened 2023-10-08
contract: Translations fix

hhgabelgaard:


contract/models/contract.py
```diff
@@ -614,14 +614,14 @@ def _add_contract_origin(self, invoices):
                 move.message_post(
                     body=(
                         _(
-                            (
-                                "%(msg)s by contract <a href=# data-oe-model=contract.contract"
-                                " data-oe-id=%(contract_id)d>%(contract)s</a>."
-                            ),
-                            msg=move._creation_message(),
```
pedrobaeza, line 621: AFAIK, this one is correct.

contract/models/contract_line.py
```diff
@@ -889,9 +889,9 @@ def cancel(self):
             raise ValidationError(_("Cancel not allowed for this line"))
         for contract in self.mapped("contract_id"):
             lines = self.filtered(lambda l, c=contract: l.contract_id == c)
-            msg = _(
-                "Contract line canceled: %s",
-                "<br/>- ".join(
+            msg = (
+                _("Contract line canceled: %s")
+                % "<br/>- ".join(
```
pedrobaeza: And this syntax is also valid and don't included in the translation. Have you tried?

contract/models/contract.py
hhgabelgaard: I don't think so. It's not in .pot file

pedrobaeza: Uhm, but you didn't update POT files with this PR to check if it's working.

contract/models/contract_line.py
hhgabelgaard: Your right, This is included in the .pot file
I'll remove my changes here

contract/models/contract.py
hhgabelgaard: Yes, the change is included in a regenerated .pot file.
Do we update .pot/.po file by PR's?
Or only by the GH action and weblate?

pedrobaeza: Well, I wanted to make sure is working, but on merge, they will be regenerated.

pedrobaeza (approved):

PR #1005, opened 2023-10-11
[IMP] contract: Allow to single invoice a contract by users 

etobella:
It should allow you to process sale orders and other kinds, as it is using the cron function

@dixmit


pedrobaeza:
Why this? You already have the button to manually invoice it on the contract itsel.

etobella:
That button is only accessible on debug mode, and it does not verify dates, so, If I push the  button three times It will create the invoice three times.

Wit my new way an standard user can create the invoices without worrying if he/she is doing something wrong

pedrobaeza:
Then I think it's better to fine tune that button, showing it to billing users, and showing the popup to decide if force or not to make the invoice.

etobella:
So, should I keep my button and remove the old one?

pedrobaeza:
I think it should be only one button to do both things according a check saying to force or not.

contract/models/contract.py
```diff
@@ -574,6 +574,9 @@ def recurring_create_invoice(self):
         This method triggers the creation of the next invoices of the contracts
         even if their next invoicing date is in the future.
         """
+        _logger.warning(
+            "recurring_create_invoice is no longer necessary. It should be removed on v17"
```
petrus-v: I've the feeling that we should remove it if it's not used anymore ?

Other wise I would expect kind of clear message such as:
```suggestion
            "recurring_create_invoice is deprecated in favor of _recurring_create_invoice instead"
```

damdam-s:
functionnal test OK

contract/models/contract.py
etobella: Yes, it is much cleaner :smile:

giarve (approved):

damdam-s:
any issue w/ this one?

etobella:
Just waiting for approvals

damdam-s:
ok thanks
@etobella FYI i'm currently working on a refactor for the cron to use the "wizard" method to have the same behaviour from the cron and from the wizard. Any objection?

etobella:
good for me

damdam-s:
@etobella here is the PR against this PR https://github.com/dixmit/contract/pull/2

etobella:
@pedrobaeza I think the changes fits your requests :smile:

contract/models/contract.py
```diff
@@ -633,17 +664,24 @@ def _get_recurring_create_func(self, create_type="invoice"):
             return self.__class__._recurring_create_invoice
 
     @api.model
-    def _cron_recurring_create(self, date_ref=False, create_type="invoice"):
+    def _cron_recurring_create(
+        self, date_ref=False, create_type="invoice", domain=None
```
pedrobaeza, line 668: Changing the number of arguments is conflicting, as any override of this method won't know about the new argument, and it will fail, or at least, ignore this new argument. Can you avoid it?

contract/tests/test_contract.py
```diff
@@ -317,7 +321,9 @@ def test_contract_invoice_followers(self):
         self.contract.message_subscribe(
             partner_ids=self.contract.partner_id.ids, subtype_ids=subtype_ids
         )
-        self.contract._recurring_create_invoice()
+        self.env["contract.manually.single.invoice"].create(
```
pedrobaeza, line 324: I don't think you should change this test, but only add a new one for the new use case. This also proves that there's no change in the previous behavior for compatibility reasons.

contract/views/contract.xml
```diff
@@ -38,11 +38,11 @@
...
                     <button
-                        name="recurring_create_invoice"
-                        type="object"
-                        attrs="{'invisible': ['|', ('create_invoice_visibility', '=', False),('generation_type','!=','invoice')]}"
-                        string="Create invoices"
-                        groups="base.group_no_one"
+                        name="%(contract.contract_manually_single_invoice_act_window)s"
+                        type="action"
+                        string="Process manually"
+                        context="{'default_contract_id': id, 'default_date': recurring_next_date}"
```
pedrobaeza, line 44: Use `active_id` instead

contract/views/contract.xml
```diff
@@ -38,11 +38,11 @@
...
+                        attrs="{'invisible': [('id', '=', False)]}"
```
pedrobaeza, line 45: Hide it in the generation_type is not invoice.

contract/models/contract.py
```diff
@@ -574,6 +601,10 @@ def recurring_create_invoice(self):
         This method triggers the creation of the next invoices of the contracts
         even if their next invoicing date is in the future.
         """
+        _logger.warning(
```
pedrobaeza, line 604: I don't get why repeating the message_post code.

contract/wizards/contract_manually_single_invoice.xml
```diff
@@ -0,0 +1,41 @@
...
+<!-- Copyright 2023 Dixmit
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+<odoo>
+
+    <record model="ir.ui.view" id="contract_manually_single_invoice_form_view">
+        <field name="name">contract.manually.single.invoice.form (in contract)</field>
+        <field name="model">contract.manually.single.invoice</field>
+        <field name="arch" type="xml">
+            <form string="Contract Manually Single Invoice">
+                <!-- TODO -->
```
pedrobaeza, line 11: ?

contract_sale_generation/models/contract.py
```diff
@@ -70,6 +74,10 @@ def recurring_create_sale(self):
         This method triggers the creation of the next sale order of the
         contracts even if their next sale order date is in the future.
         """
+        _logger.warning(
```
pedrobaeza, line 77: This is for sales, not invoices.

pedrobaeza (changes requested):

PR #1007, opened 2023-10-13
[IMP] contract: Simplify management of recurring_next_date

etobella:
The usage of this field was confusing. It was set as editable, but I found several issues:

On creation, the change was not applied on contract recurring_next_date
On edition of lines, the change was showed, but got back to the computed value when saving.

IMO, it has much more sense to disable the edition. WDYT? May I be missing something?


etobella:
Well, this change might have some unexpected related errors, so I will close it...

PR #1009, opened 2023-10-19
ADD module contract_fixed_discount

damdam-s:


contract_fixed_discount/tests/test_contract_discount_fixed.py
```diff
@@ -0,0 +1,70 @@
...
+        self.assertEqual(self.contract.contract_line_ids.price_subtotal, 50.0)
+
+    def test_price_subtotal_discount_fixed(self):
+        with Form(self.contract) as contract:
+            with contract.contract_line_ids.edit(0) as line:
+                line.discount_fixed = 42
+                line.save()
+        contract.save()
+
+        self.assertEqual(self.contract.contract_line_ids[0].price_subtotal, 58.0)
```
petrus-v: I'm wondering why not using contract4 to run this test ?

petrus-v (approved):
LGTM

contract_fixed_discount/tests/test_contract_discount_fixed.py
damdam-s: changed

petrus-v: :heart: thanks

contract_fixed_discount/views/abstract_contract_line.xml
```diff
@@ -0,0 +1,23 @@
...
+        >contract.abstract.contract.line form view (in contract)</field>
+        <field name="model">contract.abstract.contract.line</field>
+        <field
+            name="inherit_id"
+            ref="contract.contract_abstract_contract_line_form_view"
+        />
+
+        <field name="arch" type="xml">
+            <field name="discount" position="after">
+                <field name="discount_fixed" optional="hide" />
```
petrus-v: testing on runboat :
* this looks like to manage the popup form view which which don't care about optional
* so we should add it on the tree view in the contract form view
* we should probably use same group as the discount fields
* I think we should change the group to rely on the same one present on sale order: `groups="product.group_discount_per_so_line"`

petrus-v: Test on runboat: thanks :heart:

petrus-v (approved):

PR #1012, opened 2023-10-23
IMP: use group product.group_discount_per_so_line instead of base.gro…

damdam-s:
…up_no_one

### DIFF
--- a/contract/views/contract.xml
+++ b/contract/views/contract.xml
@@ -235,6 +235,7 @@
                                     <field
                                         name="discount"
                                         groups="product.group_discount_per_so_line"
+                                        optional="show"
                                     />
                                     <field name="price_subtotal" />
                                     <field name="recurring_interval" invisible="1" />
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):

petrus-v (approved):

pedrobaeza:
There's not too much sense to add also optional of the field. Just don't add the user to the group instead.

damdam-s:
@pedrobaeza I understand you point of view. However, some users in the group may not want to see the discount (for view simplification) and that's why I added the optional. 
Another reason is to match the optional of the discount_fixed field from #1009 but maybe I should add the optional part in the later PR.
Opinion?

pedrobaeza:
OK, my only fear here is that someone hides the column, forget it, and then see that the subtotal is "incorrect" not seeing the discount, but it's not a big deal. Please rebase your branch to avoid merge commits though.

PR #1013, opened 2023-10-26
[16.0][ADD] Module contract_invoice_auto_post: to 16.0

manuelcalerosolis:
This module gets the value of the auto post variable in client contract and it automatically transfers the value of the variable to the invoice. Moreover, it will check periodically the status of this value and will confirm the invoice when it is programed to happen.


carlosdauden (approved):
The commit message should look something like this:
[ADD] contract_invoice_auto_post: New module
The PR title should look something like this:
[16.0][ADD] contract_invoice_auto_post: New module

pedrobaeza:
I think this can be integrated in the main module, being an optional check. What I don't see is the cron process. The posting should be done immediately after creating the invoice, so no need of a later process.

manuelcalerosolis:
@pedrobaeza
I also thought about that behavior, but I decided to use the workflow that Odoo has in creating invoices, in addition to having the same options as the standard one.

pedrobaeza:
AFAIK, that flow is for posting journal entries in the future, but I don't think here applies.

PR #1016, opened 2023-11-02
[MIG] agreement_rebate: Migration to 14.0

Latour12:


pedrobaeza:
I'm afraid the commit history is missing. You should follow the migration guide, but reversing branches names for getting the history from 15.0.

Latour12:
Absolutely, I'll follow the migration guide with the reversed branch names to correct the commit history. Thank you for the guidance.

PR #1017, opened 2023-11-03
[14.0][MIG] agreement_rebate: Migration to 14.0

Latour12:


giarve (changes requested):
`account_invoice_refund_link` should not be included here

giarve (approved):
Functional OK (tested locally)

PR #1018, opened 2023-11-03
14.0 fix contract manually single invoice

damdam-s:
based on #1005 

review only last commit

if the method creating invoice generates more than one invoice, we have
`[account.move(ID1, ID2)]` for example and the `message_post()` fails w/ a `singleton error`.

Adding a loop level allows to properly post messages no matter of the number of invoices generated by the wizard


damdam-s:
@etobella I don't mind if you cherry-pick this one to your branch 😄

etobella:
Thanks!! I made the cherry-pick and renamed the commit in my branch. Thanks for detecting it :smile:

PR #1019, opened 2023-11-07
[16.0][FIX] contract: Usage of analytic distribution

fkantelberg:
Because of the changes in 16.0 analytic accounting the `analytic_distribution` is more important now and only this will be copied onto the invoices.

I copied the invoice line preparation mostly from `sale.order.line` to generate the distribution also from the `analytic_account_id`. Therefore no migration is required.


CRogos (commented):
I agree to follow the same procedures as in sale.order. I think this PR would fix the issue that the analytic_account_id is not copied to the invoice when used in contract.line.

image

Nevertheless there is a general different between sale.order and contract.contract, where I am not sure if this is intended or should be changed. 
While the analytic_account_id is on **sale.order** the analytic_account_id in contract is on **contract.line** instead of **contract.contract**.
(I even do not understand why there is an analytic_account_id in sale.order at all and not only an analytic_distribution on sale.order.line and that's it ... maybe usability)

Shouldn't we either remove analytic_account_id at all (migrate the data into analytic_distribution) or move analytic_account_id into "contract.contract" to create the same logic as in sale.order? (analytic_account_id on sale.order is continued on 17.0, so there might be a reason I do not understand)

image

I am willing to prepare a PR, but I would like to have some opinions first.
@pedrobaeza @wpichler @rousseldenis

fkantelberg:
`analytic_account_id` on `sale.order` is used and auto-generated for expense products. [1] I don't think that we need this feature because contracts aren't generated on the fly by the system or change to state too often. `analytic_account_id` seems to exist on the contract.line because it was the way before 16.0 and it was just left there during the migration and the distribution was added. For the contract lines we could remove it but have to implement a migration at the same time.

[1] https://github.com/OCA/OCB/blob/16.0/addons/sale/models/sale_order.py#L854

pedrobaeza:
Analytic account on contract lines was there for transferring such analytic line to the invoice lines, so it's something needed. In v16, it should be switched to the analytic distribution, to be transferred the same way.

CRogos:
This script works in an post-migration.py, but I cannot access env["contract.line"] when it is in a pre-migration.py. And the analytic_account_id is not available anymore, when we remove it from contract.line.

Is a better way than creating temporary table in a pre-migration script storing the data and deleting this table in the post-migration step?
Another (ugly) options would be to leave the field in the code and add a comment to remove it with the Odoo 17 mig. 

```
from openupgradelib import openupgrade

def _mig_analytic_account(env):
    """Migrate analytic account from contract line to analytic distribution"""
    contract_lines = (
        env["contract.line"]
        .with_context(active_test=False)
        .search(
            [
                ('analytic_account_id', '!=', False),
            ]
        )
    )
    for line in contract_lines:
        analytic_distribution = line.analytic_distribution
        analytic_account_id = str(line.analytic_account_id.id)
        if analytic_distribution:
            analytic_distribution[analytic_account_id] = (
                analytic_distribution.get(analytic_account_id, 0) + 100
            )
            line.analytic_distribution = analytic_distribution
        else:
            line.analytic_distribution = {analytic_account_id: 100}

@openupgrade.migrate()
def migrate(env, version):
    _mig_analytic_account(env)

```

@fkantelberg do you want to add the discussed changes on this PR or shell I create an updated PR?
- add migration script
- remove analytic_account_id from views

fkantelberg:
@CRogos I added the migration and dropped the `analytic_account_id` field. Can you review it?

CRogos (approved):
LGTM: Code reviewed, tested migration and analytic distribution is copied to invoices.

contract/models/contract_line.py
```diff
@@ -559,6 +555,7 @@ def _prepare_invoice_line(self):
             self.last_date_invoiced, self.recurring_next_date
         )
         name = self._insert_markers(dates[0], dates[1])
+
```
pedrobaeza: Don't introduce a useless line break inside a method (and more being unrelated).

fkantelberg: Removed again. Was left because I could remove the initial code here.

pedrobaeza (approved):
/ocabot merge nobump

PR #1032, opened 2024-01-09
[14.0][FIX] subscription_oca: fix template duration change

PicchiSeba:
Changing the template duration doesn't change the `date` field in model `sale.subscription`

### DIFF
--- a/subscription_oca/models/sale_subscription.py
+++ b/subscription_oca/models/sale_subscription.py
@@ -181,7 +181,7 @@ class SaleSubscription(models.Model):
             slash = "/" if template_code and code else ""
             record.name = "{}{}{}".format(template_code, slash, code)
 
-    @api.depends("template_id", "date_start")
+    @api.depends("template_id.recurring_rule_boundary", "date_start")
     def _compute_rule_boundary(self):
         for record in self:
             if record.template_id.recurring_rule_boundary == "unlimited":


subscription_oca/models/sale_subscription.py
```diff
@@ -181,7 +181,7 @@ def _compute_name(self):
             slash = "/" if template_code and code else ""
             record.name = "{}{}{}".format(template_code, slash, code)
 
-    @api.depends("template_id", "date_start")
+    @api.depends("template_id.recurring_rule_boundary", "date_start")
```
rousseldenis, line 184: @PicchiSeba Not sure about this as a change on the template value will change all subscriptions (is it really wanted ?).

PicchiSeba: @rousseldenis 

I made that change because I found myself in the following scenario:

1. Create a subscription
2. Assign a subscription template having duration set to "Forever"
3. Go back to the subscription and see that the "Finish date" field hasn't been set
4. You realize you made a mistake and the template actually has an end date. So you go to the template and set the duration to "Fixed"  with a rule count of "6 months"
5. Returning to the subscription still shows no end date

I find this to be confusing from a user perspective. I realize this could create issues for Subscriptions having the state "in progress".

We could block writes to a certain template if it is assigned to an active subscription.

I'm really interested in your feedback, and sorry for taking this long to answer.

PicchiSeba: Or block writes at all if it's assigned to at least a subscription

PR #1033, opened 2024-01-09
[15.0][IMP] contract: Protect line recurrrency change

pedrobaeza:
If you have defined several contract lines, each one with their recurrency information, unmarking the check "Line recurrency" by mistake and saving will make you lose the configuration in one shot, and also modify the invoceability conditions.

Thus, it's reasonable to avoid to uncheck the mark once you have at least one invoice issued for a contract with this mark checked.

@Tecnativa TT47015

### DIFF
--- a/contract/views/contract.xml
+++ b/contract/views/contract.xml
@@ -148,7 +148,11 @@
                         </group>
                     </group>
                     <group name="recurring_invoices">
-                        <field name="line_recurrence" class="oe_inline" />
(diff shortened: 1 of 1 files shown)


carolinafernandez-tecnativa (approved):
TT47015

CRogos (approved):
Functional and Code review LGTM

rousseldenis (approved):

PR #1047, opened 2024-02-07
[14.0][IMP]subscription_oca: readonly company on sale_subscription, templates must have the same company or none

PicchiSeba:
1. The company of `sale.subscription` is readonly and set to the current active company.
2. `sale.subscription.template`s can be linked to sale_subscriptions only if they share the same company or if it's not set


subscription_oca/views/sale_subscription_views.xml
```diff
@@ -77,8 +77,6 @@
                                 string="Invoices"
                             />
                         </button>
-
-
```
renda-dev: **nitpick:** please, remove empty changes in this PR

subscription_oca/models/sale_subscription.py
```diff
@@ -44,6 +45,7 @@ class SaleSubscription(models.Model):
         comodel_name="sale.subscription.template",
         required=True,
         string="Subscription template",
+        domain=lambda self: [("company_id", "in", [False, self.env.company.id])],
```
renda-dev: **nitpick:** have you tried using `check_company=True`?

subscription_oca/models/sale_subscription.py
```diff
@@ -551,4 +553,11 @@ def create(self, values):
                 .search([("type", "=", "pre")], order="sequence desc", limit=1)
                 .id
             )
+        template = self.env["sale.subscription.template"].browse(values["template_id"])
+        if template.company_id and self.env.company != template.company_id:
+            raise ValidationError(
+                _(
+                    "Cannot link a Template owned by another company to this subscription"
+                )
+            )
```
renda-dev: **suggestion:** with `check_company` on `template_id`, this shouldn't be needed

renda-dev (commented):
Just a couple of small suggestions, for better scalability

subscription_oca/models/sale_subscription.py
PicchiSeba: Oh I had no idea, thank you so much!

subscription_oca/models/sale_subscription.py
PicchiSeba: @renda-dev 
I tested with and without `check_company=True` but it seems that these lines are still necessary.

Normally a User could assign templates having the same company as the active one or no company at all. In this scenario the check_company line does the job.

But the user can circumvent this check by creating a template on the fly and assigning to it another active company without any error. This is why that check it's still necessary.
I could write a comment if this would make it clearer

renda-dev: I mean, it's a really weird edge case, that i think it should be fixed on the template directly, preventing the user to assign the template to a company without any restriction

PicchiSeba: I found a less invasive way to do what we need by using the context.

Let me know you you think about it

renda-dev: It's great, another solution could be adding just a `_check_company_id` with an `@api.constrains("template_id")` that doesn't let you choose a 'wrong' company if a template is selected :)

That would also cover the editing and not only the creation, but you do you, both are fine by me

PicchiSeba: Oh right! Why didn't I think about that?

Well there we go.

subscription_oca/models/sale_subscription.py
```diff
@@ -161,6 +163,15 @@ def cron_subscription_management(self):
                     subscription.action_start_subscription()
                     subscription.generate_invoice()
 
+    @api.constrains("template_id")
+    def _constrain_template_id(self):
+        if self.template_id.company_id not in [False, self.env.company]:
```
renda-dev: **suggestion:** according to the [guidelines](https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#L902-L903):
```suggestion
    def _check_template_id(self):
        if self.template_id.company_id not in [False, self.env.company]:
```

**question:** have you tested this without selecting any template? does it still work fine?

PicchiSeba: > **question**: have you tested this without selecting any template? does it still work fine?

The template is `required` for all `sale.subscription`s

https://github.com/OCA/contract/blob/29c8c6ce14eec392da7a245c2be66b9ea306955c/subscription_oca/models/sale_subscription.py#L43-L47

renda-dev (approved):
Code review, LGTM

francesco-ooops (approved):
Functional ok!

PR #1048, opened 2024-02-09
[16.0][FIX] contract: Show discount lines for users with the correct group

fkantelberg:
According to the readme [1] the field should be visible if you have the group `Discount on lines` hence it seems like a bug that it's not visible.

[1] `To view discount field in contract line, you need to set *Discount on lines* in
user access rights.`


carlosdauden (approved):

pedrobaeza (approved):
/ocabot merge minor

PR #1050, opened 2024-02-13
[16.0][IMP] contract: Protect line recurrrency change

pedrobaeza:
Forward-port of #1033 

If you have defined several contract lines, each one with their recurrency information, unmarking the check "Line recurrency" by mistake and saving will make you lose the configuration in one shot, and also modify the invoceability conditions.

Thus, it's reasonable to avoid to uncheck the mark once you have at least one invoice issued for a contract with this mark checked.

@Tecnativa

### DIFF
--- a/contract/views/contract.xml
+++ b/contract/views/contract.xml
@@ -148,7 +148,11 @@
                         </group>
                     </group>
                     <group name="recurring_invoices">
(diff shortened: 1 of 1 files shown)


flotho (approved):
easy change,
thanks
LGTM

victoralmau (approved):

PR #1054, opened 2024-03-11
[16.0][IMP] contract and contract_sale_generation: Add translations

aliciagaarzo:
The 'contract' and 'contract_sale_generation' modules have been modified so that the generation type selection values are translatable.


contract/models/abstract_contract.py
```diff
@@ -51,15 +51,11 @@ class ContractAbstractContract(models.AbstractModel):
...
     generation_type = fields.Selection(
-        selection=lambda self: self._selection_generation_type(),
+        selection=[("invoice", "Invoice")],
         default=lambda self: self._default_generation_type(),
         help="Choose the document that will be automatically generated by cron.",
     )
 
-    @api.model
-    def _selection_generation_type(self):
-        return [("invoice", "Invoice")]
```
amarcosg, line 61: Deleting the function you may break other funcionality that depends on it. Dont delete de function, just add the _ in the option

```
    @api.model
    def _selection_generation_type(self):
        return [("invoice", _("Invoice"))]
```

contract_sale_generation/models/abstract_contract.py
```diff
@@ -2,16 +2,14 @@
...
+from odoo import fields, models
 
 
 class ContractAbstractContract(models.AbstractModel):
     _inherit = "contract.abstract.contract"
 
     sale_autoconfirm = fields.Boolean()
 
-    @api.model
-    def _selection_generation_type(self):
```
amarcosg, line 14: Same dont delete the function just add the _ `_("Sale")`

amarcosg (changes requested):
With the requested changes you may need to reexport the translations

contract_sale_generation/models/abstract_contract.py
amarcosg: I tried to workaround it but the problem seems related to Selection translations in Abstract model, which is the case. https://github.com/odoo/odoo/issues/64065

Taking that in account for me the change is OK

amarcosg (approved):
LGTM

flalexg:
Functional Review. LGTM.
image
image

flalexg (approved):

pedrobaeza (approved):
/ocabot merge minor

PR #1055, opened 2024-03-11
[FIX] subscription_oca: migration script for rules

GSLabIt:


renda-dev:
thanks for the contribution!

could you please explain why this is needed? those rules should be created right away after updating to version `14.0.1.3.0` (https://github.com/OCA/contract/pull/1047)

PR #1056, opened 2024-03-12
[15.0][IMP] contract: Remove buttons from contract lines if recurrence is not per line

victoralmau:
Remove buttons from contract lines if recurrence is not per line

If we manage the contract at header level, we must set the end date and last invoice date there, otherwise data will be inconsistent.

@Tecnativa TT48273


pedrobaeza (approved):

carolinafernandez-tecnativa (approved):
Functional review LGTM :) Thanks

PR #1059, opened 2024-03-14
[16.0][FIX] contract: Fix modifications layout

pedrobaeza:
On the migration to v16, the primary layout view was incorrectly removed. It's needed because of 2 reasons:

- It's referenced in the following template doing a t-call. Well, this one can be avoided calling the new layout upstream template, but...
- This layout is done for placing some extra content (the modifications table), as part of the layout that is not shown in the chatter for avoiding noise.

We recreate this template inheriting the new layout template, and adapting it properly.

Before:

imagen

After:

imagen

@Tecnativa TT45646

### DIFF
--- a/contract/data/template_mail_notification.xml
+++ b/contract/data/template_mail_notification.xml
@@ -1,5 +1,14 @@
 <?xml version="1.0" ?>
 <odoo>
+    <template
+        id="mail_notification_contract"
+        inherit_id="mail.mail_notification_layout"
+        primary="True"
+    >
+        <xpath expr="//div[@t-out='message.body']" position="after">
+            <t t-raw="0" />
+        </xpath>
+    </template>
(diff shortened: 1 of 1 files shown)


carolinafernandez-tecnativa:
Hi i have tried on my local and in oca runboat but im not being able to see the table.

Check this example:

image

I think it is sending email after check "Send" moves to True and that's why im not being able to see it when sending email.

pedrobaeza:
The example is correct and that's the intended way to work: we don't want the table to be shown in the chatter. And you don't need to try locally. Runboat has a Mailhog incorporated to test output mails.

carolinafernandez-tecnativa (approved):
I have tested with mailHog. Functional review LGTM :) Thanks!

victoralmau (approved):
Functional review OK

PR #1061, opened 2024-03-16
[16.0][IMP] contract: Remove buttons from contract lines if recurrence is not per line

pedrobaeza:
Forward-port of #1056 

If we manage the contract at header level, we must set the end date and last invoice date there, otherwise data will be inconsistent.

@Tecnativa TT48273


carolinafernandez-tecnativa (approved):
Functional review LGTM :) Thanks!

PR #1051, opened 2024-02-15
[17.0][MIG] contract: Migration to 17.0

peluko00:

PR #1067, opened 2024-03-23
[17.0][MIG] contract: Migration to 17.0 

CRogos:
Supersede #1051 

- cleanup commit history
- added missing commits
- fix errors on views


contract/static/src/js/section_and_note_fields_backend.js
CRogos, line 1: Is this code needed?

contract/static/src/js/contract_portal_tour.esm.js
CRogos, line 1: This tour isn't working.

contract/tests/test_portal.py
```diff
@@ -14,12 +14,7 @@ def test_tour(self):
         )
         user_portal = self.env.ref("base.demo_user0")
         contract.message_subscribe(partner_ids=user_portal.partner_id.ids)
-        self.browser_js(
-            "/",
-            "odoo.__DEBUG__.services['web_tour.tour'].run('contract_portal_tour')",
-            "odoo.__DEBUG__.services['web_tour.tour'].tours.contract_portal_tour.ready",
-            login="portal",
-        )
+        self.start_tour("/", "contract_portal_tour", login="portal")
```
CRogos, line 17: This test is still not working. anyone an idea how to fix this?

CRogos:
@pedrobaeza do you have a hint what I need to do to get the last test valid. 

It is a "tour" which does not recognize the end state. I've already tried a lot, but couldn't find the problem.

pedrobaeza:
I don't know very much JS tours. Maybe my colleague @chienandalu can say

contract/static/src/js/contract_portal_tour.esm.js
```diff
@@ -0,0 +1,18 @@
+/** @odoo-module **/
+
+import {registry} from "@web/core/registry";
+
+registry.category("contract.tour").add("contract_portal_tour", {
```
chienandalu: You're not adding the tour to the tours registry...
```suggestion
registry.category("web_tour.tours").add("contract_portal_tour", {
```

CRogos: always that simple... it is working :) thank you!

CRogos:
OK finished testing, please review.

MohamedOsman7 (approved):
Code & Functional review LGTM

contract/models/abstract_contract_line.py
```diff
@@ -0,0 +1,269 @@
...
+    )
+    # Just to have a currency_id here - will get overwriten in contract.line
+    # model with the related currency from the contract
+    currency_id = fields.Many2one("res.currency")
+    specific_price = fields.Monetary()
+    price_unit = fields.Monetary(
+        string="Unit Price",
+        compute="_compute_price_unit",
+        inverse="_inverse_price_unit",
+    )
```
CRogos: @pedrobaeza what do you think about changing this to float regarding this conversation: https://github.com/OCA/account-invoicing/pull/1674#pullrequestreview-1975650402

pedrobaeza: OK, agreed.

pedrobaeza:
Please include #1069

rousseldenis (approved):
Code review

CRogos:
@pedrobaeza could you merge?

contract/tests/test_multicompany.py
```diff
@@ -8,12 +8,13 @@ class ContractMulticompanyCase(TestContractBase):
...
     def setUpClass(cls):
         super().setUpClass()
-        chart_template = cls.env.ref("l10n_generic_coa.configurable_chart_template")
         cls.company_obj = cls.env["res.company"]
         cls.company_1 = cls.env.ref("base.main_company")
         vals = {"name": "Company 2"}
         cls.company_2 = cls.company_obj.create(vals)
-        chart_template.try_loading(company=cls.company_2)
+        cls.env["account.chart.template"].try_loading(
+            "generic_coa", company=cls.company_2, install_demo=False
```
pedrobaeza: The module shouldn't load generic COA. Is it really needed to create a new company for the tests?

pedrobaeza: What I mean is that there shouldn't be a hidden dependency over `l10n_generic_coa` for tests to work.

CRogos: how about this solution?

pedrobaeza: What happens if there's no `l10n_xx` installed? I repeat my initial question: Is it really needed to create a new company for the tests?

CRogos: I did not write this test. But my assumption was that you need 2 companies for a multicompany test. 
I guess in the default installation there is only one company, correct? This would be not enough for a multi company field.

Deleting the test would also prevent creating a second company, but I don't think you have this in mind. I need more than a question I cannot answer.

pedrobaeza: Yeah, I suppose this came as a regression test for a multi-company problem in the past, but it's better to keep it. Let's continue with your last change, and see when integrating this module in future installations my CI blows away.

pedrobaeza (approved):
/ocabot merge nobump

PR #1071, opened 2024-04-10
[16.0][IMP] contract: solve error for last_date_invoiced field

aliciagaarzo:
When working with contracts, an error has been found in the creation of sales/invoice when the end date on the lines is exceeded.

## To Reproduce

1. Create a contract and add start date

2. Add an end date (check that the value is passed to the lines)

3. Create contracts until we reach the end date and the error occurs

A small change has been made to the code to create order/invoice when lines exceed the end date, which is what causes the error.

### DIFF
--- a/contract/models/contract.py
+++ b/contract/models/contract.py
@@ -514,6 +514,7 @@ class ContractContract(models.Model):
                 not contract_line.is_canceled
                 and contract_line.recurring_next_date
                 and contract_line.recurring_next_date <= date_ref
+                and contract_line.next_period_date_start
             )
 
         lines2invoice = previous = self.env["contract.line"]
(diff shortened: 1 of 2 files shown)


pedrobaeza:
The videos are not correct, but anyway, please reflect the steps to reproduce the problem with text, and add a regression tests exercising it.

aliciagaarzo:
Hi @pedrobaeza ,
I have reviewed and with a small change the error is resolved and resolved. I have verified that the tests were already failing and now they work without problem.

dalonsofl (approved):

pedrobaeza (approved):
I think the problem is because the end date doesn't match with the end of the period. In that case, I think there is going to be a new invoice for that "half-invoice"

Anyway, better to not crash:

/ocabot merge patch

pedrobaeza:
Please fw-port it to 17.0

PR #1075, opened 2024-04-15
[17.0][IMP] contract: solve error for last_date_invoiced field

aliciagaarzo:
Transfer bug fix from v16 to v17, an error has been found in the creation of sales/invoice when the end date on the lines is exceeded.

## To Reproduce

1. Create a contract and add start date

2. Add an end date (check that the value is passed to the lines)

3. Create contracts until we reach the end date and the error occurs

A small change has been made to the code to create order/invoice when lines exceed the end date, which is what causes the error.


pedrobaeza (approved):
Thanks for the fw-port. Fast-tracking as mostly equal to v16 one:

/ocabot merge patch

PR #1076, opened 2024-04-16
[FIX] product-contract: non-existent external dependency in manifest

Pexers:
The `dateutil` is a non-existent dependency. This modification replaces it by `python-dateutil`.
- https://pypi.org/project/python-dateutil/


Pexers:
@sbejaoui ? 😄

PR #1080, opened 2024-04-24
[16.0][IMP] contract: Big changes in recurrency fields 

flalexg:
Modules instaled:
* contract
* contract_sale
* contract_sale_generation

Description of the problem(s):
TO report this problem, I changed to invisible fields to visible in the view in order to see ther error better.
We have made all the screenshots generating invoices, but it also occurs when generating sales.

### Problem 1
If you add the products and then you change the recurrency (for exemple Every 1 week), the recurrency fields are not updated correctly.

1. We add a product.
image

2. We change the recurrency
image

3. We save and we create 3 invoices
image

As we can see, the recurrence is every 1 Month and not every 1 Week.

### Problem 2

The contract never ends.

1. We create this contract with a Date End.
image

2. We add some products (in order to avoid the problem above).
image

As we can see, the Date End is not set on the lines. (origin of the error).

3. We generate infinite invoices.

image

We are able to generate infinite invoices. Maybe the scheduled action controls this, but this should be controlled manually also.

### Problem 3

Date end dissapear when changing the recurrency.

1. We create a contract. It must has a Date End. We save it.
image

2. We want to change de recurrency (it can happens). We notice the date end dissapears when we save the changes.
image

### Problem 4

When having 2 or more lines, the last date invoice field doesn't compute correctly.

1. We create a contract. To see more clear the error, we set the recurrency every 1 week.
image

2. We create 1 invoice.

image

As we can see, the field last_date_invoice is mismatching between the 2 lines, when it should be the same because we aren't in a recurrency line level. If we had set a date end, we would be see an error at the last invoices, where only one product would be invoiced due to this error.

After exposing this, I made some several changes in the structure of the module contract.

### Problem 5
If we put a Date End and we change the Date Of Next Invoice manually, the date end dissapears

1. We create a contract with Date End
image

2. We manually change the Date Of Next Invoice and we save it.

image

As you can see, the Date End dissapears.

### Problem 6
We can't change the Date of Next Invoice in recurrence at line level

1. We create a contrat with recurrence at line level
image

2. We create a line. We only establish a product and we save.
image

3. We edit the line and we stablish the 'Date of Next Invoice' on day later. We save and close. The Next Period End is correctly changed.
image

4. We open the line again.
image

As we can see, the Date of Next Invoice is not changed making the next period end beig bad calculated.

### DIFF
--- a/contract/models/abstract_contract.py
+++ b/contract/models/abstract_contract.py
@@ -80,3 +80,20 @@ class ContractAbstractContract(models.AbstractModel):
                 contract.journal_id = journal.id
             else:
                 contract.journal_id = None
+
+    @api.onchange(
+        "date_start",
+        "date_end",
+        "recurring_interval",
+        "recurring_rule_type",
+        "recurring_invoicing_type",
+    )
+    def _onchange_recurrency_fields(self):
+        if self.contract_line_fixed_ids:
+            self.contract_line_fixed_ids.date_start = self.date_start
+            self.contract_line_fixed_ids.date_end = self.date_end
+            self.contract_line_fixed_ids.recurring_interval = self.recurring_interval
+            self.contract_line_fixed_ids.recurring_rule_type = self.recurring_rule_type
+            self.contract_line_fixed_ids.recurring_invoicing_type = (
+                self.recurring_invoicing_type
+            )
--- a/contract/models/abstract_contract_line.py
+++ b/contract/models/abstract_contract_line.py
@@ -85,6 +85,12 @@ class ContractAbstractContractLine(models.AbstractModel):
         readonly=False,
         copy=True,
     )
+    date_end = fields.Date(
+        compute="_compute_date_end",
+        store=True,
+        readonly=False,
+        copy=True,
+    )
     last_date_invoiced = fields.Date()
     is_canceled = fields.Boolean(string="Canceled", default=False)
     is_auto_renew = fields.Boolean(string="Auto Renew", default=False)
@@ -166,11 +172,9 @@ class ContractAbstractContractLine(models.AbstractModel):
     def _compute_date_start(self):
         self._set_recurrence_field("date_start")
 
-    # pylint: disable=missing-return
-    @api.depends("contract_id.recurring_next_date", "contract_id.line_recurrence")
-    def _compute_recurring_next_date(self):
-        super()._compute_recurring_next_date()
-        self._set_recurrence_field("recurring_next_date")
+    @api.depends("contract_id.date_end", "contract_id.line_recurrence")
+    def _compute_date_end(self):
+        self._set_recurrence_field("date_end")
 
     @api.depends("display_type", "note_invoicing_mode")
     def _compute_is_recurring_note(self):
--- a/contract/models/contract_recurrency_mixin.py
+++ b/contract/models/contract_recurrency_mixin.py
@@ -48,6 +48,7 @@ class ContractRecurrencyBasicMixin(models.AbstractModel):
         help="Invoice every (Days/Week/Month/Year)",
     )
     date_start = fields.Date()
+    date_end = fields.Date()
     recurring_next_date = fields.Date(string="Date of Next Invoice")
 
     @api.depends("recurring_invoicing_type", "recurring_rule_type")
@@ -78,7 +79,10 @@ class ContractRecurrencyMixin(models.AbstractModel):
 
     date_start = fields.Date(default=lambda self: fields.Date.context_today(self))
     recurring_next_date = fields.Date(
-        compute="_compute_recurring_next_date", store=True, readonly=False, copy=True
+        compute="_compute_recurring_next_date",
+        store=True,
+        readonly=False,
+        copy=True,
     )
     date_end = fields.Date(index=True)
     next_period_date_start = fields.Date(
--- a/contract/views/contract.xml
+++ b/contract/views/contract.xml
@@ -255,6 +255,11 @@
                                         groups="base.group_no_one"
                                     />
                                     <field
+                                        name="next_period_date_start"
+                                        invisible="1"
+                                    />
+                                    <field name="next_period_date_end" invisible="1" />
+                                    <field
                                         name="create_invoice_visibility"
                                         invisible="1"
                                     />
--- a/contract/models/contract.py
+++ b/contract/models/contract.py
@@ -80,7 +80,6 @@ class ContractContract(models.Model):
     create_invoice_visibility = fields.Boolean(
         compute="_compute_create_invoice_visibility"
     )
-    date_end = fields.Date(compute="_compute_date_end", store=True, readonly=False)
     payment_term_id = fields.Many2one(
         comodel_name="account.payment.term", string="Payment Terms", index=True
     )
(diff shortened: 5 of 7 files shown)


javierjcf:
Recently i was affected by problem 4, this PR solve it. I think this PR should be merged, it fix a lot of errors. what is happening with runbot building?

ChristophAbenthungCibex:
I ran into similar problems and this PR fixes them. LGTM

contract/models/abstract_contract_line.py
```diff
@@ -166,11 +172,9 @@ def _compute_recurring_interval(self):
     def _compute_date_start(self):
         self._set_recurrence_field("date_start")
 
-    # pylint: disable=missing-return
-    @api.depends("contract_id.recurring_next_date", "contract_id.line_recurrence")
-    def _compute_recurring_next_date(self):
```
pedrobaeza, line 171: Why are you removing this?

contract/models/contract_recurrency_mixin.py
```diff
@@ -48,6 +48,8 @@ class ContractRecurrencyBasicMixin(models.AbstractModel):
         help="Invoice every (Days/Week/Month/Year)",
     )
     date_start = fields.Date()
+    date_end = fields.Date()
+
```
pedrobaeza: Don't add empty lines here in the middle.

contract/models/contract_recurrency_mixin.py
```diff
@@ -89,7 +94,7 @@ class ContractRecurrencyMixin(models.AbstractModel):
         string="Next Period End",
         compute="_compute_next_period_date_end",
     )
-    last_date_invoiced = fields.Date(readonly=True, copy=False)
+    last_date_invoiced = fields.Date(copy=False)
```
pedrobaeza: Don't remove readonly, which is unrelated to the goal of this PR.

contract/views/contract.xml
```diff
@@ -249,11 +249,17 @@
...
                                     <field
                                         name="last_date_invoiced"
                                         groups="base.group_no_one"
                                     />
+                                    <field
+                                        name="next_period_date_start"
+                                        invisible="1"
+                                    />
+                                    <field name="next_period_date_end" invisible="1" />
+
```
pedrobaeza: Don't add empty lines in the middle.

contract/views/contract.xml
```diff
@@ -249,11 +249,17 @@
                                     />
                                     <field name="recurring_next_date" invisible="1" />
                                     <field name="date_start" invisible="1" />
-                                    <field name="date_end" />
+                                    <field name="date_end" invisible="1" />
```
pedrobaeza: Why making it invisible?

contract/models/abstract_contract_line.py
flalexg: Hi @pedrobaeza. Sorry for the delay for the response.

I removed this because it's redundant. Since we are inheriting from contract.recurrency.mixin.basic in the contract.line, it's redundant having the same on the abstract.contract.line.

contract/models/contract_recurrency_mixin.py
flalexg: Changed

contract/models/contract_recurrency_mixin.py
flalexg: My bad. Sorry.

contract/views/contract.xml
flalexg: Well, I make it invisible since the other fields like date_start or recurring_next_date are invisible. To be uniform, but I'm letting it visible.

contract/views/contract.xml
flalexg: Changed.

contract/models/abstract_contract_line.py
pedrobaeza: But this override was needed AFAIK.

pedrobaeza:
FYI we are planning to refactor whole module in https://github.com/OCA/contract/issues/1108

contract/models/abstract_contract_line.py
flalexg: Mmmm i'm going to check it now, but one of the problems is that the contract never ends. This happens because this inherit. 
On a first step, I created the _compute_date_end function below. This, trigger a chain of computes (next_period_date_start compute that triggers the recurring_next_date compute). 
Then, I checked this function and I concluded that it was redundant and not necessary. 

I've been working with this changed since I created the MR and it's working very well.

If u want, I can let function as it was initially since is not having any impact on the code.

HaraldPanten:
> FYI we are planning to refactor whole module in #1108

We're not far from creating a PR.

flalexg:
Hi @pedrobaeza @HaraldPanten!
I'm reading the messages right now.  So, as far as I read, we are planning to refactor the `subscription_oca`, but which are the plans for the `contract` module?

pedrobaeza:
I want to retake the refactoring soon, but can't say exactly when, as it depends when I finish other jobs.

HaraldPanten:
Hi @pedrobaeza @flalexg We'll make the refactoring of the contract module (V17). In a few weeks we will continue working on it.

flalexg:
OK! We can remove the stale tag at least from this PR @pedrobaeza?

ygol:
@pedrobaeza would this be merged or your refactoring will 'out date' it?

pedrobaeza:
This one still contains changes that are not acceptable,like removing computed methods and replace it by onchanges. I don't say the problems don't exist, but we should handle them without losing the modern approach with computed writable fields. That's why the possible refactoring (not yet completed) may help, simplifying current logic, which is indeed a bit ugly.

remi-filament:
Hi @flalexg thank you for your work on this PR.
You would need to rebase your work on OCA 16.0 branch.

I have added some tests for sections and notes for your PR : https://github.com/factorlibre/contract/pull/3
The tests show that having markers on section or note when not using recurrency at line level is not working properly (there is no dates on sections / notes so it gets a start_date = Today which is not necessarily the date being used on "normal" lines). I have also added a warning on contract form view and in ROADMAP section of readme to report this issue.

contract/models/abstract_contract.py
```diff
@@ -80,3 +80,20 @@ def _compute_journal_id(self):
...
+
+    @api.onchange(
+        "date_start",
+        "date_end",
+        "recurring_interval",
+        "recurring_rule_type",
+        "recurring_invoicing_type",
+    )
+    def _onchange_recurrency_fields(self):
+        if self.contract_line_fixed_ids:
```
rlizana, line 92: The `contract_line_fixed_ids` field is not declared in the `contract.abstract.contract` model, it is declared in `contract.contract`. This method shouldn't be here.

rlizana: Hi @flalexg I've added https://github.com/factorlibre/contract/pull/4 to fix it

rousseldenis:
> This one still contains changes that are not acceptable,like removing computed methods and replace it by onchanges. I don't say the problems don't exist, but we should handle them without losing the modern approach with computed writable fields. That's why the possible refactoring (not yet completed) may help, simplifying current logic, which is indeed a bit ugly.

I agree here with Pedro. Onchanges should be now avoided. Everything should be adressed through computed fields.

@flalexg

ygol:
what is the status? @pedrobaeza   is there still a plan to refactor? @HaraldPanten do you mean the same as Pedro or your work is on something else? I see that you closed https://github.com/OCA/contract/issues/1108 as completed. @rlizana addresses the onchange at https://github.com/factorlibre/contract/pull/4 is it acceptable ? Until the refactoring, is there something we can fix in the current PR that would be OK and provide a solution to the existing bugs? 
Thank you

pedrobaeza:
The refactoring was already made in 18.0: https://github.com/OCA/contract/tree/18.0/contract, and it's not going to be backported. You can propose though the proper changes (with the compute) here in 16, or directly go with the improvements in 18.

PR #1095, opened 2024-07-02
[16.0][IMP] contract: add company_id on contract lines

sbidoul:
This is useful to later add `check_company=True` on some contract line fields.


pedrobaeza (approved):
Multi-company rules should be also added to contract lines.

hparfr (approved):

PR #1100, opened 2024-07-11
[14.0][FIX] contract: product is also required in contract line tree view

HekkiMelody:
The Product is a mandatory field in the contract line form view

https://github.com/OCA/contract/blob/14.0/contract/views/abstract_contract_line.xml#L23

but it's not set as mandatory in the embedded tree view of the contract line. This commit fixes this inconsistency.

This causes problems in particular with module contract_sale_generation, since, unlike invoices, it's mandatory to have a product on a Sales Order line.

### DIFF
--- a/contract/views/contract_line.xml
+++ b/contract/views/contract_line.xml
@@ -97,7 +97,11 @@
                 decoration-info="create_invoice_visibility and not is_canceled"
             >
                 <field name="sequence" widget="handle" />
-                <field name="product_id" />
+                <field name="display_type" invisible="1" />
+                <field
+                    name="product_id"
+                    attrs="{'required': [('display_type', '=', False)]}"
+                />
                 <field name="name" />
                 <field
                     name="analytic_account_id"


rousseldenis:
@aleuffre I don't agree on this.

This should be defined as a constraint in sale module. You should be able to generate contract invoices for instance without product but just a description

pedrobaeza:
IIRC, right now is inconsistent, being already required in the popup form, so this is only bringing consistency. We can debate if instead, we change the current behavior to put as required=False also the product field in the form view (and subsequent required for sale generation).

rousseldenis:
The debate is old on this : https://github.com/OCA/contract/pull/667

HekkiMelody:
I'm happy to just put the required in `contract_sale_generation` instead if it'll get merged quicker/easier, since in that case it's for sure required. Will make a new PR asap.

Thank you both for your time.

pedrobaeza:
Seeing that old debate, I think this is correct, as requiring always the product is sane (for all the automatic price thing).

HekkiMelody:
Failing tests are unrelated, and due to subscription_oca

PR #1111, opened 2024-08-08
[13.0][FIX]: contract: Migration script check if column old_invoice_line_id exists

EmilioPascual:
@Shide @edlopen please review.

@moduon MT-6161

### DIFF
--- a/contract/migrations/13.0.1.0.0/post-migration.py
+++ b/contract/migrations/13.0.1.0.0/post-migration.py
@@ -6,24 +6,27 @@ from openupgradelib import openupgrade  # pylint: disable=W7936
 
 @openupgrade.migrate()
 def migrate(env, version):
-    openupgrade.logged_query(
-        env.cr,
-        """
-        UPDATE account_move am
-        SET old_contract_id = ai.old_contract_id
-        FROM account_invoice ai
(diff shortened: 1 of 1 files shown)


Shide (approved):

pedrobaeza (changes requested):
Starting with, please put the contrary condition and return early for minimizing the diff.

This seems to avoid crashes when coming from an enterprise migration, so masking all scripts is not the correct formula and you will have more troubles as intermediate DB schemas are not applied. 

Please don't pollute OCA with these hacks.

rafaelbn:
Dear @pedrobaeza ,

I appreciate your input on this PR. However, I would like to kindly remind you that the OCA's Code of Conduct encourages us to contribute positively to the community. Your response seems to be more focused on assumptions rather than on constructive feedback. The purpose of this PR is to correct an issue without causing any breakage, and it would be more productive to focus on whether it technically resolves the problem without introducing new issues.

Thank you for your understanding and collaboration.

Best regards,  
@rafaelbn 

Ping @OCA/project-service-maintainers

pedrobaeza:
Is my assumption incorrect?

rafaelbn:
Your assumption is an assumption. Is the PR incorrect? Breaks something?

PR #1113, opened 2024-08-08
[15.0][FIX] contract: delete record translation if exists xml_id

EmilioPascual:
@Shide @edlopen please review

@moduon MT-6161


Shide (approved):

pedrobaeza (changes requested):
This is not correct. The records are not removed and exist. The call is for removing the translations that are obsolete and they are not loaded through noupdate changes.

PR #1120, opened 2024-08-14
[17.0][IMP] product_contract: Add contract configurator instead of making tree not editable

carolinafernandez-tecnativa:
Before these changes, when trying to edit a line of sale order, it was opening the form of the line. But following the Odoo way of working with the events sales, we have make a new contract configurator that will be opened in a popup when selecting a contract product.

@Tecnativa TT50535

### DIFF
--- a/product_contract/models/product_template.py
+++ b/product_contract/models/product_template.py
@@ -61,6 +61,73 @@ class ProductTemplate(models.Model):
         string="Renewal type",
         help="Specify Interval for automatic renewal.",
     )
+    contract_start_date_method = fields.Selection(
+        [
+            ("manual", "Manual"),
+            ("start_this", "Start of current period"),
+            ("end_this", "End of current period"),
+            ("start_next", "Start of next period"),
+            ("end_next", "End of next period"),
+        ],
+        "Start Date Method",
+        default="manual",
+        help="""This field allows to define how the start date of the contract will
+        be calculated:
+
+        - Manual: The start date will be selected by the user, by default will be the
+        date of sale confirmation.
+        - Start of current period: The start date will be the first day of the actual
+        period selected on 'Invoicing Every' field. Example: If we are on 2024/08/27
+        and the period selected is 'Year(s)' the start date will be 2024/01/01.
+        - End of current period: The start date will be the last day of the actual
+        period selected on 'Invoicing Every' field. Example: If we are on 2024/08/27
+        and the period selected is 'Year(s)' the start date will be 2024/12/31.
+        - Start of next period: The start date will be the first day of the next
+        period selected on 'Invoicing Every' field. Example: If we are on 2024/08/27
+        and the period selected is 'Year(s)' the start date will be 2025/01/01.
+        - End of next period: The start date will be the last day of the actual
+        period selected on 'Invoicing Every' field. Example: If we are on 2024/08/27
+        and the period selected is 'Year(s)' the start date will be 2025/12/31.
+        """,
+    )
+    force_month_yearly = fields.Selection(
+        [
+            ("1", "January"),
+            ("2", "February"),
+            ("3", "March"),
+            ("4", "April"),
+            ("5", "May"),
+            ("6", "June"),
+            ("7", "July"),
+            ("8", "August"),
+            ("9", "September"),
+            ("10", "October"),
+            ("11", "November"),
+            ("12", "December"),
+        ],
+        "Force Month",
+    )
+    force_month_quarterly = fields.Selection(
+        [
+            ("1", "First month"),
+            ("2", "Second month"),
+            ("3", "Third month"),
+        ],
+        "Force Month",
+        help="Force the month to be used inside the quarter",
+    )
+    force_month_semesterly = fields.Selection(
+        [
+            ("1", "First month"),
+            ("2", "Second month"),
+            ("3", "Third month"),
+            ("4", "Fourth month"),
+            ("5", "Fifth month"),
+            ("6", "Sixth month"),
+        ],
+        "Force Month",
+        help="Force the month to be used inside the semester",
+    )
 
     def write(self, vals):
         if "is_contract" in vals and vals["is_contract"] is False:
--- a/product_contract/models/sale_order_line.py
+++ b/product_contract/models/sale_order_line.py
@@ -7,6 +7,13 @@ from dateutil.relativedelta import relativedelta
 from odoo import _, api, fields, models
 from odoo.exceptions import ValidationError
 
+MONTH_NB_MAPPING = {
+    "monthly": 1,
+    "quarterly": 3,
+    "semesterly": 6,
+    "yearly": 12,
+}
+
 
 class SaleOrderLine(models.Model):
     _inherit = "sale.order.line"
@@ -22,26 +29,9 @@ class SaleOrderLine(models.Model):
         string="Contract Template",
         compute="_compute_contract_template_id",
     )
-    recurring_rule_type = fields.Selection(
-        [
-            ("daily", "Day(s)"),
-            ("weekly", "Week(s)"),
-            ("monthly", "Month(s)"),
-            ("monthlylastday", "Month(s) last day"),
-            ("quarterly", "Quarter(s)"),
-            ("semesterly", "Semester(s)"),
-            ("yearly", "Year(s)"),
-        ],
-        default="monthly",
-        string="Invoice Every",
-        copy=False,
-    )
+    recurring_rule_type = fields.Selection(related="product_id.recurring_rule_type")
     recurring_invoicing_type = fields.Selection(
-        [("pre-paid", "Pre-paid"), ("post-paid", "Post-paid")],
-        default="pre-paid",
-        string="Invoicing type",
-        help="Specify if process date is 'from' or 'to' invoicing date",
-        copy=False,
+        related="product_id.recurring_invoicing_type"
     )
     date_start = fields.Date()
     date_end = fields.Date()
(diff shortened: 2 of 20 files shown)


pilarvargas-tecnativa (approved):
:+1:

carolinafernandez-tecnativa:
ping @pedrobaeza

product_contract/views/sale_order.xml
```diff
@@ -95,17 +95,64 @@
...
-                position="after"
+                expr="//field[@name='order_line']/tree//field[@name='price_unit']"
+                position="before"
             >
+                <field name="contract_template_id" column_invisible="1" />
+                <field name="is_contract" column_invisible="1" />
+                <field
+                    name="contract_id"
+                    options='{"no_create": True}'
+                    column_invisible="parent.is_contract == False"
```
pedrobaeza: Shorter, but also more correct, as `is_contract` may be other value than `False`, but also a falsy value.

```suggestion
                    column_invisible="not parent.is_contract"
```

Do the same on the rest.

product_contract/views/sale_order.xml
```diff
@@ -95,17 +95,64 @@
...
+                />
+                <field
+                    name="recurring_invoicing_type"
+                    column_invisible="parent.is_contract == False"
+                />
                 <field
                     name="date_start"
                     column_invisible="parent.is_contract == False"
+                    required="is_contract == True"
```
pedrobaeza:
```suggestion
                    required="is_contract"
```

and the same for the rest of the occurrences.

pedrobaeza (changes requested):
I'm not finding the corresponding fields in the sales order line tree:

image


Anyway, seeing how Odoo has adapted other similar needs, where we need extra info for a sales order line, showing a popup with that extra details, I think it should be the way to go:

Peek 2024-08-16 13-30

It seems not very complicated to be added, taking `event_sale` as example:

https://github.com/odoo/odoo/tree/17.0/addons/event_sale/static/src/js

@CarlosRoca13 can you please take this and perform it, as you already knows JS and the event_sale controller?

@sbejaoui do you agree?

Another question: why not doing a `detailed_type='contract'` instead of having a special field `is_contract` at product level?

CarlosRoca13:
@pedrobaeza @carolinafernandez-tecnativa pushed changes to add the product contract configurator like odoo does for events

carolinafernandez-tecnativa:
@pedrobaeza LGTM, what do you think?

pedrobaeza:
The PR title is incorrect, and I think the first commit is now not needed if it has to be undo in the second one.

pedrobaeza (commented):
About the functionality, this sounds better, although the "Contract" and "Contract line recurrency" should be at the end in a separate section, as this a very special case. I'm also surprised that the "End date" is filled by default, where it shouldn't.

And what about adding some demo master data to ease the tests?

product_contract/models/product_template.py
```diff
@@ -61,6 +61,71 @@ class ProductTemplate(models.Model):
...
+            ("11", "November"),
+            ("12", "December"),
+        ],
+        "Force Month",
+    )
+    force_month_quarterly = fields.Selection(
+        [
+            ("1", "1"),
+            ("2", "2"),
+            ("3", "3"),
```
pedrobaeza:
```suggestion
            ("1", "First month"),
            ("2", "Second month"),
            ("3", "Third month"),
```

product_contract/models/product_template.py
```diff
@@ -61,6 +61,71 @@ class ProductTemplate(models.Model):
...
+        ],
+        "Force Month",
```
pedrobaeza, line 116:
```suggestion
        "Force Month",
        help="Force the month to be used inside the quarter",
```

product_contract/models/product_template.py
```diff
@@ -61,6 +61,71 @@ class ProductTemplate(models.Model):
...
+    )
+    force_month_semesterly = fields.Selection(
+        [
+            ("1", "1"),
+            ("2", "2"),
+            ("3", "3"),
+            ("4", "4"),
+            ("5", "5"),
+            ("6", "6"),
```
pedrobaeza: The same here about the labels and the help.

product_contract/models/sale_order_line.py
```diff
@@ -269,3 +283,53 @@ def _compute_qty_to_invoice(self):
...
+                    + relativedelta(months=1 + increment)
+                    - relativedelta(days=1)
+                )
+            if "_next" in line.contract_start_date_method and start_date <= today:
+                start_date = start_date + relativedelta(months=month_nb)
+                if is_end:
+                    start_date = (
+                        start_date
+                        + relativedelta(day=1, months=1)
+                        - relativedelta(days=1)
```
pedrobaeza: Does this work?

```suggestion
                        + relativedelta(day=1, months=1, days=-1)
```

product_contract/models/sale_order_line.py
```diff
@@ -269,3 +283,53 @@ def _compute_qty_to_invoice(self):
...
+                    forced_month = month_nb * period_number + forced_value
+            # If forced_month is set, use it, but if it isn't use the month_period
+            start_date = today + relativedelta(
+                day=1, month=forced_month or month_period
+            )
+            if is_end:
+                increment = month_nb if not forced_month else 0
+                start_date = (
+                    start_date
+                    + relativedelta(months=1 + increment)
```
pedrobaeza: Isn't this missing?

```suggestion
                    + relativedelta(months=1 + increment, day=1)
```

product_contract/readme/ROADMAP.md
```diff
@@ -0,0 +1,2 @@
+Is not added support for dates computed from confirmation date of sale order for
+recurrent types daily, weekly or monthlylastday
```
pedrobaeza:
```suggestion
- There's no support right now for computing the start date for the
  following recurrent types: daily, weekly and monthlylastday.
```

product_contract/models/sale_order_line.py
```diff
@@ -269,3 +283,53 @@ def _compute_qty_to_invoice(self):
...
+            if (
+                line.contract_start_date_method == "manual"
+                or line.recurring_rule_type in ["daily", "weekly", "monthlylastday"]
+            ):
+                continue
+            is_end = "end_" in line.contract_start_date_method
+            today = fields.Date.today()
+            month_period = month = today.month
+            month_nb = MONTH_NB_MAPPING[line.recurring_rule_type]
+            # The period number is started by 0 to be able calculate the month
```
pedrobaeza:
```suggestion
            # The period number is started by 0 to be able to calculate the month
```

product_contract/wizards/product_contract_configurator.py
```diff
@@ -0,0 +1,127 @@
...
+    partner_id = fields.Many2one("res.partner")
+    company_id = fields.Many2one("res.company")
+    product_uom_qty = fields.Float("Quantity")
+    contract_id = fields.Many2one(comodel_name="contract.contract", string="Contract")
+    contract_template_id = fields.Many2one(
+        comodel_name="contract.template",
+        string="Contract Template",
+        compute="_compute_contract_template_id",
+    )
+    recurring_rule_type = fields.Selection(
```
pedrobaeza: Now that we have configured all the start date thing in the product, it doesn't seem OK to also have again this configuration here, don't you think?

product_contract/models/sale_order_line.py
CarlosRoca13: Already established in the above definition https://github.com/OCA/contract/pull/1120/files#diff-308d8794eb63f4f473a8ce812eba0aacdbdf86a8c0c8f9c3786f1de1baade116R317-R319

product_contract/wizards/product_contract_configurator.py
CarlosRoca13: I think we should keep it in case the product is set to manual mode, but the unnecessary fields should be hidden in the view.

CarlosRoca13:
Changes done. Let me add some tests for this new feature

product_contract/wizards/product_contract_configurator.py
pedrobaeza: @sbejaoui which was the reason to define a recurrency in the product and ask for that recurrency again in the sales order line? Can we strip that part for simplifying code and UI?

CarlosRoca13:
Tests added

pedrobaeza (commented):
Seems to work OK, but I find something weird, as if you add any line in the contract template, when confirming the quotation, both the sales order line and the template lines are added in the resulting contract. Was it that way in the original 12.0 implementation?

Something that may be related to this PR is that the extras lines don't have the proper start date.

pedrobaeza:
Ah, another thing: I think that the extra description about the recurrency and so on shouldn't be transferred to the contract line. Maybe you can add some markers at the beginning and the end to easily recognize it and remove it when creating the contract line.

CarlosRoca13:
> Was it that way in the original 12.0 implementation?

Yes, when the template has a line that line is added to the contract.

> Something that may be related to this PR is that the extras lines don't have the proper start date.

What date should be set to that lines?

> I think that the extra description about the recurrency and so on shouldn't be transferred to the contract line. Maybe you can add some markers at the beginning and the end to easily recognize it and remove it when creating the contract line.

Ok, I'll let you know when it's ready.

CarlosRoca13:
Change done

pedrobaeza:
> What date should be set to that lines?

The same that was computed for the sales order line.

pedrobaeza (approved):
/ocabot merge major

PR #1124, opened 2024-09-05
[FIX][16.0][contract] Fix issue on last_date_invoice contract line

MTantin:
## Module
contract

## Describe the bug
On multiline reccuring contract, the contract line last_date_invoice field is correct on the first line and is shift by 1 period time on the others lines (verified in database). This issue generate visible bug only when #START keywork is used in line note.

## To Reproduce
1. Generate a reccuring contract with a product and a line note using #START and #END keywords
2. Generate invoices for this contract
3. The first invoice will be OK but the next invoices will be shift in line note (for example "from 2024-10-01 to 2024-09-30" on september invoice)

**Expected behavior**
Have correct date in #START and #END keywords of line notes

## Correction
The bug came from `_update_recurring_next_date` method. This method is using `next_period_date_end` field to fill `last_date_invoiced` field but `last_date_invoiced` is used to recompute `recurring_next_date` of contract and impact `next_period_date_end`. So the first line is correctly edited but the next line use the new `recurring_next_date` and it create a shift for the next lines.

I keep the older mechanism of `_update_recurring_next_date` without `invoicing_date` and i've add a new mechanism with a specified `invoicing_date` to use the actual invoicing date.

### DIFF
--- a/contract/models/contract_line.py
+++ b/contract/models/contract_line.py
@@ -609,7 +609,7 @@ class ContractLine(models.Model):
         for rec in self:
             last_date_invoiced = rec.next_period_date_end
             if invoicing_date:
-                last_date_invoiced = self.get_next_period_date_end(
+                last_date_invoiced = rec.get_next_period_date_end(
                     invoicing_date,
                     rec.recurring_rule_type,
                     rec.recurring_interval,
--- a/contract/models/contract_line.py
+++ b/contract/models/contract_line.py
@@ -603,11 +603,18 @@ class ContractLine(models.Model):
         name = name.replace("#END#", last_date_invoiced.strftime(date_format))
         return name
 
-    def _update_recurring_next_date(self):
+    def _update_recurring_next_date(self, invoicing_date=False):
         # FIXME: Change method name according to real updated field
         # e.g.: _update_last_date_invoiced()
         for rec in self:
             last_date_invoiced = rec.next_period_date_end
+            if invoicing_date:
+                last_date_invoiced = self.get_next_period_date_end(
+                    invoicing_date,
+                    rec.recurring_rule_type,
+                    rec.recurring_interval,
+                    rec.date_end,
+                )
             rec.write(
                 {
                     "last_date_invoiced": last_date_invoiced,
--- a/contract/models/contract.py
+++ b/contract/models/contract.py
@@ -575,7 +575,7 @@ class ContractContract(models.Model):
                     )
             invoices_values.append(invoice_vals)
             # Force the recomputation of journal items
-            contract_lines._update_recurring_next_date()
+            contract_lines._update_recurring_next_date(date_ref)
         return invoices_values
 
     def recurring_create_invoice(self):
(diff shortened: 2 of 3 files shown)


contract/models/contract_line.py
```diff
@@ -603,11 +603,18 @@ def _insert_markers(self, first_date_invoiced, last_date_invoiced):
...
         return name
 
-    def _update_recurring_next_date(self):
+    def _update_recurring_next_date(self, invoicing_date=False):
         # FIXME: Change method name according to real updated field
         # e.g.: _update_last_date_invoiced()
         for rec in self:
             last_date_invoiced = rec.next_period_date_end
+            if invoicing_date:
+                last_date_invoiced = self.get_next_period_date_end(
```
rousseldenis:
```suggestion
                last_date_invoiced = rec.get_next_period_date_end(
```
?

remi-filament:
Thanks @MTantin for taking care of this issue. One of my customer experiences the same having a section and then a product line.
However, as far as I can tell, the problem is not only on date_end but also on next invoice date.

Please find below the output of generating the 3 next invoices for a contract post-paid every year starting on the 12th of Feb 2023, both on runboat without your fix and with your fix :
Without your fix :
image
--> we miss one invoice in 2025

First invoice (in 2024) lines are correct : 
image

On second invoice (in 2026) end dates are correct but start dates are different on every line : 
image

On third invoice (in 2027), the 2 first lines are correct, the start_date on the 3rd is not : 
image


With your fix : 
image
--> Invoices are generated only every 2 years (missing 1 in 2025, one in 2027)

However, dates on lines are correct every time : 
image

So it is better on lines, but there is still an issue on both last_date_invoiced and recurring_next_date and therefore on next invoice to be generated.

MTantin:
@remi-filament Yeah, i noticed some bugs coming from my fix over the different tests of the addon.
I have something else more important to do in the month but i will try to investigate more and fix this in october.

remi-filament:
Hi @MTantin , we worked on trying to find a solution for the above problem.
It turns out in reccurence.mixin, we had chained compute and _compute_next_period_date_end() was called 3 times with different values.
Removing dependencies on that function from :
- date_end (that would trigger change of next_period_date_start through _compute_next_period_date_start() that would then trigger _compute_next_period_date_end() anyhow)
- recurring_next_date (that would only be triggered by next_period_date_start change that would also trigger _compute_next_period_date_end())
is solving issues on our side. We proposed the change on your branch directly here : https://github.com/MTantin/contract/pull/1

Let us know if you have time to review it, or if you prefer that we start another PR on OCA including your changes ?

@pedrobaeza you may want to have a look at that one since you introduced the recurrency.mixin in v13 migration (with https://github.com/OCA/contract/commit/cd086ddbb4a9b85ccfcac288e53fb93e716d9446) ?

MTantin:
Hi @remi-filament , i've merged your commit without tests on my side, i've no time for that right now.
Are you sure that modifying the dependency won't prevent the field from being modified in cases that would otherwise have been valid?
I let you validate with OCA reviewers!

remi-filament:
Thanks @MTantin could you please rebase your work to get latest v16 commits before ?
Also, you may need to fix tests if still failing after rebase.

remi-filament:
Hi @MTantin I still have issues with the code from this PR (whether before or after the fix from https://github.com/MTantin/contract/pull/1).
I have to report that using code from https://github.com/OCA/contract/pull/1080 fixes these issues though.

PR #1131, opened 2024-09-23
[13.0][IMP] contract : Update res_partner_view.xml

Martronic-SA:
No need to write on customer_rank and supplier_rank here. This creates a conflict with partner_manual_rank module.

### DIFF
--- a/contract/views/res_partner_view.xml
+++ b/contract/views/res_partner_view.xml
@@ -8,7 +8,7 @@
         <field name="groups_id" eval="[(4, ref('account.group_account_invoice'))]" />
         <field type="xml" name="arch">
             <xpath expr="//div[@name='button_box']" position="inside">
-                <field name="customer_rank" invisible="1" readonly="1"/>
+                <field name="customer_rank" invisible="1" readonly="1" />
                 <button
                     name="act_show_contract"
                     type="object"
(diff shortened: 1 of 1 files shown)


pedrobaeza:
If they are invisible, there can't be any write on them. Or am I missing something?

rousseldenis:
@Martronic-SA Could you improve your commit message like `[IMP] contract: (...)` and fix pre-commit ? Thanks

rousseldenis:
> If they are invisible, there can't be any write on them. Or am I missing something?

Mmmmh, not sure they are not part of write() values.

pedrobaeza:
No as long as they are not modified, which is the case being invisible.

Martronic-SA:
Maybe beacause they are part of partner_manual_rank fields dependence?
The problem appears only when both modules are installed.
You set "Is Customer" on new contact, save, then modify the contact (modify the name), save, and the "Is Customer" check has gone. I fixed this by setting readonly on customer_rank and supplier_rank in contract form view.

pedrobaeza:
That sounds very weird to me... but it seems something to be fixed on `partner_manual_rank` instead of here. And please don't ignore the comments about the commit message format and don't add several commits.

https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message

Martronic-SA:
Ok it's happening only when adding a contact to the main partner.
Somehow the contact sets back it's customer_rank = 0 to the parent and as long as customer_rank is added as "writable" field in the form by contract module, it then writes the new customer_rank on save.
So this PR is not really relevant here.

Martronic-SA:
I close the PR. Needs further investigations.

PR #1142, opened 2024-11-08
[16.0][TEST] Add unit test to demonstrate incorrect customer assignment when generating contract invoices

rlizana:
This Pull Request adds a unit test in the contract module that highlights an error in the invoice generation process using the cron_recurring_create_invoice cron job.

**Error Description**
The issue occurs when multiple invoices are generated for contracts with different customers. Although the cron successfully creates the required invoices, the customer assigned to all the invoices is incorrect. Instead of assigning the correct customer to each invoice, the same customer is for all invoices.

**Unit Test**
The unit test configures multiple contracts with different customers and runs the cron job to generate recurring invoices. It checks that each invoice has the correct customer assigned, but the test fails, demonstrating the existing error.

### DIFF
--- a/contract/tests/test_contract.py
+++ b/contract/tests/test_contract.py
@@ -1744,6 +1744,26 @@ class TestContract(TestContractBase):
             len(invoice_lines),
         )
 
+    def test_cron_recurring_create_invoice_several_partners(self):
+        self.acct_line.date_start = "2018-01-01"
+        self.acct_line.recurring_invoicing_type = "post-paid"
+        self.acct_line.date_end = "2018-03-15"
+        other_partner = self.env.ref('base.res_partner_1')
+        contracts = self.contract2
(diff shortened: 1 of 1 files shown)


rlizana:
The correct field to use for invoice customer assignment is invoice_partner_id, not partner_id. Despite the confusion, I’m leaving the unit test here in case it’s useful for extending the test coverage or as a reference for future use.

rafaelbn:
@yajo please, could you review? Thank you! Ping @fcvalgar

rousseldenis:
> **Unit Test**
> The unit test configures multiple contracts with different customers and runs the cron job to generate recurring invoices. It checks that each invoice has the correct customer assigned, but the test fails, demonstrating the existing error.

@rlizana Tests are not failing... :sweat_smile:

PR #1149, opened 2024-12-03
[16][FIX] contract : ignore date constraint for a note

florian-dacosta:
In some cases there is a constraint  when adding a note because of the `recurring_next_date`.
Ignore the constraint for note, as it is already the cases for section

@bguillot


marcelsavegnago (approved):

pedrobaeza (approved):
Please fw-port it to upper versions

/ocabot merge patch

PR #1151, opened 2024-12-03
[16.0] [ADD] recurring_payment_stripe module

mjavint:
Add recurring payments with stripe


recurring_payments_stripe/models/account_move.py
```diff
@@ -0,0 +1,94 @@
...
+                        customer=token.provider_ref,
+                        payment_method=token.stripe_payment_method,
+                        # Para pagos automáticos sin intervención del usuario
+                        off_session=True,
+                        # Confirmar el PaymentIntent inmediatamente
+                        confirm=True,
+                        metadata={"odoo_invoice_id": str(invoice.id)},
+                    )
+
+                    # Manejar el resultado del PaymentIntent
```
janikvonrotz: Comments should be in English.

recurring_payments_stripe/__manifest__.py
```diff
@@ -0,0 +1,18 @@
+{
+    "name": "Recurring Payments with Stripe",
+    "version": "16.0.1.0.0",
+    "summary": """ Recurring Payments with Stripe """,
```
janikvonrotz: Not sure about the spacing before and after the `"""`.

janikvonrotz (commented):
Is there a Stripe API to run tests?

It would be nice to see a `tests/test_sale_subscription.py` and I guess this would require some kind of mocking of the actual Stripe API.

pedrobaeza:
Module name shouldn't contain plurals. Coupling this also to subscription module doesn't seem modular. A base module with the API, and another for linking with subscriptions is a better design.

jorgeglez1990 (commented):
I have reviewed the addon, and it appears to be functioning well. However, I’ve noticed that while the payment for the first invoice related to the subscription is being correctly recorded, the corresponding entries are not being created in the payment_transaction table for subsequent invoices. While the payment for the first invoice shows up as expected, the payment transactions for the following invoices are not being registered in payment_transaction.

recurring_payments_stripe/models/account_move.py
mjavint: Fixed translation problem in the comment

recurring_payments_stripe/__manifest__.py
mjavint: Fixed spaces before and after within summary content

rousseldenis:
> Module name shouldn't contain plurals. Coupling this also to subscription module doesn't seem modular. A base module with the API, and another for linking with subscriptions is a better design.

Agree

mjavint:
This PR is now ready for review.

Issue #1143, opened 2024-11-13
[16][IMP] Contract: Smart Button "Contracts" not displayed on Contact until an invoice is published


fcvalgar:
Module
contract

### Describe the bug

The "Contracts" smart button on the contact form view does not appear when a contract is created and in progress. It only becomes visible after an associated invoice is published.

### To Reproduce
Affected versions: Odoo 16

- Create a contract for a contact.
- Set the contract to "in progress" status.
- Go to the contact's form view to check for the "Contracts" smart button.

### Expected behavior:

- [ ] The "Contracts" smart button should be displayed on the contact form as soon as there is an active contract, regardless of whether there is a published invoice.

Additional context This issue limits the ability to view ongoing contracts directly from the contact form unless an invoice has been published.

MT-7885 @moduon

rafaelbn:
Hello @OCA/project-service-maintainers !

Do you agree in this improvement? 😄 

Let me know please! ❤️ 

Thank you!

rousseldenis:
@rafaelbn This is due to :

![image](https://github.com/user-attachments/assets/6f2a2ab0-b311-4d2c-8303-6dc7d0ae90c8)


So the `customer_rank`should be updated when we create a contract IMHO (that's logical). But IMHO too, we should not rely on customer rank to display the smart button.

rafaelbn:
Thank you for answering @rousseldenis !

> But IMHO too, we should not rely on customer rank to display the smart button.

fcvalgar:
To avoid relying on the customer_rank field for the visibility of the smart button, we propose modifying its logic to directly use the contract model.

We can implement a method to check for active contracts associated with the customer. 

> @rafaelbn This is due to :
> 
[...]

PR #1152, opened 2024-12-04
[FIX] [16.0] contract: Show contract button on partner when they have contracts too

Shide:
Show button on partner when they have contracts

fixes https://github.com/OCA/contract/issues/1143

MT-7885 @moduon @rafaelbn @Gelojr @fcvalgar @rousseldenis please review if you want :)

### DIFF
--- a/contract/views/res_partner_view.xml
+++ b/contract/views/res_partner_view.xml
@@ -19,7 +19,7 @@
                     groups="account.group_account_invoice"
                     icon="fa-book"
                     context="{'default_contract_type': 'sale', 'contract_type': 'sale'}"
-                    attrs="{'invisible': [('customer_rank','=',0)]}"
+                    attrs="{'invisible': [('customer_rank','=',0), ('sale_contract_count','=',0)]}"
                     help="Show the sale contracts for this partner"
                 >
                     <field
@@ -40,7 +40,7 @@
                     groups="account.group_account_invoice"
                     icon="fa-book"
                     context="{'default_contract_type': 'purchase', 'contract_type': 'purchase'}"
-                    attrs="{'invisible': [('supplier_rank','=',0)]}"
+                    attrs="{'invisible': [('supplier_rank','=',0), ('purchase_contract_count','=',0)]}"
                     help="Show the purchase contracts for this partner"
                 >
                     <field


contract/views/res_partner_view.xml
```diff
@@ -19,7 +19,7 @@
                     groups="account.group_account_invoice"
                     icon="fa-book"
                     context="{'default_contract_type': 'sale', 'contract_type': 'sale'}"
-                    attrs="{'invisible': [('customer_rank','=',0)]}"
+                    attrs="{'invisible': [('sale_contract_count','=',0)]}"
```
carlosdauden: With this change the contracts button will not appear when creating a new customer.

When creating a customer from, for example, Invoicing > Customers the customer created is already assigned a customer_rank value, so I don't understand the problem with the current behaviour.

https://github.com/odoo/odoo/blob/62af1c7b04557e94d18d6078544dc867e61ef2d2/addons/account/views/partner_view.xml#L280

carlosdauden (changes requested):

contract/views/res_partner_view.xml
fcvalgar: In our use case, we use the Members module, and contacts are created from this module. Thank you for the comment!

fcvalgar (approved):
Functional review.

LGTM, Thank you @Shide.

image

image

contract/views/res_partner_view.xml
rousseldenis: I would say, this is semantically more correct as the context is contract oriented. So, having contracts to show the button is better than 'being a customer'.

BUT, I'm wondering if showing it even in no contracts is better or not as with that feature, user can create a new contract for the partner directly. What do you think of ?

carlosdauden: Let's see if whith this we can find a consensus solution:
```suggestion
                    attrs="{'invisible': [('customer_rank','=',0), ('sale_contract_count','=',0)]}"
```

contract/views/res_partner_view.xml
```diff
@@ -40,7 +40,7 @@
                     groups="account.group_account_invoice"
                     icon="fa-book"
                     context="{'default_contract_type': 'purchase', 'contract_type': 'purchase'}"
-                    attrs="{'invisible': [('supplier_rank','=',0)]}"
+                    attrs="{'invisible': [('purchase_contract_count','=',0)]}"
```
carlosdauden:
```suggestion
                    attrs="{'invisible': [('supplier_rank','=',0), ('purchase_contract_count','=',0)]}"
```

carlosdauden (changes requested):

contract/views/res_partner_view.xml
rousseldenis: Ouch, I would say this is worse...

Shide:
@carlosdauden sure, your reasoning makes sense. 
Applied both conditions to invisible attr so the default values can be applied directly on the first contract.

contract/views/res_partner_view.xml
carlosdauden: A fairly common flow is to create the customer from the Invoicing > Customers option and once created click the smartbutton to create a new contract.
Any changes that involve losing the most basic flow to solve data integrity problems or "incorrect" processes by the user should not be included in the "base" modules.

carlosdauden (approved):

contract/views/res_partner_view.xml
rousseldenis: > A fairly common flow is to create the customer from the Invoicing > Customers option and once created click the smartbutton to create a new contract.

customer_rank == 0 if you create it from `Contacts` menu. So, this is not allowed in that case. Removing at all the domain is fair enough IMHO.

rafaelbn (approved):
/ocabot merge minor

PR #1153, opened 2024-12-04
[15.0][IMP] contract: Add INVOICEMONTHNAME marker to line description

victoralmau:
FWP from 14.0: https://github.com/OCA/contract/pull/946 + https://github.com/OCA/contract/pull/966

@Tecnativa


pedrobaeza (approved):
Can you please add in the next version fw-port the Spanish translations (under a new commit with `[I18N]`) to get in pair with the former version.

Fast-tracking this one meanwhile

/ocabot merge minor

PR #1154, opened 2024-12-04
[16.0][IMP] contract: Add INVOICEMONTHNAME marker to line description

victoralmau:
FWP from 15.0: https://github.com/OCA/contract/pull/1153

@Tecnativa


pedrobaeza (approved):
Thanks!

Fast-tracking for finishing the chain:

/ocabot merge minor

PR #1155, opened 2024-12-04
[16.0][IMP] contract: make invoicing offset modifiable

sbidoul:
This is not perfect as apparently the line must
be saved for this to take effect, as the on-screen value is not refreshed immediately.


carlosdauden:
ping @sergio-teruel

Issue #1160, opened 2024-12-16
[15.0] Error translate marker month name


Bilbonet:
Hello @victoralmau 
Since this change #1153 
I have this error generating recurring invoices.

>   File "../contract/models/contract_line.py", line 613, in _translate_marker_month_name
>     return months[month_name]
> Exception
> 
> The above exception was the direct cause of the following exception:
> 
> Traceback (most recent call last):
>   File "../server/odoo/http.py", line 658, in _handle_exception
>     return super(JsonRequest, self)._handle_exception(exception)
>   File "../server/odoo/http.py", line 301, in _handle_exception
>     raise exception.with_traceback(None) from new_cause
> KeyError: 'diciembre'

Thanks

victoralmau:
Hello,

I have tried to reproduce the error you indicate but I have not been able to: http://oca-contract-15-0-2c48c9335eb1.runboat.odoo-community.org/web#id=2&cids=1&menu_id=108&action=268&model=contract.contract&view_type=form + http://oca-contract-15-0-2c48c9335eb1.runboat.odoo-community.org/web#id=3&cids=1&menu_id=108&action=268&model=contract.contract&view_type=form 
Can you indicate the steps to reproduce it?

Thanks.

Bilbonet:
Hello

I think the problem is here:

https://github.com/Tecnativa/contract/blob/85ae222dd3593ebb43f61b89826c196fe4495de8/contract/models/contract_line.py#L626

If your locale is in other laguage the name of the month doesn't much. You need to use locale for ensuring that the name of the month is always correct.

```
import locale
locale.setlocale(locale.LC_ALL, 'en_US.UTF-8')

first_date_invoiced.strftime("%B") = "december"
```

victoralmau:
This can be closed, it is already solved https://github.com/OCA/contract/pull/1163

rousseldenis:
@Bilbonet Could you check ?

PR #1163, opened 2024-12-20
[15.0][IMP] contract: Use %m instead of %B to obtain month

victoralmau:
Use `%m` instead of `%B` to obtain month

Related to https://github.com/OCA/contract/pull/1153 Fixes https://github.com/OCA/contract/issues/1160

Use the month number to set the text of the month and avoid problems according to the language defined in locale.

@Tecnativa

### DIFF
--- a/contract/models/contract_line.py
+++ b/contract/models/contract_line.py
@@ -597,18 +597,18 @@ class ContractLine(models.Model):
 
     def _translate_marker_month_name(self, month_name):
         months = {
-            "January": _("January"),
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):
Yes, this approach is more resilient and doesn't have any side effect. As it doesn't mean any change at user level, let's fast-track:

/ocabot merge patch

Bilbonet:
Really cool 🫶
Thanks @victoralmau

PR #1164, opened 2024-12-20
[15.0][FIX] product_contract: Fix tests (do not import a test file that does not exist)

victoralmau:
Fix tests (do not import a test file that does not exist)

@Tecnativa


pedrobaeza (approved):
Straight-forward:

/ocabot merge patch

PR #1179, opened 2025-01-10
[16.0] [IMP] _compute_auto_renew on quotation sent

victor-champonnois:
This allows to make the product contract settings work on ecommerce orders. The product auto-renew settings are only applied on product change. This doesn't happen for sale orders created from the ecommerce. This PR force the computation of auto-renew settings on checkout.

### DIFF
--- a/product_contract/models/sale_order.py
+++ b/product_contract/models/sale_order.py
@@ -110,6 +110,11 @@ class SaleOrder(models.Model):
                 line.create_contract_line(line.contract_id)
         return contract_model.browse(contracts)
 
+    def action_quotation_sent(self):
+        for line in self.order_line:
+            line._compute_auto_renew()
+        return super().action_quotation_sent()
+
     def action_confirm(self):
         """If we have a contract in the order, set it up"""
         self.filtered(


huguesdk (changes requested):
instead of calling `._compute_auto_renew()` explicitly, i think it would be better to fix the `@api.depends()` call of that method by listing all the fields that are used, like this:

```python
    @api.depends(
        "product_id.is_contract",
        "product_id.default_qty",
        "product_id.recurring_rule_type",
        "product_id.recurring_invoicing_type",
        "product_id.is_auto_renew",
        "product_id.auto_renew_interval",
        "product_id.auto_renew_rule_type",
    )
    def _compute_auto_renew(self):
```

there is a slight problem though: the compute method calls `fields.Date.today()` if `date_start` is not set. this shouldn’t be used in a compute method, but rather as a default field value.

moreover, `date_end` shouldn’t be computed in this method, but in another method depending on `date_start`, `recurring_rule_type` and `product_uom_qty` (like `.onchange_date_start()`, which might maybe be transformed into a compute method).

victor-champonnois:
@huguesdk 
actually just fixing the api depends by adding the ones you suggest does not suffice to trigger the method call. I don't know why though, it's just what I experience on my tests.

huguesdk:
if i understand correctly, when a sale order is created from the e-commerce, `._compute_auto_renew()` is not called? i think this is because there are only 3 fields that use this compute method (although it computes other ones) and they are all 3 writable and have a default value. this means that there is indeed nothing to compute at creation time. for it to work, in addition with the changes from my previous comment, these fields should be redefined as (writable and storable) computed fields using `._compute_auto_renew` as compute method since they are all set in this method:

*   `product_uom_qty`
*   `recurring_rule_type`
*   `recurring_invoicing_type`
*   `date_start`
*   `date_end`

anyway, i think that there is a design problem in this model. ok to leave it like this for now.

victor-champonnois:
@huguesdk thank you for your analysis, I think you are right. OK to leave it like that in this PR

PR #1180, opened 2025-01-13
[ADD] contract_sale_generation_payment_mode

victor-champonnois:
Set the payment mode from the contract on sales orders created by the contract.


huguesdk (approved):
lgtm! (amended some slight fixes to the commit.)

huguesdk:
@OCA/human-resources-maintainers could someone please review or merge this?

rousseldenis:
@victor-champonnois Thanks for this. Maybe a little test ? Thanks

victor-champonnois:
@rousseldenis thanks, I've added some tests

victor-champonnois:
I don't manage to see why the codecov stays red, an idea on how to fix this ? 
It states that these lines are not covered, but I don't see a way to cover them more than what I've already did. Can someone help ? :)

```
+         res = super()._prepare_recurring_sales_values(date_ref)
+         if res:
+             res[0]["payment_mode_id"] = self.payment_mode_id.id
+         return res
```

huguesdk:
according to the logs and the coverage report, the test files are loaded, but the tests don’t run. those of `contract_sale_generation` don’t run either. this is due to a recent change in the tests of `contract_sale_generation` (in #1201). i fixed it in #1305 (and tagged the tests as `post_install` here also). i think the coverage will be fine as soon as #1305 is merged.

victor-champonnois:
many thanks @huguesdk !

huguesdk:
@OCA/project-service-maintainers could someone please merge this?

contract_sale_generation_payment_mode/models/contract.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+class ContractContract(models.Model):
+    _inherit = "contract.contract"
+
+    def _prepare_recurring_sales_values(self, date_ref=False):
+        res = super()._prepare_recurring_sales_values(date_ref)
+        if res:
+            res[0]["payment_mode_id"] = self.payment_mode_id.id
+        return res
```
mamcode: I think it's fine, but it seems more consistent with the module's design to put this code in the `_prepare_sale()` method. What do you think?

huguesdk: @mamcode good catch, thanks! indeed. it is fixed now, and this fixes a bug where the payment mode would only be set on the first sale order if the method was called on a record set with multiple records.

rousseldenis:
@huguesdk Could you make module name appear in commit message ?

image

huguesdk:
@rousseldenis done (in the body of the message, as [the subject must be maximum 50 characters long](https://chris.beams.io/git-commit#limit-50) and the module name is too long for this).

rousseldenis:
> @rousseldenis done (in the body of the message, as [the subject must be maximum 50 characters long](https://chris.beams.io/git-commit#limit-50) and the module name is too long for this).

Nope, you should do it in subject. Thanks

huguesdk:
@rousseldenis done (although i think it is less readable).

please note the [oca guidelines about commit messages](https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message) state:

> Write a short commit summary without prefixing it. It should not be longer than 50 characters: This is a commit message
> 
> Then, in the message itself, specify the part of the code impacted by your changes (module name, lib, transversal object, ...) and a description of the changes.

but afterwards, they state:

> *   Use tags as [listed in the Odoo Guidelines](https://www.odoo.com/documentation/16.0/contributing/development/git_guidelines.html#tag-and-module-name) with the following extensions:
>     *    **[MIG]** for migrating a module

and then gives examples like:

> ```
> [FIX] website: remove unused alert div
> 
> Fix look of input-group-btn
> (…)
> ```

the odoo guidelines state:

> Tags are used to prefix your commit. They should be one of the following
> 
> *   [FIX] for bug fixes: (…)
> *   (…)
> 
> After tag comes the modified module name. Use the technical name as functional name may change with time.

so the examples seem to follow the odoo guidelines, but they contradict what is written in the oca guidelines. maybe those should be updated?

rousseldenis:
The aim is to make git history more readable at first look and help technical review more easy.

PR #1185, opened 2025-01-21
[17.0][IMP] product_contract: Allow creating sales orders with lines that lack a contract template

carlos-lopez-tecnativa:
Before this commit, confirming a sales order would raise an exception if any line lacked a configured contract template, preventing the creation of the contract.

Now, sales order lines without a contract template are automatically added to a new contract without a template, allowing the process to continue seamlessly.

TT54544
@Tecnativa @pedrobaeza @CarlosRoca13 @sergio-teruel  could you please, review this

### DIFF
--- a/product_contract/models/sale_order.py
+++ b/product_contract/models/sale_order.py
@@ -48,8 +48,11 @@ class SaleOrder(models.Model):
 
     def _prepare_contract_value(self, contract_template):
         self.ensure_one()
+        name = (
+            f"{contract_template.name}: {self.name}" if contract_template else self.name
+        )
         return {
-            "name": f"{contract_template.name}: {self.name}",
+            "name": name,
             "partner_id": self.partner_id.id,
             "company_id": self.company_id.id,
             "contract_template_id": contract_template.id,
@@ -63,6 +66,7 @@ class SaleOrder(models.Model):
     def action_create_contract(self):
         contract_model = self.env["contract.contract"]
         contracts = []
+        lines_without_contract = self.env["sale.order.line"].browse()
         for rec in self.filtered("is_contract"):
             line_to_create_contract = rec.order_line.filtered(
                 lambda r: not r.contract_id and r.product_id.is_contract
@@ -80,17 +84,8 @@ class SaleOrder(models.Model):
                     rec.company_id
                 ).property_contract_template_id
                 if not contract_template:
-                    raise ValidationError(
-                        _(
-                            "You must specify a contract "
-                            "template for '%(product_name)s' product "
-                            "in '%(company_name)s' company."
-                        )
-                        % {
-                            "product_name": order_line.product_id.name,
-                            "company_name": rec.company_id.name,
-                        }
-                    )
+                    lines_without_contract |= order_line
+                    continue
                 contract_templates |= contract_template
             for contract_template in contract_templates:
                 order_lines = line_to_create_contract.filtered(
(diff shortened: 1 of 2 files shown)


product_contract/models/sale_order.py
```diff
@@ -63,6 +66,7 @@ def _prepare_contract_value(self, contract_template):
     def action_create_contract(self):
         contract_model = self.env["contract.contract"]
         contracts = []
+        line_without_contract = self.env["sale.order.line"]
```
pedrobaeza: In this case, the plural is justified. It's just on module names where we avoid it.

```suggestion
        lines_without_contract = self.env["sale.order.line"]
```

product_contract/models/sale_order.py
```diff
@@ -109,6 +104,15 @@ def action_create_contract(self):
                 order_lines.write({"contract_id": contract.id})
             for line in line_to_update_contract:
                 line.create_contract_line(line.contract_id)
+            if line_without_contract:
```
pedrobaeza: Shouldn't this be a `for` loop for more than one sales order lines?

product_contract/models/sale_order.py
carlos-lopez-tecnativa: Done

product_contract/models/sale_order.py
carlos-lopez-tecnativa: It is not necessary in this case, as the `create_contract_line` method already performs the loop. Please see https://github.com/OCA/contract/blob/17.0/product_contract/models/sale_order_line.py#L191-L195

product_contract/models/sale_order.py
```diff
@@ -109,6 +104,15 @@ def action_create_contract(self):
                 order_lines.write({"contract_id": contract.id})
             for line in line_to_update_contract:
                 line.create_contract_line(line.contract_id)
+            if lines_without_contract:
+                contract = contract_model.create(
+                    rec._prepare_contract_value(self.env["contract.template"])
+                )
+                contracts.append(contract.id)
+                contract._onchange_contract_template_id()
```
pedrobaeza: I don't think this onchange should be executed.

carlos-lopez-tecnativa: You are right, thanks

pedrobaeza (approved):

pedrobaeza:
@sbejaoui do you agree?

product_contract/models/sale_order.py
```diff
@@ -63,6 +66,7 @@ def _prepare_contract_value(self, contract_template):
     def action_create_contract(self):
         contract_model = self.env["contract.contract"]
         contracts = []
+        lines_without_contract = self.env["sale.order.line"]
```
rousseldenis:
```suggestion
        lines_without_contract = self.env["sale.order.line"].browse()
```
is more correct syntaxically

carlos-lopez-tecnativa: done

product_contract/models/sale_order.py
```diff
@@ -109,6 +104,14 @@ def action_create_contract(self):
                 order_lines.write({"contract_id": contract.id})
             for line in line_to_update_contract:
                 line.create_contract_line(line.contract_id)
+            if lines_without_contract:
+                contract = contract_model.create(
+                    rec._prepare_contract_value(self.env["contract.template"])
+                )
+                contracts.append(contract.id)
+                contract._onchange_contract_type()
```
rousseldenis, line 112: Something to do on contract side to put a TODO to change `automatic_price` on lines to computed.

pedrobaeza: Isn't the same? When you put `env[model]`, a new recordset is instantiated, so no possible pool sharing between successive calls to this.

rousseldenis:
> Now, sales order lines without a contract template are automatically added to a new contract without a template, allowing the process to continue seamlessly.

Isn't it a bit weird as some wanted values are not populated ?

pedrobaeza:
Which values? We have only identified the template name.

carlos-lopez-tecnativa:
> > Now, sales order lines without a contract template are automatically added to a new contract without a template, allowing the process to continue seamlessly.
> 
> Isn't it a bit weird as some wanted values are not populated ?

@rousseldenis  Please provide me with more details; it is not clear.

pedrobaeza:
Merging after some time open for not dealing with conflicts:

/ocabot merge minor

PR #1187, opened 2025-01-24
[17.0][IMP] product_contract: Set description under product name instead of description

CarlosRoca13:
Using the feature extraLines of Many2one field Component we can add the info under the product and avoid noice on the product description.

You can show this info on the description by using the system parameters:
    - product_contract.show_recurrency
    - product_contract.show_invoicing_type
    - product_contract.show_date

cc @Tecnativa TT54693

ping @pedrobaeza @carlos-lopez-tecnativa @sergio-teruel

### DIFF
--- a/product_contract/models/sale_order_line.py
+++ b/product_contract/models/sale_order_line.py
@@ -6,6 +6,7 @@ from dateutil.relativedelta import relativedelta
 
 from odoo import _, api, fields, models
 from odoo.exceptions import ValidationError
+from odoo.tools import str2bool
 
 MONTH_NB_MAPPING = {
     "monthly": 1,
@@ -73,6 +74,9 @@ class SaleOrderLine(models.Model):
     )
     contract_start_date_method = fields.Selection(
         related="product_id.contract_start_date_method"
+    )
+    product_contract_description = fields.Text(
+        compute="_compute_product_contract_description"
     )
 
     @api.constrains("contract_id")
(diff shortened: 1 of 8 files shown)


product_contract/readme/CONFIGURE.md
```diff
@@ -0,0 +1,5 @@
+You can show the details showed under the product on description by using the next system parameters.
```
pedrobaeza:
```suggestion
You can include the contract details on the sales order description by
using the following system parameters:
```

product_contract/models/sale_order_line.py
```diff
@@ -307,64 +310,95 @@ def _set_contract_line_start_date(self):
...
-                        force_month_label = field_selection.get(
-                            line.product_id["force_month_%s" % line.recurring_rule_type]
-                        )
-                        date_text += " (%s)" % force_month_label
-                field_info = dict(
-                    self._fields["recurring_rule_type"].get_description(self.env)
-                )
-                field_selection = dict(field_info.get("selection"))
-                recurring_rule_label = field_selection.get(line.recurring_rule_type)
+    def get_product_contract_date_text(self):
```
pedrobaeza: Put these methods private.

pedrobaeza (commented):
imagen

CarlosRoca13:
@pedrobaeza changes done!

product_contract/models/sale_order_line.py
```diff
@@ -307,64 +310,103 @@ def _set_contract_line_start_date(self):
...
+    @api.depends(
+        "date_start", "date_end", "recurring_rule_type", "recurring_invoicing_type"
+    )
+    def _compute_name(self):
+        res = super()._compute_name()
+        ICP = self.env["ir.config_parameter"].sudo()
+        for line in self:
+            if line.is_contract:
+                description = ""
+                if ICP.get_param("product_contract.show_recurrency") and (
```
pedrobaeza: Please use `str2bool` tool method

CarlosRoca13: Done

sergio-teruel (approved):

pedrobaeza (approved):
/ocabot merge minor

product_contract/models/sale_order_line.py
```diff
@@ -307,64 +311,103 @@ def _set_contract_line_start_date(self):
...
-    - Invoicing Type: {invoicing_type}
-    - Date: {date_text}
-                """
-                ).format(
-                    product=line.product_id.display_name,
-                    recurring_rule=recurring_rule_label,
-                    invoicing_type=invoicing_type_label,
-                    date_text=date_text,
+                force_month_label = field_selection.get(
+                    self.product_id["force_month_%s" % self.recurring_rule_type]
```
carlos-lopez-tecnativa, line 339: I think it is necessary to take into account whether the field is set or not before accessing it, or make it a required field in the product form view to prevent mistakes.
I set the field as empty intentionally, but the user can do the same. What do you think?
image

image

carlos-lopez-tecnativa (commented):
@CarlosRoca13 sorry for the delay in revision. I found an error. It could be that the issue was present before this change, but I think it’s good to fix it. Let me know what you think.

product_contract/models/sale_order_line.py
CarlosRoca13: I think this could be attended in this other PR: https://github.com/OCA/contract/pull/1188

PR #1189, opened 2025-02-01
[16.0][FIX] contract: Invoice creation message translatable

pedrobaeza:
As it was, the text to translate is not correctly extracted, so it's not translated, creating a mix of languages, as `_creation_message` is returning a translated sentence.

We take also the opportunity to change the way the sentence is built for having as translatable only the "by contract" words, to avoid possible errors in translations due to the placeholders.

Before:
imagen

After:
imagen

@Tecnativa


pedrobaeza:
Fast-tracking as obvious bug:

/ocabot merge patch

PR #1191, opened 2025-02-04
[17.0][FIX] product_contract: Compute date_start from the current date instead of date_order

carlos-lopez-tecnativa:
This commit standardizes `date_start` across methods, ensuring it is set to the current date instead of date_order. If a user adds a new line after confirming the order (several days or months later), `date_start` would otherwise be outdated.

TT54892

@Tecnativa @pedrobaeza @sergio-teruel @CarlosRoca13 Could you please review this

### DIFF
--- a/product_contract/models/sale_order_line.py
+++ b/product_contract/models/sale_order_line.py
@@ -108,12 +108,13 @@ class SaleOrderLine(models.Model):
 
     @api.depends("product_id")
     def _compute_date_start(self):
+        today = fields.Date.context_today(self)
         for sol in self:
             if sol.contract_start_date_method == "start_this":
-                sol.date_start = sol.order_id.date_order.replace(day=1)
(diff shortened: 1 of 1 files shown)


pedrobaeza (commented):
I didn't remember all the compute thing, but my idea was not to have this field pre-computed, but calculated when confirming the sales order. Why it's pre-filled if you don't know when the order is going to be confirmed? @sergio-teruel you did the change. What was your goal with it?

carlos-lopez-tecnativa:
ping @sergio-teruel @CarlosRoca13

sergio-teruel (approved):

pedrobaeza (approved):
I'm not totally convinced about this being stored and not being computed on confirmation, but at least this fixes the new lines added after confirmation, so let's merge it.

/ocabot merge patch

PR #1192, opened 2025-02-08
ADD sale-generation

marionumza:


pedrobaeza (changes requested):
This is not `ADD`, but `[MIG]` and have to follow https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0

PR #1193, opened 2025-02-12
[17.0][FIX] product_contract: Make configuration tab compatible with sale_product_configurator

CarlosRoca13:
Before:
1. Create a contract product
2. Add variants and select Product Configurator
3. Add the product on a Sale Order
4. Try to edit the variants with the configurator

You will not be able to open the product configurator

After:
The product configurator will be opened first and then the contract configurator.

cc @Tecnativa TT55066

ping @pedrobaeza @carlos-lopez-tecnativa @sergio-teruel

### DIFF
--- a/product_contract/README.rst
+++ b/product_contract/README.rst
@@ -65,8 +65,8 @@ To use this module, you need to:
 Known issues / Roadmap
 ======================
 
-- There's no support right now for computing the start date for the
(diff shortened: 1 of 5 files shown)


pedrobaeza (approved):
Tested on runboat with the test product "Contract", and working.

carlos-lopez-tecnativa (approved):
Tested, LGTM

PR #1194, opened 2025-02-12
[17.0][IMP] product_contract: Set new button to configure contract

CarlosRoca13:
With this change, we allow the user to modify the contract without going through the product configuration stage, which overwrites the product and causes the loss of some information, even when no changes are made.

Removed too the description feature because is changing the description when no option is activated. Since we have the information about what has been configured under the product, this is not a completely necessary feature. Added BTW on the ROADMAP.

cc @Tecnativa TT55066

ping @pedrobaeza @carlos-lopez-tecnativa

### DIFF
--- a/product_contract/README.rst
+++ b/product_contract/README.rst
@@ -42,16 +42,6 @@ invoice directly.
 .. contents::
    :local:
 
-Configuration
-=============
-
-You can include the contract details on the sales order description by
-using the following system parameters:
-
-1. **Recurrency** -> product_contract.show_recurrency
-2. **Invoicing Type** -> product_contract.show_invoicing_type
-3. **Date** -> product_contract.show_date
-
 Usage
 =====
 
@@ -67,6 +57,10 @@ Known issues / Roadmap
 
 -  There's no support right now for computing the start date for the
    following recurrent types: daily, weekly and monthlylastday.
+-  It would be interesting to be able to configure what to display in
+   the line description using a configuration option, but using
+   \_compute_name is not valid because, if nothing is configured, it
+   would reset what we have in the description.
 
 Bug Tracker
 ===========
--- a/product_contract/README.rst
+++ b/product_contract/README.rst
@@ -65,8 +65,8 @@ To use this module, you need to:
 Known issues / Roadmap
 ======================
 
-- There's no support right now for computing the start date for the
-  following recurrent types: daily, weekly and monthlylastday.
+-  There's no support right now for computing the start date for the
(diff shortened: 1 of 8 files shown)


pedrobaeza:
Put the removal in a separate commit.

CarlosRoca13:
Done @pedrobaeza

pedrobaeza (approved):
Shouldn't you remove also the context storage thing?

carlos-lopez-tecnativa (changes requested):
Please check this error:

- Add a product with variants and a contract.
- Edit the product, confirm the variant, but cancel the contract configuration.
- The product is removed—why?
- Obviously, when trying to edit again, the error occurs because the product is empty.

Peek 2025-02-12 13-20

CarlosRoca13:
@pedrobaeza The storage of the context can be removed, but when changing a product attribute, the wizard's information will not be automatically filled in.

@carlos-lopez-tecnativa The issue that is occurring is that when pressing OK in the product configurator, the product is assigned regardless of whether it is changed or not. As a result, the new product process is executed. If, in a new product, the contract configuration is canceled, this causes the product not to be assigned, just as it happens with events.

carlos-lopez-tecnativa:
> @carlos-lopez-tecnativa The issue that is occurring is that when pressing OK in the product configurator, the product is assigned regardless of whether it is changed or not. As a result, the new product process is executed. If, in a new product, the contract configuration is canceled, this causes the product not to be assigned, just as it happens with events.

I understand your point, but when the record has not been saved yet, it's fine for the product to be empty.

Would it be possible to hide the buttons if no product is selected? This would prevent users from clicking and causing a traceback.

However, if the sale.order has already been saved and has an ID, canceling the wizard should not clear the product. Also, I think that on the second attempt, the `Edit Configuration` button should not open the contract configuration. There is now a dedicated button for that, so this should only happen when adding a new line. This change could also resolve the issue.

This is an edge case that usually doesn’t occur. I'm just testing possible scenarios based on customer feedback.

What do you think?

CarlosRoca13:
I understand your point and I agree, but when the product is overwritten (by the variant configurator), the *onchange* events are triggered, which will cause the information on the screen to be cleared.  

By launching the wizard again, we ensure that the information is retained since we are saving the context before the overwriting occurs.  

I will see if I can write this information directly into the record when this happens, thus avoiding the need for that second step through the wizard.

CarlosRoca13:
I think that I achieve it :smile:

PR #1201, opened 2025-02-26
[16.0][FIX] contract: Post-install test + fallback to load CoA

luisDIXMIT:


contract/tests/test_contract.py
```diff
@@ -2334,7 +2345,7 @@ def test_recurrency_propagation(self):
         vals = {
             "recurring_rule_type": "yearly",
             "recurring_interval": 2,
-            "date_start": to_date("2020-01-01"),
+            "date_start": fields.Date.today(),
```
luisDIXMIT: Changed the date contract to today

pedrobaeza: Why? Having variable dates is the recipe for future failures in extreme cases like February with leap years and so.

rousseldenis:
@pedrobaeza I'm more concern about introducing those changes if not needed.

In fact, if I have understood the issue, this should fail only if company is set to another country than 'us' (as no template is loaded) ? Correct ?

pedrobaeza:
Not exactly, it's also reproducible if you don't have l10n_us installed although the country is set as "us". The difference is that previously a of there was any CoA, it was loaded for the main company, no matter the country or the installed localization packages. 

Now you are forced to install l10n_us in all your CIs, or l10n_generic_coa if inheriting from TestAccountInvoicingCommon for having a green CI.

This little code is mimicking the previous behavior.

contract/tests/test_contract.py
luisDIXMIT: Without the change, the tests are failing. However, in local, test are correct. I will investigate it. Thanks!

pedrobaeza:
The problem may be due to the pass of tests to post-install, where all the installed modules in the repository affects. In fact, this is good, as it may spot a potential compatibility problem, but it's true that it's an annoyance to hunt the root cause.

luisDIXMIT:
ping @etobella

rousseldenis:
> The problem may be due to the pass of tests to post-install, where all the installed modules in the repository affects. In fact, this is good, as it may spot a potential compatibility problem, but it's true that it's an annoyance to hunt the root cause.

My concern is about introducing code for 'possible behaviour' in repositories that don't pull particular localizations... That should be avoided.

contract/tests/test_contract.py
```diff
@@ -12,17 +12,28 @@
...
 class TestContractBase(common.TransactionCase):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
+        if not cls.env.company.chart_template_id:
+            # Load a CoA if there's none in current company
+            coa = cls.env.ref("l10n_generic_coa.configurable_chart_template", False)
+            if not coa:
+                # Load the first available CoA
+                coa = cls.env["account.chart.template"].search(
```
rousseldenis, line 32: Moreover, IMHO, this is not correct. If several coa's have been installed, how can you be sure the good one will be loaded for the current company ?

pedrobaeza: System just need an installed CoA for issuing an invoice. It doesn't matter which CoA is installed though.

rousseldenis: IMHO, this kind of code should be set in particular modules, not all.

Tests are not passing by the added code...

etobella:
@rousseldenis The problem of not adding this code is that we are disallowing Full Test execution of a system.

OCA is providing Unit Testing in all their modules and this is awesome. Unit tests allows us to test that everything works as expected. However, this tests are executed only at repository level, so we might have collisions between modules.

As Partners, I think that it is recommended to provide a full environment Tests to our customers in order to ensure that their system will work as expected and we will have no collisions between modules. With gitaggregate and pip we are able to have just the necessary modules and make a complete test of the environment easily. IMO this is good value to partners and customers and it gives more trust to the full solution we are providing.

Isn't it enough reason?

It is not just @pedrobaeza opinion. In my case, tests are failing in my customers due to this change. It is not related to spanish localization, as it affects also other country customers.

Also, in most cases, this piece of code will not affect the system, it will be executed only in some specific cases. This problem is similar to tests that rely on demo data. It is a good practice to get rid of it for the same reason.

sbidoul:
I don't fully understand the cause, but this looks like a lot of boilerplate code to add to all OCA tests.

Could it be Odoo will revert or change at some point? I suppose they will soon have customers with the same problem, since odoo.sh also runs all odoo+custom code tests.

rousseldenis:
> As Partners, I think that it is recommended to provide a full environment Tests to our customers in order to ensure that their system will work as expected and we will have no collisions between modules

That's why IMHO it is recommended to do integration tests and not to run at new OCA modules' tests in your environment. OCA modules tests are not intended to be run like that.

If I have customer's environment constraints, do I have to put my particular usage here ? I don't think so.

sbidoul:
> That's why IMHO it is recommended to do integration tests and not to run at new OCA modules' tests in your environment. OCA modules tests are not intended to be run like that.

That's a vast topic... I won't pronounce on that ;)

contract/tests/test_contract.py
victoralmau: Reviewing the current test error I see that it happens because `product_contract` is installed and https://github.com/OCA/contract/blob/d87cdf378e6c181a6fd76c3c2c65ca721c7de070/product_contract/models/contract_line.py#L36 changes the behavior.

I think the proposed solution is the right one, another proposal (although it is really “the same”) is to define in this test `@freeze_time(“2020-01-01”)`

pedrobaeza: @luisDIXMIT please try with the `freeze_time` one.

luisDIXMIT: @victoralmau @pedrobaeza  That works. Thanks a lot, Victor, for your help. This has been a struggle for me.

README.md
```diff
@@ -31,7 +31,7 @@ addon | version | maintainers | summary
...
 [contract_update_last_date_invoiced](contract_update_last_date_invoiced/) | 16.0.1.0.1 | [![rafaelbn](https://github.com/rafaelbn.png?size=30px)](https://github.com/rafaelbn) [![edlopen](https://github.com/edlopen.png?size=30px)](https://github.com/edlopen) | This module allows to update the last date invoiced if invoices are deleted.
-[contract_variable_qty_timesheet](contract_variable_qty_timesheet/) | 16.0.1.0.0 | [![carlosdauden](https://github.com/carlosdauden.png?size=30px)](https://github.com/carlosdauden) [![pedrobaeza](https://github.com/pedrobaeza.png?size=30px)](https://github.com/pedrobaeza) [![danypr92](https://github.com/danypr92.png?size=30px)](https://github.com/danypr92) | Add formula to invoice
+[contract_variable_qty_timesheet](contract_variable_qty_timesheet/) | 16.0.1.0.1 | [![carlosdauden](https://github.com/carlosdauden.png?size=30px)](https://github.com/carlosdauden) [![pedrobaeza](https://github.com/pedrobaeza.png?size=30px)](https://github.com/pedrobaeza) [![danypr92](https://github.com/danypr92.png?size=30px)](https://github.com/danypr92) | Add formula to invoice
```
pedrobaeza: This change shouldn't be here. Maybe an incorrect rebase?

etobella: True! Fixed :wink:

pedrobaeza (approved):
Although there can be some controversy about the path chosen for solving the existing issue, this patch in tests cannot break any other thing, and this way we avoid other things like touching all the CI systems, so let's apply it. Merging it counting with your tolerance for speeding up this and closing this phase.

/ocabot merge patch

PR #1204, opened 2025-02-26
[FIX] contract_variable_quantity: Post-install test + fallback to load CoA

luisDIXMIT:


luisDIXMIT:
This change has been moved to https://github.com/OCA/contract/pull/1201

PR #1136, opened 2024-11-04
[18.0][MIG][REF] split contract module into: contract, contract_line_successor, contract_termination

sbejaoui:
- Standard migration to 18.0
- Remove abstract.contract.mixin and abstract.contract.line.mixin, and replace them with direct inheritance from contract.template and contract.template.line
- Extract the contract line life cycle into a new module contract_line_successor
- Extract the contract termination feature into a new module contract_termination
- Fix contract portal pages
- Clean up the UI and improve UX

image

image

image

image

PR #1206, opened 2025-02-28
[18.0][MIG] contract_sale_invoicing: Migration to 18.0 from 17.0

tobiaszehntner:
Sale Order on 18.0 has no longer the field `analytic_account_id`. Instead, a variable distribution of multiple Analytic Accounts per Sale Order Line is possible. We adapted the module to consider only Sale Orders where *all lines match 100%* on the analytic account.


tobiaszehntner:
Depends on:

- [x] <https://github.com/OCA/contract/pull/1136>

acsonefho (approved):

victoralmau:
Can you delete the commit “_[DON'T MERGE] test-requirements.txt_” (the linked PRs are already merged) and rebase so we can review it in runboat?

tobiaszehntner:
> Can you delete the commit “_[DON'T MERGE] test-requirements.txt_” (the linked PRs are already merged) and rebase so we can review it in runboat?

done

contract_sale_invoicing/models/contract.py
```diff
@@ -0,0 +1,69 @@
...
+            lambda c: c.invoicing_sales and c.recurring_next_date
+        )
+        sale_domain = expression.OR(
+            [c._get_invoiceable_sales_domain() for c in affected_contracts]
+        )
+        all_sales = self.env["sale.order"].search(sale_domain)
+        for contract in affected_contracts:
+            possible_sales = all_sales.filtered_domain(
+                contract._get_invoiceable_sales_domain()
+            )
```
victoralmau:
```suggestion
        for contract in affected_contracts:
            possible_sales = self.env["sale.order"].search(contract._get_invoiceable_sales_domain())
```
I don't think `all_sales` adds much value if we then call the `_get_invoiceable_sales_domain()` method again.

contract_sale_invoicing/models/contract.py
```diff
@@ -0,0 +1,69 @@
...
+            sales = contract._get_matching_analytic_account_sales(possible_sales)
```
victoralmau:
```suggestion
```
IMO this shouldn't be necessary. There was nothing similar in v17 (https://github.com/OCA/contract/blob/d1709eddc7dfca0123e81f71433f0e94a6f2b698/contract_sale_invoicing/models/contract.py#L37), and perhaps you don't always want to distribute 100%. What do you think, @pedrobaeza ?

victoralmau (changes requested):

contract_sale_invoicing/models/contract.py
pedrobaeza: I need more context.

contract_sale_invoicing/models/contract.py
tobiaszehntner: it's to do only one `search` (for all contracts in `affectect_contracts`) for performance, inside the for-loop we then only apply the domain on the found records related to the specific contract.

contract_sale_invoicing/models/contract.py
tobiaszehntner: @pedrobaeza 
The module's intention described on line 52: "When creating invoices, add pending Sale Orders with same Analytic Account"

However, in v18, Sale Order has no longer an analytic account, and it's lines can have multiple each (analytic_distribution). To not invoice a Sale Order multiple times, I decided to only search for Sale Orders where lines are using only the specific account at 100%.

This function is separated out and can be overwritten to search differently for pending Sale Orders.

But apart from migrating I haven't used this module and have no use-cases, so I'm very much open to suggestions. 🙌

tobiaszehntner: @victoralmau it was basically to make this line from v17 work in v18:
```
("analytic_account_id", "=", contract.group_id.id),
```
Maybe there's a much simpler approach.

pedrobaeza: Thanks for the extra explanations, @tobiaszehntner. One strategy aligned with the change in v18 is to only invoice lines with the analytic account, which involves to invoice the sales order, and remove invoice lines which doesn't fit.

tobiaszehntner: @pedrobaeza how about:
```diff
    def _recurring_create_invoice(self, date_ref=False):
        invoices = super()._recurring_create_invoice(date_ref)
        for contract in self:
            if not contract.invoicing_sales or not contract.recurring_next_date:
                continue
            sales = self.env["sale.order"].search(
                [
-                   ("analytic_account_id", "=", contract.group_id.id),
+                   (
+                       "order_line.distribution_analytic_account_ids",
+                       "in",
+                        self.group_id.ids,
+                   ),
                    (
                        "partner_invoice_id",
                        "child_of",
                        contract.partner_id.commercial_partner_id.ids,
                    ),
                    ("invoice_status", "=", "to invoice"),
                    (
                        "date_order",
                        "<=",
                        f"{contract.recurring_next_date} 23:59:59",
                    ),
                ]
            )
            if sales:
-              invoices |= sales._create_invoices()
+              invoices |= sales.with_context(filter_on_analytic_account=contract.group_id.id)._create_invoices()
        return invoices
```
and in `sale_order.py`:
```diff

+ from odoo import models
+ from odoo.tools import float_compare
+
+ class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    def _get_invoiceable_lines(self, final=False):
+        lines = super()._get_invoiceable_lines(final)
+        matching_lines = lines
+        if analytic_account := self.env.context.get("filter_on_analytic_account"):
+            for line in lines:
+                for account_ids, percent in line.analytic_distribution:
+                    if analytic_account not in [int(acc_id) for acc_id in account_ids.split(",") or not float_compare(
+                        percent, 100, precision_digits=line.analytic_precision
+                    )]:
+                        matching_lines -= line
+        return matching_lines
```

pedrobaeza: Yeah, that can serve! You can remove the extra variable `matching_lines` if you want.

pedrobaeza (approved):
Code review.

victoralmau (approved):

PR #1216, opened 2025-04-10
[FIX] minor fixes in spanish translation

carlossainznetkia:
Minor fixes in Spanish translation for the contract module.

### DIFF
--- a/contract/i18n/es.po
+++ b/contract/i18n/es.po
@@ -5,6 +5,7 @@
 # Translators:
 # OCA Transbot <transbot@odoo-community.org>, 2017
 # Pedro M. Baeza <pedro.baeza@gmail.com>, 2017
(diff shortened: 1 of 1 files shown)


carlosdauden (approved):

pedrobaeza:
The translation should be done in the Weblate platform:

https://odoo-community.org/page/translate

I would accept this here, but it's missing several things:

- The commit message is not correct. https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message
- The style should be "usted", not "tú", so you need to change also that part in addition to the gender.

Thus, I'm going to close this, and please perform the changes directly in the platform.

carlossainznetkia:
Thanks for the clarification and the detailed feedback!

I’ll make the changes directly on the Weblate platform as suggested.
Apologies for the oversight, and I’ll make sure to follow the guidelines more carefully next time.

PR #1219, opened 2025-04-15
[17.0][FIX]Subscription icon

Gerloav:
New Icon for 17.0


rousseldenis:
I'm not fan of the design. Why not using instead https://spilymp.github.io/ibo/

Even if the tool is locked for version <= 15, IMHO it is better than this

PR #1221, opened 2025-04-17
[16.0][IMP] contract: Improve recurring invoice offset handling and resolve issue 1220 

### LINKED ISSUE #1220: [16.0] Multiple issues with recurring_invoicing_offset and #INVOICEMONTHNAME# marker in contract lines
## Module
`contract` (specifically affecting `contract.line` model)

## Describe the bug
This issue report consolidates several problems related to the `recurring_invoicing_offset` field (Invoice Offset) and the `#INVOICEMONTHNAME#` description marker within contract lines in Odoo 16.0:

1.  **`recurring_invoicing_offset` is not editable:** The field `contract.line.recurring_invoicing_offset` is currently defined without `store=True` and `readonly=False` (implicitly or explicitly read-only depending on the view/inheritance), making it impossible to directly modify the invoice offset on a contract line via the UI or standard means.
2.  **Negative offset validation is too strict:** The validation checks (`_check_last_date_invoiced` and `_check_recurring_next_date_start_date`) prevent setting a negative offset if the resulting `recurring_next_date` is before `last_date_invoiced` or `date_start`. However, this check triggers *even if no invoices have been generated yet* for the contract, which seems overly restrictive. The validation should likely only apply if there's an actual overlap risk with existing invoiced periods.
3.  **`recurring_next_date` does not update dynamically:** When the `recurring_invoicing_offset` value is changed (assuming it's made editable via fix), the `recurring_next_date` field does not update immediately in the form view. It only recalculates upon saving the contract record. This lacks responsiveness for the user.
4.  **`#INVOICEMONTHNAME#` marker issues:**
    *   **Potential `KeyError`:** The `_insert_markers` method calls `_translate_marker_month_name` passing the *translated* month name (based on partner language) but the invoked method attempts a dictionary lookup (`months[month_name]`) which expects the English month name as the key, potentially leading to a `KeyError`. Using `.get()` might be safer, but the core issue remains.
    *   **Conceptual limitation:** The `#INVOICEMONTHNAME#` marker is only meaningful for monthly recurring billing. It provides little to no value (and can be confusing) for contracts with quarterly, semi-annual, annual, or other billing frequencies. Its usage could potentially be deprecated or disabled as it was in previous versions.

## To Reproduce
**Affected versions**: Odoo 16.0 (running the OCA `contract` module version for 16.0)

**Steps to reproduce the behavior:**

**Issue 1 (Non-editable offset):**
1. Go to the Contracts application.
2. Open or create a Contract.
3. Add or select a recurring Contract Line.
4. Try to find and edit the "Invoice Offset" (`recurring_invoicing_offset`) field in the form view.
5. *Observed:* The field is typically not present or is read-only, preventing direct modification.

**Issue 2 (Strict negative offset check):**
1. Create a new Contract with a recurring Contract Line (e.g., Start Date today, Monthly recurrence).
2. **Ensure no invoices have been generated for this contract yet.**
3. Make the `recurring_invoicing_offset` field editable (e.g., via fix).
4. Set a negative "Invoice Offset" (e.g., -5) such that the calculated `recurring_next_date` would be earlier than the line's `date_start`.
5. Save the Contract.
6. *Observed:* A validation error like "You can't have a date of next invoice anterior to the start of the contract line..." or "You can't have the next invoice date before the date of last invoice..." is raised, even though no invoices exist to create a conflict.

**Issue 3 (No dynamic update):**
1. Open a Contract with a recurring Contract Line. Note the current "Next Invoice Date" (`recurring_next_date`).
2. Make the `recurring_invoicing_offset` field editable (e.g., via fix).
3. Change the value in the "Invoice Offset" field.
4. *Observed:* The "Next Invoice Date" field in the form view does *not* change immediately. It only updates after the entire Contract record is saved.

**Issue 4 (#INVOICEMONTHNAME# error/issue):**
1. Create/Edit a Contract Line and add the marker `#INVOICEMONTHNAME#` to its description (`name` field).
2. Ensure the Contract's Partner has a language set to something other than English (e.g., Italian, French, Spanish).
3. Trigger a process that calls `contract.line._insert_markers()`, such as generating an invoice for this line.
4. *Observed (Potential Error):* A `KeyError` might occur within `_translate_marker_month_name` due to the mismatch between the translated month name passed and the keys expected by the `months` dictionary.
5. *Observed (Conceptual Issue):* If the contract line's recurrence is not monthly (e.g., Quarterly), the meaning/usefulness of `#INVOICEMONTHNAME#` becomes unclear or irrelevant.

**Expected behavior**

1.  **Editable Offset:** The `recurring_invoicing_offset` field should be configurable (e.g., `store=True`, `readonly=False` added to the base model definition or managed via views) to allow users to set invoice offsets directly on contract lines.
2.  **Smarter Validation:** The validation checks in `_check_last_date_invoiced` and `_check_recurring_next_date_start_date` should only raise the date-related errors (`recurring_next_date` vs `last_date_invoiced`/`date_start`) if `self.contract_id._get_related_invoices()` returns a non-empty recordset (i.e., if at least one invoice has already been generated for the contract).
3.  **Dynamic Date Update:** The `recurring_next_date` should be recalculated and updated in the form view immediately (via an `onchange` method) whenever the `recurring_invoicing_offset` field (or other relevant fields like `recurring_rule_type`, `recurring_interval`, `date_start`, etc.) is modified.
4.  **Marker Handling:**
    *   At minimum, the potential `KeyError` in `_translate_marker_month_name` should be fixed (e.g., using `.get()` or ensuring the correct key is used).
    *   Ideally, considering its limited applicability, the handling/usage of the `#INVOICEMONTHNAME#` marker in `_insert_markers` should be reconsidered, potentially removing it entirely as proposed in the original analysis, reverting to behavior similar to older versions.

**Additional context**
*   Odoo Version: 16.0
*   OCA `contract` module
*   The proposed code snippets in the original analysis (provided by the user reporting this) offer potential solutions:
    *   Making the field editable.
    *   Adding `if self.contract_id._get_related_invoices()` condition to validations.
    *   Implementing an `@api.onchange("recurring_invoicing_offset")` to call `_compute_recurring_next_date`.
    *   Commenting out the `#INVOICEMONTHNAME#` replacement line in `_insert_markers`.

These issues impact the usability and robustness of managing invoice timing offsets and description formatting in recurring contracts.

mktsrl:
Fixes https://github.com/OCA/contract/issues/1220

### DIFF
--- a/contract/models/contract_line.py
+++ b/contract/models/contract_line.py
@@ -459,7 +459,10 @@ class ContractLine(models.Model):
             ):
                 continue
             if line.date_start and line.recurring_next_date:
-                if line.date_start > line.recurring_next_date:
+                if (
+                    line.date_start > line.recurring_next_date
+                    and line.contract_id._get_posted_related_invoices()
+                ):
                     raise ValidationError(
                         _(
                             "You can't have a date of next invoice anterior "
@@ -467,6 +470,13 @@ class ContractLine(models.Model):
                         )
                         % line.name
                     )
+
+    @api.onchange(
+        "recurring_invoicing_offset",
+    )
+    def onchange_recurring_invoicing_offset(self):
+        for rec in self:
+            rec._compute_recurring_next_date()
 
     @api.constrains(
         "date_start", "date_end", "last_date_invoiced", "recurring_next_date"
@@ -493,6 +503,7 @@ class ContractLine(models.Model):
                 )
             if (
                 rec.recurring_next_date
+                and rec.contract_id._get_posted_related_invoices()
                 and rec.recurring_next_date <= rec.last_date_invoiced
             ):
                 raise ValidationError(
@@ -621,12 +632,6 @@ class ContractLine(models.Model):
         name = self.name
         name = name.replace("#START#", first_date_invoiced.strftime(date_format))
         name = name.replace("#END#", last_date_invoiced.strftime(date_format))
-        name = name.replace(
-            "#INVOICEMONTHNAME#",
-            self.with_context(lang=lang.code)._translate_marker_month_name(
-                first_date_invoiced.strftime("%m")
-            ),
-        )
         return name
 
     def _update_recurring_next_date(self):
--- a/contract/tests/test_contract.py
+++ b/contract/tests/test_contract.py
@@ -291,6 +291,9 @@ class TestContract(TestContractBase):
         self.assertTrue(invoice_daily)
         self.assertEqual(self.acct_line.recurring_next_date, recurring_next_date)
         self.assertEqual(self.acct_line.last_date_invoiced, last_date_invoiced)
+        self.assertFalse(self.contract._get_posted_related_invoices())
+        invoice_daily[0].action_post()
+        self.assertTrue(self.contract._get_posted_related_invoices())
 
     def test_contract_invoice_followers(self):
         self.acct_line.recurring_next_date = "2018-02-23"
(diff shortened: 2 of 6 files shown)


TennyMkt (approved):

danall59 (approved):
Chnges tested and OK

rousseldenis:
@mktsrl Thanks for this.

Could you improve your commit message by making module appear like `[IMP] contract: <summary - a clear text>` and add what has been changed in lines under ? Many thanks

contract/models/contract_line.py
```diff
@@ -621,12 +632,12 @@ def _insert_markers(self, first_date_invoiced, last_date_invoiced):
         name = self.name
         name = name.replace("#START#", first_date_invoiced.strftime(date_format))
         name = name.replace("#END#", last_date_invoiced.strftime(date_format))
-        name = name.replace(
-            "#INVOICEMONTHNAME#",
-            self.with_context(lang=lang.code)._translate_marker_month_name(
-                first_date_invoiced.strftime("%m")
-            ),
-        )
+        # name = name.replace(
```
rousseldenis: Please don't keep commented lines of code.

contract/models/contract_line.py
```diff
@@ -468,6 +471,13 @@ def _check_recurring_next_date_start_date(self):
                         % line.name
                     )
 
+    @api.onchange(
+        "recurring_invoicing_offset",
+    )
+    def onchange_recurring_invoicing_offset(self):
```
rousseldenis, line 477: Why not putting field in compute method's depends ?

contract/tests/test_contract.py
```diff
@@ -517,10 +520,208 @@ def test_check_date_end(self):
             self.acct_line.date_end = "2015-12-31"
 
     def test_check_recurring_next_date_start_date(self):
-        with self.assertRaises(ValidationError):
-            self.acct_line.write(
```
rousseldenis, line 521: Is that normal that somenthing that should fail do not anymore?

contract/models/contract_line.py
mktsrl: Done. Commented code removed.

contract/models/contract_line.py
mktsrl: The `onchange `was added primarily to fix the UI issue (Point 3 in the original issue description) where the recurring_next_date wasn't updating immediately when the offset changed. It just calls the compute method to provide that instant feedback.
You're right that adding the offset to `depends `is generally best practice for ensuring the compute runs on save. We did explore this, but noticed it caused unexpected failures in several existing core tests related to date calculations.
To keep this PR focused on fixing the reported offset issues without needing to refactor unrelated tests, we decided the `onchange` sufficiently addressed the user-facing problem for now.

contract/tests/test_contract.py
mktsrl: Yes, the change in test_check_recurring_next_date_start_date is indeed expected.
This test previously failed because the validation was overly strict, blocking valid negative offsets even on new contracts (as noted in the original issue, point 2). The fix relaxes this by checking for posted invoices before raising the error.
Since this specific test runs without posted invoices, it correctly passes now with the new logic. The scenario where the error should occur (with posted invoices) is covered by the 2 new test_constraint_..._with_posted_invoice tests added in this PR.

PR #1242, opened 2025-05-22
[17.0][FIX]contract_variable_qty: Fixes bad attribute inherit, setting the …

javierjcf:
…value to a boolean expresion instead of string.

Since the invisibility condition of the quantity field was set as a string, it was always evaluated as True, which caused the tests of other modules to fail.

See #1234 to underestand the tests problem.


pedrobaeza (approved):
Yes!

/ocabot merge patch

PR #1243, opened 2025-05-22
[17.0][FIX]contract_variable_quantity:Fix all occurrences of the invi…

javierjcf:
…sible modifier. The expression must be boolean.

Is the same work as #1242 . Need to fix tests in #1234


pedrobaeza (approved):
/ocabot merge patch

PR #1254, opened 2025-06-16
[18.0][FIX] contract line successor: use parent.is_auto_renew to avoid EvalError in column_invisible

Anxo82:
@HaraldPanten 
[T-8579]
When creating a new contract template and selecting "supplier" as the contract_type, show an EvalError _"Can not evaluate python expression: (bool(parent.contract_type == 'purchase'  and not is_auto_renew))"_

image

The fix replaces `is_auto_renew` with `parent.is_auto_renew` to ensure the expression is evaluated correctly in the context of the parent form. To make this work, the is_auto_renew field was added as a computed Boolean field on the contract.template model.

### DIFF
--- a/contract/static/src/js/contract_portal_tour.esm.js
+++ b/contract/static/src/js/contract_portal_tour.esm.js
@@ -1,5 +1,5 @@
-import {registry} from "@web/core/registry";
 import {redirect} from "@web/core/utils/urls";
+import {registry} from "@web/core/registry";
 
 registry.category("web_tour.tours").add("contract_portal_tour", {
     test: true,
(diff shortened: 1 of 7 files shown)


Jaimermaccione (approved):
@Anxo82 I’ve tested the described case and the error no longer occurs. I’d say this works fine now from a functional point of view. LGTM!

Tisho99 (approved):
LGTM Technical review

HaraldPanten:
@pedrobaeza Could we merge this FIX?

THX! 🙏

pedrobaeza (approved):
/ocabot merge patch

PR #1256, opened 2025-06-19
[IMP] contract: Allow to single invoice a contract by users 17.0

barkat-matthias:


pedrobaeza:
There's already the possibility in current code without this PR to generate an invoice manually from a contract.

barkat-matthias:
> There's already the possibility in current code without this PR to generate an invoice manually from a contract.

Yes, but the wizard also allow to bill multiple period

pedrobaeza:
I don't totally understand you. Anyway, apart from the button, there's already a manual invoicing wizard in the current module, where you put the date to be taken into account. If what you want to do is to limit to a selection of contracts, expand the current wizard for that usage.

StephaneMangin (approved):
LGTM

PR #1257, opened 2025-06-21
[FIX] contract: fix price_unit computation with different UoM

xabispacebiker:
Before this fix, the computed 'price_unit' on contract lines did not reflect the correct value when selecting a unit of measure different from the product's reference UoM.

This fix ensures the price is converted using `_compute_price()` to reflect the selected UoM, similar to how Sales and Invoices work.

Steps to reproduce:
- Create a product with UoM "License" (e.g. yearly, 12€)
- Add a secondary UoM "Month" with a 1/12 factor
- Add this product to a contract line using the "Month" UoM
- Before: price_unit = 12 (wrong)
- After: price_unit = 1 (correct)

This brings contract pricing logic in line with standard Odoo behavior.

### DIFF
--- a/contract/models/abstract_contract_line.py
+++ b/contract/models/abstract_contract_line.py
@@ -9,6 +9,7 @@
 from odoo import api, fields, models
 from odoo.exceptions import ValidationError
 from odoo.tools.translate import _
+from odoo.tools.float_utils import float_round
 
 
 class ContractAbstractContractLine(models.AbstractModel):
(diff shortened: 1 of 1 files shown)


contract/models/abstract_contract_line.py
```diff
@@ -206,18 +208,20 @@ def _compute_price_unit(self):
...
                     partner=line.contract_id.partner_id.id,
                     date=line.env.context.get(
                         "old_date", fields.Date.context_today(line)
                     ),
                 )
-                line.price_unit = pricelist._get_product_price(product, quantity=1)
+                base_price = pricelist._get_product_price(product, quantity=quantity)
+                price = product.uom_id._compute_price(base_price, line.uom_id)
+                line.price_unit = float_round(
+                    price, precision_digits=line.currency_id.decimal_places or 2
```
pedrobaeza, line 223: The price unit precision is not the currency one. In fact, you don't have to round anything, as the ORM does it through the `digits` parameter when storing.

PR #1259, opened 2025-06-26
Make contract LGPL

baimont:


sbejaoui (approved):

pedrobaeza (changes requested):
I don't agree with this change as contributor of the module. Let's not prostitute the OCA modules by enterprise/closed source interests.

rousseldenis (approved):
> Let's not prostitute the OCA modules by enterprise/closed source interests.

I personally think that even we should privilege the full open side of contributions, excellent modules like this could be a base to extend our customers usages. That's not prostitution, but allow flexibility for those who want to use even enterprise modules.

I agree with the move.

@OCA/project-service-maintainers

pedrobaeza:
I don't think that way, as if you continue switching to LGPL everything, and not forcing contributors to open their contributions on top of these, there will be another big private ecosystem apart from Odoo and OCA, which is not desirable, and I don't want anyone to benefit that way from my contributions. If I contribute something, is for others to play the same game. I don't want to be one of the only fools ones contributing here, and you promote that with this license change.

I don't approve this move in any of our modules (except website ones, for protection to my customers having published webs - but this is not the case -), and this one has big part of us in it.

ValentinVinagre (changes requested):
I do not agree with the license change.

rvalyi:
Hello, I'm not a module contributor, but I think major contributors like @pedrobaeza should have their copyright respected so I voice against such a change.

My personal stance is that it's okay to LGPL a few small trivial framework modules when authors agree with it (because otherwise they are simply copied). But I strongly believe complex key modules like contract are better sticking to the free software philosophy of AGPL, otherwise soon enough there will be huge proprietary offers competing with the OCA and there will always be somebody with more money than you to win this financial game against you. Better close the door to the sharks now.

JordiBForgeFlow:
My preference is also to license as LGPL small, more technical features.

lmignon:
@pedrobaeza @rvalyi @JordiBForgeFlow @ValentinVinagre  , we fully understand and support your point of view. The challenge we're facing is to find a way to make the module we like and to which we contribute cohabit with the Odoo helpdesk module used by our customer. It is absolutely not our intention to allow third-party players to benefit from all the work that the authors and contributors have invested in this module to develop their paid-for solution. I'll be talking to my colleagues to find the right approach while respecting the current license.

Regards,

lmi

@baimont Can you close this PR and the others one on the same subject plz?

pedrobaeza:
It's a pity that a company good contributor like you is affected by these problems, provoked by the enterprise outbreak.

I had it clear when LGPL license change + enterprise rise happened: I didn't go to put any enterprise module in my customer instances, even if some of them pay enterprise licenses. Why? Because it makes you as implementator very dependent on Odoo SA, lock your customers to pay for life for enterprise licenses, and it breaks open source spirit: you can't share the improvements you make over that modules with the community, nor maintain them in the same way.

Why don't you switch to the OCA helpdesk one? It's very powerful, and from Tecnativa we have improved a lot its features. Look for example the number of PRs we have made over the repository (114 to the date - well, counting fw-ports -):

https://github.com/OCA/helpdesk/pulls?q=is%3Apr+involves%3ATecnativa+is%3Aclosed

We can help you if any feature is missing, although the data transformation from enterprise model to community would be on your side, but I think it's not too much and OpenUpgrade has bigger challenges.

PR #1262, opened 2025-06-26
Make product_contract LGPL

baimont:
depends on https://github.com/OCA/contract/pull/1261


sbejaoui (approved):

pedrobaeza (changes requested):
I don't agree with this change as contributor of the module. Let's not prostitute the OCA modules by enterprise/closed source interests.

PR #1264, opened 2025-06-26
Make contract_variable_quantity LGPL

baimont:
depends on https://github.com/OCA/contract/pull/1259


pedrobaeza (changes requested):
I don't agree with changing to LGPL, sorry. Please work with community and totally open modules.

PR #1267, opened 2025-06-30
[18.0][IMP] contract: make label for notes visible on contract view form

glitchov:
The note label is not visible on contract view form so i changed the it's declared to make it visible.

### DIFF
--- a/contract/views/contract.xml
+++ b/contract/views/contract.xml
@@ -256,7 +256,7 @@
                                     />
                                 </list>
                             </field>
-                            <field name="note" />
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Maybe it's better to put a placeholder instead. And please remove `[18.0]` from the commit message (not from the PR title, which is OK).

pedrobaeza (approved):

marcelsavegnago (approved):

baimont (approved):

PR #1268, opened 2025-06-30
[IMP] Avoid child_of for multi-company rules

samirGuesmi:


samirGuesmi:
Odoo is moving away from using the child_of operator in multi-company record rules.

PR #1273, opened 2025-07-18
[IMP] product_contract - non contract sale order lines should not have contract dates

baimont:
While migrating stuff from v12 to v18, we noticed that a behaviour changed in between these versions.

On a new database, install product_contract and sale.
Then, create a new product that is not a contract.
And finally, you create a sale order line with the new product.

In v12, dates won't be filled on the sale order line. In v18, they will. We think that it is not normal and that this behaviour should be the same as v12: non contract products in sale order lines should not fill dates fields.


qgroulard (approved):

luisDIXMIT (approved):
Tested and LGTM!

PR #1278, opened 2025-07-28
[18.0][FIX] contract: migration scripts

MiquelRForgeFlow:
This PR fixes 2 things:

- "Renew Contract lines" cron was moved in a split module too.
- If you don't install the split modules (they should be optional), you should at least remove all `noupdate=1` data that is moved.


marcelsavegnago (approved):
LGTM

MiquelRForgeFlow:
@JordiBForgeFlow

pedrobaeza (approved):

PR #1279, opened 2025-08-04
[17.0][FIX]subscription_oca: Search by user company cron

DorianMAG:
Adding a search filter specific to the user of the company to prevent the validation of invoices without being on the correct company


flotho (approved):
Hi, thanks,
Make sense and LGTM

PR #933, opened 2023-03-07
[14.0][MIG] product_contract

ilyasProgrammer:
based on https://github.com/OCA/contract/pull/652

PR #1280, opened 2025-08-04
[18.0][IMP] Contract line: compute display name

qgroulard:
Forward port of feature lost between branches `14.0` and `15.0`.

Compute the `display_name` as it was until `14.0` instead of always `False` as it is now.

Original commit: https://github.com/OCA/contract/pull/933/commits/72c77432932d28bf19d47d2dfe73d01e4f33c3cf

### DIFF
--- a/contract/models/contract_line.py
+++ b/contract/models/contract_line.py
@@ -42,6 +42,11 @@ class ContractLine(models.Model):
     )
     product_id = fields.Many2one(index=True)
 
+    @api.depends("name", "date_start")
+    def _compute_display_name(self):
+        for rec in self:
+            rec.display_name = f"{rec.date_start} - {rec.name}"
+
     @api.model
     def _compute_first_recurring_next_date(
         self,


contract/models/contract_line.py
```diff
@@ -42,6 +42,11 @@ class ContractLine(models.Model):
     )
     product_id = fields.Many2one(index=True)
 
+    @api.depends("name", "date_start")
+    def _compute_display_name(self):
+        for rec in self:
+            rec.display_name = self.env._("%s - %s") % (rec.date_start, rec.name)
```
pedrobaeza: This shouldn't be translatable AFAIK.

pedrobaeza:
Anyway, just now is not `False`. It's simply the `name` field, which is valid for most of the uses.

contract/models/contract_line.py
qgroulard: I thought it would let the opportunity to the users to edit the format as they wish.
I can remove it but I think it doesn't hurt anyway.

qgroulard:
> Anyway, just now is not `False`. It's simply the `name` field, which is valid for most of the uses.

You can try it on runboat: http://oca-contract-18-0-6517aabbc6bd.runboat.odoo-community.org/odoo/action-328/1

image

contract/models/contract_line.py
pedrobaeza: This has high chances of being broken by incorrect translations (someone removes a marker), and anyway, you can even not customize the order of both parameter, as they are not keyword placeholders. Please remove it.

qgroulard: Done

pedrobaeza (approved):

yvaucher (approved):

glitchov (approved):
LGTM

Issue #1275, opened 2025-07-19
[18.0 ] product_contract doesn't play nice with variant configurator


lk-eska:
## Module
product_contract

## Describe the bug
Variant configurator pops up only once when you add a product and contract configurator popup right after. Once you select the variant, there is no way to open up variant configurator, it doesn't appear anymore.

Another issue is that it overwrites the variant name with contract description with
"""{product}:
    - Recurrency: {recurring_rule}
    - Invoicing Type: {invoicing_type}
    - Date: {date_text}
  """
Better it appends this under the the variant name generated.

## To Reproduce
Make a product with variant a contract product and try to add it to sale order.

1. open sale order
2. add a line with contract + variant product
3. variant popup appears then contract popup appears
4. click on configure button, only contract popup appears

**Expected behavior**
It should open variant configuration then contract. (or another approach would be to use the variant configurator window.

pedrobaeza:
In v17, we handle that. Please check if working in runboat in 17, and you can forward-port that changes.

PR #1281, opened 2025-08-07
[18.0][FIX] product_contract:  divide variant and contract configuration button

bo-eska:
#1275

### DIFF
--- a/product_contract/static/src/js/sale_product_field.esm.js
+++ b/product_contract/static/src/js/sale_product_field.esm.js
@@ -1,7 +1,27 @@
-import {SaleOrderLineProductField} from "@sale/js/sale_product_field";
 import {patch} from "@web/core/utils/patch";
+import {SaleOrderLineProductField} from "@sale/js/sale_product_field";
 
 patch(SaleOrderLineProductField.prototype, {
+    setup() {
+        super.setup(...arguments);
+        this.lastContractData = false;
+    },
+
+    get extraLines() {
+        var res = super.extraLines;
+        if (
+            this.props.record.data.is_contract &&
+            this.props.record.data.product_contract_description
+        ) {
+            for (var val of this.props.record.data.product_contract_description.split(
+                "||"
+            )) {
+                res.push(val);
+            }
+        }
+        return res;
+    },
+
     async _onProductUpdate() {
         super._onProductUpdate(...arguments);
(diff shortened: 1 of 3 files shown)


CarlosRoca13 (changes requested):
I don't understand the reason for making this PR; without including these changes, the following button already exists:
image
That button serves the same purpose as the one you're adding. Could you provide more context on why this change is being made?

CarlosRoca13:
Okay, I see, is with variants configurator :sweat_smile:

CarlosRoca13 (changes requested):
Please squash the two commits and use proper naming for the change. It would also be helpful to include the reason for the change in the commit message to provide context.

https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#71commit-message

pedrobaeza:
CI is red. Can you please explain me visually which is the problem?

CarlosRoca13:
Why this PR is closed?

bo-eska:
> Why this PR is closed?

I tried to fix some tests

product_contract/static/src/js/sale_product_field.esm.js
```diff
@@ -1,40 +1,69 @@
-import {SaleOrderLineProductField} from "@sale/js/sale_product_field";
+/** @odoo-module **/
```
CarlosRoca13:
```suggestion
```
Remove this

pedrobaeza:
But why is needed to divide between template and variant? AFAIK, you already have a button to configure the variant (in the standard), and this other button to configure the contract options. Or that have been lost from 17?

CarlosRoca13:
> But why is needed to divide between template and variant? AFAIK, you already have a button to configure the variant (in the standard), and this other button to configure the contract options. Or that have been lost from 17?

It has been lost

lk-eska:
This was not a simple upgrade. It would be best for original author of the previous commit to check this @bo-eska @CarlosRoca13

CarlosRoca13 (commented):
I have tested and seems to be working correctly. 

But please set all changes in a single commit

bo-eska:
I set all changes in a single commit @CarlosRoca13

CarlosRoca13 (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #1282, opened 2025-08-12
[FIX] contract_forecast: multi-company rule for all authorized companies

glitchov:
Modification of the multi-company rule taking all authorized companies into account against the default company. This allows checking failed forecast jobs of child companies for example.


qgroulard (approved):
:+1: 

Using the default company of the user (field `company_id`) doesn't make sense anymore.

yvaucher (approved):

Issue #1277, opened 2025-07-19
[18.0] product_contract additional configuration (auto price, manual renew required etc.)


lk-eska:
**Is your feature request related to a problem?**
No

**Describe the solution you'd like**
It would be very good if auto price, manual renew required fields to be selected on the contract product.

**Describe alternatives you've considered**
Now you need to set them on the contract

pedrobaeza:
Yeah, it can be a good enhancement. Can you contribute it? Avout auto_renew, I would say that feature is now split from `contract` itself, so a new glue module should be done.

PR #1284, opened 2025-08-18
[FIX] product_contract: add auto price and manual renew needed

bo-eska:
#1277


pedrobaeza:
Shouldn't this go in a glue module?

bo-eska:
> Shouldn't this go in a glue module?

Since the `automatic_price` and `manual_renew_needed` fields are already defined in the [contract_line_successor](https://github.com/OCA/contract/blob/01f9a0c258d1e3ad8dbee4c37fb78a562563b0c6/contract_line_successor/models/contract_line.py#L42) and contract.line models in the [contract](https://github.com/OCA/contract/blob/01f9a0c258d1e3ad8dbee4c37fb78a562563b0c6/contract/models/contract_template_line.py#L61) module, wouldn't it be more convenient to add these fields by default in the configurator in sale.order.line in product_contract module?

pedrobaeza:
No, as that would make `product_contract` depending on that module. I'm seeing that the module already depends on `product_termination`, which is also not correct for a proper modularity.

lk-eska:
contract_termination is already a dependency of product_contract and seems to be designed that way. manual_renew_needed is inside contract_line_successor anyway and I am not sure if it is correct. contract module is divided into modules that depend one to another serially (contract<contract_line_successor<contract_termination<product_contract). I believe what you propose requires some serious refactor. @pedrobaeza

pedrobaeza:
No, I'm just proposing adding these extra fields in a new module `product_contract_successor`, which is the correct way to do it. And I will extract `contract_termination` dependency from `product_contract` as well.

lk-eska:
@bo-eska would you test removing dependency? if it works properly it's best to follow @pedrobaeza 's suggestion.

bo-eska:
> No, I'm just proposing adding these extra fields in a new module `product_contract_successor`, which is the correct way to do it. And I will extract `contract_termination` dependency from `product_contract` as well.

So you mean like defining `termination_notice_interval` `termination_notice_rule_type` fields and [constrains](https://github.com/OCA/contract/blob/64065ca12becad5ef9f3c7e38297b3d8bf4c5aef/product_contract/models/sale_order.py#L17) in a module other than product_contract and making product_contract depend on it?

lk-eska:
@bo-eska If I understand correctly all those dependencies have to go into two seperate modules (product_contract_successor & product_contract_termination).

Is that correct @pedrobaeza

pedrobaeza:
That's it, @lk-eska !

PR #1289, opened 2025-08-25
[18.0] [ADD] product_contract_recurrence_in_price 

qgroulard:
This module extends the functionality of ``product_contract`` to add an option to compute
the total amounts on "contract" Sale Order lines, including every invoicing planned.

For instance, a product worth 10€ invoiced every month for one year would have a total
of 120€ (unit_price * #invoices) instead of 10€ (unit_price * quantity).


yvaucher (approved):

product_contract/views/sale_order.xml
```diff
@@ -121,7 +121,7 @@
                     domain="[('contract_id','=',contract_id)]"
                     optional="hide"
                 />
-                <field name="recurrence_number" optional="hide" />
+                <field name="recurrence_number" optional="show" />
```
pedrobaeza: Please don't change this.

qgroulard: Don't you think it will be confusing for the users ?
The total of the line will be different than qty*price, but without showing the column "recurrence_number" it's difficult to understand why.

pedrobaeza: By default that field is 1, so no difference. And showing number but not interval is for me the same confusing, don't you think?

qgroulard: Default is 1 but if you install this module this is most probably to play with this number.
I'm ok with showing the interval too.
My point is that the price is key on a sale order, imo every factor of the price should clearly be displayed to the user. Else, the SO might become hard to understand or look buggy.
I can already hear my customers scream at me because the SO shows wrong numbers :sweat_smile:

pedrobaeza: But as the module includes its own dialog for configuring this, what we don't want is that the list includes a lot of columns, because that's a bad practice, so please keep both hidden by default. If the user needs them, they can show both.

qgroulard:
I have pushed a new version.
The first approach was a bit naive, it was based on the field `recurrence_number`, which is not the actual number of times the line will be invoiced.
Thus, I add a new field `recurring_invoicing_number` for this.

pedrobaeza:
But why do you need an extra multiplier and the quantity is not enough?

qgroulard:
> But why do you need an extra multiplier and the quantity is not enough?

Let's say you sell a subscription over 3 years, invoiced every quarter with a unit price of 100€.

Prior to https://github.com/OCA/contract/commit/f50e15dbe72e09b07bcf7c2763cf00c45c30e67c, you would have:
- `price_unit` = 100.0
- `product_uom_qty` = 12 (there are 12 quarters in 3 years)

Thus the subtotal of the line would be 1200€ (`price_unit` * `product_uom_qty`).
It makes sense, in total the customer will have to pay 1200€ for this subscription.

After https://github.com/OCA/contract/commit/f50e15dbe72e09b07bcf7c2763cf00c45c30e67c, you would have:
- `price_unit` = 100.0
- `product_uom_qty` = 1
- `recurring_interval` = 1
- `recurring_rule_type` = "quarterly"

Without this PR, the subtotal of the line would be 100€ (`price_unit` * `product_uom_qty`), which is wrong.
With this PR, we have:
- `recurring_invoicing_number` = 12 (there are 12 quarters in 3 years) / 1 (`recurring_interval`)

And the subtotal of the line would be 1200€: 100 (`price_unit`) * 12 (`recurring_invoicing_number`) * 1 (`product_uom_qty`)

pedrobaeza:
The original intention of the system is that you put on the `sale.order.line` the "setup" price, and then, when confirmed, the contract is created for the recurring "maintenance" price. What you are telling is that you put in the quotation the full price of the whole "maintenance" price period. That use case sounds very weird, and I think it's over-complicating this module, so it should be handled on extra modules on top.

qgroulard:
> The original intention of the system is that you put on the `sale.order.line` the "setup" price, and then, when confirmed, the contract is created for the recurring "maintenance" price. What you are telling is that you put in the quotation the full price of the whole "maintenance" price period. That use case sounds very weird, and I think it's over-complicating this module, so it should be handled on extra modules on top.

The use case is the sale of subscriptions.
Recurrences are not maintenance cost, this is a service that the customer pays every month for X years.

I understand that this module can solve different use cases, therefore I have added this as a new option in a new module.

lmarion-source (approved):

product_contract_recurrence_in_price/models/sale_order_line_contract_mixin.py
```diff
@@ -0,0 +1,64 @@
...
+    )
+
+    @api.depends("product_id")
+    def _compute_product_contract_data(self):
+        res = super()._compute_product_contract_data()
+        for rec in self:
+            if rec.product_id.is_contract:
+                rec.include_recurrence_in_price = (
+                    rec.product_id.include_recurrence_in_price
+                )
```
AnizR: You aren't setting your field in every cases
```suggestion
                rec.include_recurrence_in_price = rec.product_id.is_contract and (
                    rec.product_id.include_recurrence_in_price
                )
```

AnizR (changes requested):

product_contract_recurrence_in_price/models/sale_order_line_contract_mixin.py
qgroulard: Done :+1:

AnizR (approved):

yvaucher:
I'm merging as it has enough approval, plus the module is separate thus shouldn't affect other module workflow.

yvaucher:
Can you rebase?

qgroulard:
> Can you rebase?

Done but the error doesn't seem to be fixed on base branch.

yvaucher:
The error seems indeed unrelated

In tests module `test_subscription_oca` in test TestSubscriptionOCA.test_subscription_oca_sub8_workflow
```The company bank account linked to this invoice is not trusted, please double-check and trust it before confirming, or remove it```

qgroulard:
> The error seems indeed unrelated
> 
> In tests module `test_subscription_oca` in test TestSubscriptionOCA.test_subscription_oca_sub8_workflow `The company bank account linked to this invoice is not trusted, please double-check and trust it before confirming, or remove it`

Yes, we can see it on other PR as well: https://github.com/OCA/contract/pull/1392, https://github.com/OCA/contract/pull/1390 for instance.

qgroulard:
Seems to be fixed @yvaucher

yvaucher:
/ocabot merge nobump


Let's try again

PR #1184, opened 2025-01-20
[18.0][MIG] contract_price_revision

AlvaroRM11:
@BinhexTeam
Depends on: https://github.com/OCA/contract/pull/1136

PR #1290, opened 2025-08-25
[18.0][MIG] contract_price_revision: Migration to 18.0

victoralmau:
Superseed https://github.com/OCA/contract/pull/1184

Migration to 18.0

@Tecnativa TT57315


contract_price_revision/__manifest__.py
```diff
@@ -6,12 +6,12 @@
...
     "name": "Contract Price Revision",
     "summary": "Easy revision of contract prices",
-    "version": "17.0.1.0.0",
+    "version": "18.0.1.0.0",
     "category": "Contract",
     "author": "ACSONE SA/NV, Tecnativa, Odoo Community Association (OCA)",
     "license": "AGPL-3",
     "website": "https://github.com/OCA/contract",
-    "depends": ["contract"],
+    "depends": ["contract_line_successor"],
```
pedrobaeza, line 14: This dependency shouldn't be changed. This module should be independent from the other one, not setting that successor fields (they are only informative anyway).

victoralmau: This dependency is necessary because the `predecessor_contract_line_id` field (used in https://github.com/OCA/contract/blob/d1709eddc7dfca0123e81f71433f0e94a6f2b698/contract_price_revision/models/contract_line.py# L13) is defined by the `contract_line_successor` module.

pedrobaeza: Uhm, I don't like too much to have it, but OK.

pedrobaeza (approved):
OK then.

contract_price_revision/security/ir.model.access.csv
```diff
@@ -0,0 +1,3 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_contract_price_revision_wizard_account_invoice,access_contract_price_revision_wizard_account_invoice,model_contract_price_revision_wizard,account.group_account_invoice,1,1,1,1
+access_contract_price_revision_wizard_account_manager,access_contract_price_revision_wizard_account_manager,model_contract_price_revision_wizard,account.group_account_manager,1,1,1,1
```
carlos-lopez-tecnativa: For me, this access right is duplicated and can be deleted, because the group_account_manager inherits from group_account_invoice, so the access right is inherited as well.

carlos-lopez-tecnativa (approved):
LGTM, just a non-blocking comment, but if you can apply it, that would be cool.

carlosdauden (approved):

PR #1294, opened 2025-09-01
[ADD] contract_invoice_queue_job

tobiaszehntner:
To avoid server time-outs when invoicing large number of contracts, this module delegates the task to batches of configurable sizes. Available both for scheduled and manual invoicing.

PR #1298, opened 2025-09-11
[FIX] contract.recurring_create_invoice: Fix singleton error

tobiaszehntner:


contract/models/contract.py
```diff
@@ -346,7 +346,7 @@ def recurring_create_invoice(self):
             body = Markup(
                 self.env._("Contract manually invoiced: %(invoice_link)s")
             ) % {"invoice_link": invoice._get_html_link(title=invoice.name)}
-            self.message_post(body=body)
+            invoice.message_post(body=body)
```
sbejaoui: The idea here is to log in the contract whenever the invoicing action is triggered manuall

```suggestion
            for contract in invoice.invoice_line_ids.contract_line_id.contract_id:
                contract.message_post(body=body)
```

sbejaoui (changes requested):

contract/models/contract.py
```diff
@@ -346,7 +346,8 @@ def recurring_create_invoice(self):
             body = Markup(
                 self.env._("Contract manually invoiced: %(invoice_link)s")
             ) % {"invoice_link": invoice._get_html_link(title=invoice.name)}
-            self.message_post(body=body)
+            for contract in invoice.invoice_line_ids.contract_line_id.contract_id:
```
acsonefho: Is it possible to avoid nested loop?

tobiaszehntner: My singleton error was due to using the wrong function in https://github.com/OCA/contract/pull/1294. @sbejaoui I added an ensure_one for clarity

sbejaoui (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #1301, opened 2025-09-15
[16.0][IMP] subscription_oca: add the tag is_move_sent True when the invoice from subscription is sent by email

mondot:
Small update to add the missing `is_move_sent` attribute on the invoice when it has been sent by email.
It concerns only the invoicing_mode `invoice_send`.

### DIFF
--- a/subscription_oca/tests/test_subscription_oca.py
+++ b/subscription_oca/tests/test_subscription_oca.py
@@ -506,6 +506,7 @@ class TestSubscriptionOCA(TransactionCase):
         subscription.generate_invoice()
         subscription.template_id.invoicing_mode = "invoice_send"
         subscription.generate_invoice()
+        self.assertEqual(1, len(subscription.invoice_ids.filtered("is_move_sent")))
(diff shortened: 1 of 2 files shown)


subscription_oca/tests/test_subscription_oca.py
```diff
@@ -506,6 +506,21 @@ def test_subscription_oca_sub8_workflow(self):
         subscription.generate_invoice()
         subscription.template_id.invoicing_mode = "invoice_send"
         subscription.generate_invoice()
+        self.assertTrue(
+            len(
+                self.env["account.move"].search(
```
rousseldenis: @mondot You can do :

```suggestion
                self.assertEqual(
                    1,
                    len(subsription.invoice_ids.filtered("is_move_sent"))
                )
```

instead.

mondot: Great, thanks for the feedback :)

tarteo (approved):

rousseldenis (approved):

PR #1233, opened 2025-05-12
[18.0][MIG] contract_sale_payment_mode: Migration to 18.0

bjouini-acsone:

PR #1212, opened 2025-03-17
[18.0][MIG] contract_payment_mode:  Migration to 18.0

AndreuOForgeFlow:
Depends on:
- https://github.com/OCA/contract/pull/1136
- https://github.com/OCA/bank-payment/pull/1439

@Forgeflow

PR #1302, opened 2025-09-16
18.0 fix contract payment mode

bjouini-acsone:


contract_payment_mode/models/contract.py
```diff
@@ -17,6 +17,8 @@ class ContractContract(models.Model):
     @api.depends("partner_id", "contract_type")
     def _compute_payment_mode_id(self):
         for rec in self:
+            if rec.payment_mode_id:
```
sbejaoui, line 20: this change was refused in https://github.com/OCA/contract/pull/1212#discussion_r2126844853, I propose you make it in contract_sale_payment_mode

sbejaoui (changes requested):

contract_payment_mode/models/contract.py
bjouini-acsone: @sbejaoui it is  fixed in this pr now :https://github.com/OCA/contract/pull/1233
i will close this one

PR #1305, opened 2025-09-19
[16.0][FIX] contract_sale_generation: fix tests not running

huguesdk:
fix tests not running because `ContractSaleCommon` was tagged as `post_install` but that only works on the test classes themselves.

### DIFF
--- a/contract_sale_generation/tests/common.py
+++ b/contract_sale_generation/tests/common.py
@@ -5,14 +5,13 @@
 from freezegun import freeze_time
 
 from odoo import fields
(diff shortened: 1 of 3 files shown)


huguesdk:
please note also that although there is a test file called [`test_contract_sale_recurrency.py`](https://github.com/OCA/contract/blob/7da8bfc6f6b1b469eebef72e3b2533cb0e5a665c/contract_sale_generation/tests/test_contract_sale_recurrency.py), it is not imported by `tests/__init__.py` so it does not run. i tried to run it but it causes errors.

PR #1312, opened 2025-10-01
[19.0] [MIG] contract: Migration to 19.0

BhaveshHeliconia:


tishmen:
@BhaveshHeliconia can you please handle the failing tests so we can push this forward?

mamcode:
@BhaveshHeliconia  could you please re-run the CI checks?
I’ve tested locally with the latest ghcr.io/oca/oca-ci:py3.11-19.0 image, and it seems the CI is using an outdated cache because I am getting different errors from unit tests.

BhaveshHeliconia:
@tishmen @mamcode : CI is green now. Thanks!

pedrobaeza:
/ocabot migration contract

Please cherry-pick #1343

BhaveshHeliconia:
> /ocabot migration contract
> 
> Please cherry-pick #1343

@pedrobaeza : Done. Thanks!

contract/models/contract_template_line.py
```diff
@@ -42,18 +42,14 @@ class ContractTemplateLine(models.Model):
         readonly=False,
     )
     quantity = fields.Float(default=1.0, required=True)
-    product_uom_category_id = fields.Many2one(
-        comodel_name="uom.category",
-        related="product_id.uom_id.category_id",
-        readonly=True,
-    )
+    allowed_uom_ids = fields.Many2many("uom.uom", compute="_compute_allowed_uom_ids")
```
pedrobaeza: Please use `Binary` field for the domain and compute it there instead of the many2many, that is less performant.

contract/tests/test_contract.py
```diff
@@ -11,30 +11,59 @@
...
-class TestContractBase(common.TransactionCase):
+class TestContractBase(BaseCommon):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
         cls.today = fields.Date.today()
         cls.pricelist = cls.env["product.pricelist"].create(
             {"name": "pricelist for contract test"}
         )
+        cls.uom_hour = cls.env["uom.uom"].create(
```
pedrobaeza: I don't think you need to create a uom, as standard one aren't demo data.

contract/tests/test_contract.py
```diff
@@ -1460,18 +1489,26 @@ def test_recurrency_propagation(self):
         self.contract.write(vals2)
         for field in vals:
             self.assertEqual(vals[field], self.acct_line[field])
-        # New contract
-        contract_form = Form(self.env["contract.contract"])
```
pedrobaeza: Why not using the `Form`? It's allowed on tests.

contract/tests/test_portal.py
```diff
@@ -16,10 +16,22 @@ def test_tour(self):
         contract = self.env["contract.contract"].create(
             {"name": "Test Contract", "partner_id": partner.id}
         )
-        user_portal = self.env.ref("base.demo_user0")
+        portal_group = self.env.ref("base.group_portal")
+        user_portal = (
```
pedrobaeza: There's a method in `BaseCommon` for creating easily a user. Please use it instead.

BhaveshHeliconia:
@pedrobaeza : I’ve completed all the changes you recommended. Thanks for the feedback!

contract/models/contract_template_line.py
```diff
@@ -42,18 +42,16 @@ class ContractTemplateLine(models.Model):
         readonly=False,
     )
     quantity = fields.Float(default=1.0, required=True)
-    product_uom_category_id = fields.Many2one(
-        comodel_name="uom.category",
-        related="product_id.uom_id.category_id",
-        readonly=True,
+    allowed_uom_ids = fields.Binary(
```
pedrobaeza: I meant to have the complete domain as Binary. Please check for example https://github.com/OCA/bank-payment/blob/03f557c4c53929db8a44404813359604d37b40f4/account_payment_order/wizard/account_payment_line_create.py#L51

contract/models/contract_template_line.py
```diff
@@ -153,13 +151,31 @@ def _compute_name(self):
...
     def _compute_uom_id(self):
         for line in self:
-            if not line.uom_id or (
-                line.product_id.uom_id.category_id.id != line.uom_id.category_id.id
-            ):
-                line.uom_id = line.product_id.uom_id
+            if not line.product_id:
+                line.uom_id = False
+                continue
+            if line.allowed_uom_ids:
```
pedrobaeza: And here you will use `filtered_domain`.

contract/tests/test_contract.py
```diff
@@ -33,8 +35,28 @@ def setUpClass(cls):
...
             }
         )
-        cls.product_1 = cls.env.ref("product.product_product_1")
-        cls.product_2 = cls.env.ref("product.product_product_2")
+        cls.product_1 = cls.env["product.product"].create(
+            {
+                "name": "Virtual Interior Design",
+                "categ_id": cls.env.ref(
+                    "product.product_category_services", raise_if_not_found=False
+                ).id,
```
pedrobaeza: Don't put `raise_if_not_found`, as at the end, if it not exists, you'll get the not null exception.

contract/tests/test_contract.py
```diff
@@ -258,23 +262,6 @@ def test_automatic_price(self):
         self.acct_line.invalidate_model()
         self.assertEqual(self.acct_line.price_unit, 10)
 
-    def test_automatic_price_min_quantity(self):
```
pedrobaeza: Don't remove tests.

contract/tests/test_contract.py
```diff
@@ -70,24 +92,6 @@ def setUpClass(cls):
                 "base": "list_price",
             }
         )
-        cls.env["product.pricelist.item"].create(
```
pedrobaeza: This data was put for the specific new feature added, so don't remove it.

pedrobaeza:
Please include #1341

BhaveshHeliconia:
> Please include #1341

Done. Thanks!

carlos-lopez-tecnativa:
Please include https://github.com/OCA/contract/pull/1351 and https://github.com/OCA/contract/pull/1345

BhaveshHeliconia:
> Please include #1351 and #1345

Done. Thanks!

victoralmau:
Please, cherry-pick https://github.com/OCA/contract/pull/1352 to commit history (before migration commit).

BhaveshHeliconia:
> Please, cherry-pick #1352 to commit history (before migration commit).

Done. Thanks!

pedrobaeza:
/ocabot migration contract

Please include #1360

bosd (approved):
Functional test :+1:

contract/models/contract_template_line.py
pedrobaeza: The usual convention is to name this as `field_domain`, which would be in this case `uom_id_domain`.

contract/models/contract.py
```diff
@@ -682,7 +682,7 @@ def _cron_recurring_create(self, date_ref=False, create_type="invoice"):
         if not date_ref:
             date_ref = fields.Date.context_today(self)
         domain = self._get_contracts_to_invoice_domain(date_ref)
```
pedrobaeza, line 680: In this method, it's better to return a `Domain`, isn't it?

pedrobaeza (commented):
Thanks, I think we are almost there.

contract/models/contract.py
BhaveshHeliconia: Yes, correct! I'll update it to return a Domain object instead of a list.

etobella:
@BhaveshHeliconia can you add 

- https://github.com/OCA/contract/pull/1378
- https://github.com/OCA/contract/pull/1358

BhaveshHeliconia:
> @BhaveshHeliconia can you add
> 
> * [[18.0][FIX] contract: Use another condition for forcingn a failure #1378](https://github.com/OCA/contract/pull/1378)
> * [[18.0][IMP] contract: add paths to actions #1358](https://github.com/OCA/contract/pull/1358)

Done. Thanks.

contract/tests/test_contract.py
```diff
@@ -1388,7 +1408,7 @@ def test_multicompany_partner_edited(self):
         # Update the partner attached to both contracts
         self.partner.with_user(unprivileged_user).with_company(company2).with_context(
             company_id=company2.id
-        ).write({"is_company": False, "parent_id": parent_partner.id})
+        ).sudo().write({"is_company": False, "parent_id": parent_partner.id})
```
etobella: why a sudo?

stferraro (approved):

stferraro (approved):

stferraro:
@pedrobaeza please review module and approved this PR

pedrobaeza:
@BhaveshHeliconia please include #1381

carlos-lopez-tecnativa:
> @BhaveshHeliconia please include #1381

@BhaveshHeliconia, please check the [latest changes](https://github.com/OCA/contract/compare/4489fb571766ef4d1f1bd35cb08b6a092bf9dab2..718590726c2856876267abbf1d054dc92bb6865e) in this PR  according to my comment: https://github.com/OCA/contract/pull/1381#discussion_r2737016316 to preserve the `optional` attribute on the field `name`.

contract/models/contract_line.py
```diff
@@ -93,19 +93,19 @@ def _check_last_date_invoiced(self):
                 raise ValidationError(
                     self.env._(
                         "You can't have the end date before the date of last "
-                        "invoice for the contract line '%s'"
+                        "invoice for the contract line '%(name)s'",
```
pedrobaeza: Were these changes really proposed by pre-commit? I don't think so, because when there's only one param, there's no possible order swap in the translations to require to name them. Right now, you have invalidated all the existing translations with this change.

contract/models/contract_template_line.py
```diff
@@ -153,12 +153,40 @@ def _compute_name(self):
...
+    @api.depends("product_id", "product_id.uom_id")
     def _compute_uom_id(self):
         for line in self:
-            if not line.uom_id or (
-                line.product_id.uom_id.category_id.id != line.uom_id.category_id.id
-            ):
+            if not line.product_id:
+                line.uom_id = False
+                continue
+            if line.uom_id_domain:
```
pedrobaeza: Why all these changes in the UoM handling (as well as the domain)? This doesn't require such long code. Check for example `account`: https://github.com/odoo/odoo/blob/42cf692eda72655be10d8519ca56baa695f28c27/addons/account/models/account_move_line.py#L846

contract/tests/test_contract.py
```diff
@@ -1378,6 +1398,10 @@ def test_multicompany_partner_edited(self):
                 "login": "test",
                 "company_id": company2.id,
                 "company_ids": [(4, company2.id, False)],
+                "group_ids": [
```
pedrobaeza: You can use the method `odoo.tests.new_test_user` for simplifying this (and other user creation).

contract/models/contract_template_line.py
```diff
@@ -0,0 +1,298 @@
...
+    product_id = fields.Many2one("product.product", string="Product")
+    name = fields.Text(
+        string="Description",
+        required=True,
+        compute="_compute_name",
+        store=True,
+        readonly=False,
+    )
+    quantity = fields.Float(default=1.0, required=True)
+    uom_id_domain = fields.Binary(
```
bosd: The uom_id_domain = fields.Binary field causes a JS error in Odoo 19 when users try to change the UoM dropdown on contract lines:

  `InvalidDomainError: Invalid domain representation: 19.00 bytes`

  The web client renders the Binary field value as a file size string ("19.00 bytes") instead of passing the JSON domain content to
  `getFieldDomain()`.

   When used as `domain="uom_id_domain"` on a Many2one, the JS client cannot interpret the Binary value.

  Odoo core's own sale.order.line both use Many2many for this exact purpose:
```
  allowed_uom_ids = fields.Many2many('uom.uom', compute='_compute_allowed_uom_ids')
  product_uom_id = fields.Many2one(domain="[('id', 'in', allowed_uom_ids)]", ...)
```
  The field should follow the same pattern to be compatible with the Odoo 19 JS client.

bosd (changes requested):

contract/models/contract_template_line.py
pedrobaeza: Has this changed in 19? This was valid across several versions...

bosd: I think so.
In 19 it is possible to change the uom of a product. ( when already used on so)

In this case, the uom on the product had changed. But the contract line still refers to the old uom. When trying to correct the uom on the line ( to correctly set it the same as on the product ) the javascript error is raised.

contract/models/contract.py
```diff
@@ -0,0 +1,706 @@
...
+    def _onchange_partner_id(self):
+        partner = (
+            self.partner_id
+            if not self.company_id
+            else self.partner_id.with_company(self.company_id)
+        )
+        self.pricelist_id = partner.property_product_pricelist.id
+        self.fiscal_position_id = partner.env[
+            "account.fiscal.position"
+        ]._get_fiscal_position(partner)
```
bosd: This  might be better suited for a follow up improvement pr..
But when selecting a "contact" on the contract the fiscal position of the contact is used instead of the one from the partner. 

```suggestion
       fiscal_partner = partner.commercial_partner_id or partner
        self.fiscal_position_id = partner.env[
            "account.fiscal.position"
        ]._get_fiscal_position(fiscal_partner)   
```

bosd (approved):

florian-dacosta:
@BhaveshHeliconia Please include https://github.com/OCA/contract/pull/1397

bosd:
@BhaveshHeliconia On one of our production instances we noticed an issue with this module due to a odoo core update.
More details can be found in pr #1409 
Can you please cherry-pick the fix commit [519819b](https://github.com/OCA/contract/pull/1409/commits/519819bef59dda63ba4bc20f4b9b56db0fd67e6d)

muaazsiddiq:
@BhaveshHeliconia For this PR, I believe you have based your branch from 18.0 instead of 19.0. Can you please fix this ?. Otherwise there will be an unrelated history error while merging.

BhaveshHeliconia:
> @BhaveshHeliconia For this PR, I believe you have based your branch from 18.0 instead of 19.0. Can you please fix this ?. Otherwise there will be an unrelated history error while merging.

@muaazsiddiq Thanks for the comment. From what I can see, the branch is correctly based on 19.0 — HeliconiaIO:19.0-mig-contract is targeting OCA:19.0, and the commit history looks clean without unrelated changes.

Could you please point out if you’re seeing something specific that suggests it’s based on 18.0? I’m happy to double-check.

muaazsiddiq:
> > @BhaveshHeliconia For this PR, I believe you have based your branch from 18.0 instead of 19.0. Can you please fix this ?. Otherwise there will be an unrelated history error while merging.
> 
> @muaazsiddiq Thanks for the comment. From what I can see, the branch is correctly based on 19.0 — HeliconiaIO:19.0-mig-contract is targeting OCA:19.0, and the commit history looks clean without unrelated changes.
> 
> Could you please point out if you’re seeing something specific that suggests it’s based on 18.0? I’m happy to double-check.

Thanks for your response it is solved. It was a commit depth issue on my end

raulovallet:
> @BhaveshHeliconia On one of our production instances we noticed an issue with this module due to a odoo core update. More details can be found in pr #1409 Can you please cherry-pick the fix commit [519819b](https://github.com/OCA/contract/pull/1409/commits/519819bef59dda63ba4bc20f4b9b56db0fd67e6d)

Hey guys, I did exactly that some time ago. You can approve it; I’m using it in production as well. I’d be happy to contribute, and if I notice anything else, I’ll let you know.

AlvaroRamirez10:
Hi! Thanks for the migration work.
While trying to integrate this PR with doodba/git-aggregator, the merge fails with: fatal: refusing to merge unrelated histories.
I confirmed that OCA/contract:19.0 and the PR head (HeliconiaIO:19.0-mig-contract) have no common ancestor (git merge-base origin/19.0 <pr-head> returns none).
Could you please rebase HeliconiaIO:19.0-mig-contract on top of OCA/contract:19.0 (or alternatively cherry-pick onto a branch based on 19.0) and force-push the updated branch?
After that, we should be able to aggregate/merge it normally. Thanks!

pedrobaeza:
@AlvaroRamirez10 you have to change $DEPTH_MERGE to 500 or something higher due to the big quantity of commits that this PR has.

eduezerouali-tecnativa (approved):
@BhaveshHeliconia please take a look at this https://github.com/OCA/oca-addons-repo-template/pull/344 to fix test

Everything else LGTM. 
cc @Tecnativa TT58611

bosd:
@BhaveshHeliconia Can you please rebase?

bosd:
Spotted a pre-existing bug while preparing the migration of contract_update_last_date_invoiced (#1368). Not introduced by this PR, 
 but flagging here:
                                                                                                                                    
 last_date_invoiced is declared on contract.recurring.mixin (models/contract_recurring_mixin.py) and therefore inherited by both    
 contract.line and contract.contract. Only the line variant is ever written — _update_last_date_invoiced (models/contract_line.py) 
 is invoked from _prepare_recurring_invoices_values on contract_lines only (models/contract.py:626). Nothing ever sets              
 contract.contract.last_date_invoiced, so on the contract record the field is always empty — even after invoices have been 
 generated. Confirmed on the runboat of this PR.

 Knock-on effect: the mixin's _compute_next_period_date_start and _compute_next_period_date_end on contract.contract always fall    
 back to date_start because last_date_invoiced is empty. The user-visible damage is small because recurring_next_date and date_end 
 are independently overridden on contract.contract to aggregate from lines (contract.py:172 and contract.py:206) —                  
 last_date_invoiced is just the missing third member of that pattern.

 I'd like to keep this PR narrow (it's already a large mig). I'll open a follow-up [IMP] PR depending on #1312 with: a stored       
 compute on contract.contract aggregating max of the lines' last_date_invoiced (mirroring how recurring_next_date aggregates with 
 min), tests, list-view + search-view exposure, and a few related UX improvements (kanban view, activities, decorations). That keeps
  backporting to 18.0/17.0/16.0 clean.

bosd (commented):
Three small drive-by suggestions noted while reviewing the diff for the follow-up `[IMP]` PR I'm preparing on top of this one. Surfaced by an automated review and worth considering here, since fixing them upstream avoids the same diff in every backport.

(Skipped two other items the bot flagged: it suggested reverting `request.env._()` calls back to a top-level `_` import, but `request.env._()` is the recommended idiom in 18+/19 and what this PR already uses, so no change there.)

contract/controllers/main.py
```diff
@@ -0,0 +1,110 @@
...
+                "label": request.env._("Date"),
+                "order": "recurring_next_date desc",
+            },
+            "name": {"label": request.env._("Name"), "order": "name desc"},
+            "code": {"label": request.env._("Reference"), "order": "code desc"},
+        }
+        # default sort by order
+        if not sortby:
+            sortby = "date"
+        order = searchbar_sortings[sortby]["order"]
```
bosd, line 63: Defensive fix: a malformed `?sortby=foo` URL parameter currently raises a `KeyError` on this line. Validate against `searchbar_sortings` before indexing.

```suggestion
        if not sortby or sortby not in searchbar_sortings:
```

(Replace the existing `if not sortby:` line.)

contract/data/mail_template.xml
```diff
@@ -0,0 +1,118 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo noupdate="1">
+    <record id="email_contract_template" model="mail.template">
+        <field name="name">Email Contract Template</field>
+        <field
+            name="email_from"
+        >{{ (object.user_id.email and '%s &lt;%s&gt;' % (object.user_id.name, object.user_id.email) or '') }}</field>
```
bosd: Use `email_formatted`, which already produces the `"Name" <email>` form and handles missing email gracefully, instead of building it by hand.

```suggestion
        <field
            name="email_from"
        >{{ object.user_id.email_formatted or '' }}</field>
```

contract/data/mail_template.xml
```diff
@@ -0,0 +1,118 @@
...
+                    style="background-color:#875A7B;padding: 8px 16px 8px 16px; text-decoration: none; color: #fff; border-radius: 5px; font-size:13px;"
+                >View contract</a>
+            </div>
+        </field>
+    </record>
+    <record id="mail_template_contract_modification" model="mail.template">
+        <field name="name">Contract Modification Template</field>
+        <field
+            name="email_from"
+        >{{ (object.user_id.email and '%s &lt;%s&gt;' % (object.user_id.name, object.user_id.email) or '') }}</field>
```
bosd: Same as the previous comment — use `email_formatted` here too for the modification template.

```suggestion
        <field
            name="email_from"
        >{{ object.user_id.email_formatted or '' }}</field>
```

BhaveshHeliconia:
@bosd Thanks for pointing this out! You’re right — last_date_invoiced on contract.contract is currently never populated.

Since recurring_next_date and date_end are already computed independently from the contract lines, the practical impact for users is quite limited.

contract/data/mail_template.xml
BhaveshHeliconia: Thanks for the suggestion — I’ve updated it accordingly.

bosd:
@pedrobaeza Is this one ok for you now? Can you please merge :pray: :rocket:

contract/models/account_move.py
```diff
@@ -0,0 +1,13 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    # We keep this field for migration purpose
+    old_contract_id = fields.Many2one("contract.contract")
```
carlos-lopez-tecnativa: Is this field still necessary?

contract/models/contract.py
```diff
@@ -0,0 +1,707 @@
...
+                    field.readonly,
+                    field.company_dependent,
+                    field.name in self.NO_SYNC,
+                )
+            ):
+                if (
+                    self.contract_template_id[field_name]
+                    and self.contract_template_id[field_name] != self[field_name]
+                ):
+                    self[field_name] = self.contract_template_id[field_name]
```
carlos-lopez-tecnativa, line 278: When a contract template is set, the recurrence is defined at the line level, so I think we should mark the `line_recurrence` field as True; otherwise, I can’t see the contract lines. This could be confusing for the user. What do you think about this change?

Or at least show the `line_recurrence` field in the contract template. This field exists, but it is not displayed in the form view. Why?

Image

contract/security/contract_security.xml
```diff
@@ -0,0 +1,46 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo noupdate="1">
+    <record id="rule_contract_contract_multi_company" model="ir.rule">
+        <field name="name">Contract contract multi-company</field>
+        <field name="model_id" ref="model_contract_contract" />
+        <field name="global" eval="True" />
+        <field name="domain_force">[('company_id', 'in', company_ids + [False])]</field>
```
carlos-lopez-tecnativa: In the contract, the company_id field is required, so False is unnecessary.


```suggestion
        <field name="domain_force">[('company_id', 'in', company_ids)]</field>
```

contract/security/contract_security.xml
```diff
@@ -0,0 +1,46 @@
...
+        <field
+            name="domain_force"
+        >[('message_partner_ids', 'in', [user.partner_id.id])]</field>
+        <field name="groups" eval="[Command.link(ref('base.group_portal'))]" />
+    </record>
+    <record id="rule_contract_line_multi_company" model="ir.rule">
+        <field name="name">Contract line multi-company</field>
+        <field name="model_id" ref="model_contract_line" />
+        <field name="global" eval="True" />
+        <field name="domain_force">[('company_id', 'in', company_ids + [False])]</field>
```
carlos-lopez-tecnativa: The same here

```suggestion
        <field name="domain_force">[('company_id', 'in', company_ids)]</field>
```

contract/security/contract_security.xml
```diff
@@ -0,0 +1,46 @@
...
+    </record>
+    <record id="rule_contract_template_multi_company" model="ir.rule">
+        <field name="name">Contract template multi-company</field>
+        <field name="model_id" ref="model_contract_template" />
+        <field name="global" eval="True" />
+        <field name="domain_force">[('company_id', 'in', company_ids + [False])]</field>
```
carlos-lopez-tecnativa:
```suggestion
        <field name="domain_force">[('company_id', 'in', company_ids)]</field>
```

contract/security/contract_security.xml
```diff
@@ -0,0 +1,46 @@
...
+    </record>
+    <record id="rule_contract_template_line_multi_company" model="ir.rule">
+        <field name="name">Contract template line multi-company</field>
+        <field name="model_id" ref="model_contract_template_line" />
+        <field name="global" eval="True" />
+        <field name="domain_force">[('company_id', 'in', company_ids + [False])]</field>
```
carlos-lopez-tecnativa:
```suggestion
        <field name="domain_force">[('company_id', 'in', company_ids)]</field>
```

carlos-lopez-tecnativa (changes requested):

carlos-lopez-tecnativa:
Additionally, I see many translation commits for the same language and the same author. For example, I think these commits can be squashed. Please reduce the number of administrative commits.

contract/models/contract.py
bosd: Good catch!

contract/security/contract_security.xml
BhaveshHeliconia: You're right, since company_id is a required field, checking for False is redundant. I've simplified the rules in contract_security.xml by removing it as suggested.

contract/models/contract.py
BhaveshHeliconia: Good point. I've updated the template onchange to set line_recurrence to True by default. This ensures the lines are visible as soon as the template is applied, which avoids the confusion you mentioned. I also went ahead and added the field to the template form view since it was missing there.

contract/models/account_move.py
BhaveshHeliconia: Actually, this field is still needed for backward compatibility. It's used in the _get_related_invoices method to find invoices created in version 11 or older. If we remove it, we’d lose the ability to track those historical records.

BhaveshHeliconia:
@carlos-lopez-tecnativa thanks for the review. I’ve addressed your comments.

bosd:
Hello @sbidoul , @AaronHForgeFlow , sorry for pinging you directly, can you pls trigger merge here?

MohamedOsman7 (approved):
Tested with runboat, LGTM

contract/models/contract.py
```diff
@@ -452,7 +452,9 @@ def _get_related_invoices(self):
         )
         # we are forced to always search for this for not losing possible <=v11
         # generated invoices
-        invoices |= self.env["account.move"].search([("old_contract_id", "=", self.id)])
+        invoices |= self.env["account.move"].search(
+            [("old_contract_id", "=", self.ids)]
+        )
```
carlos-lopez-tecnativa: I’m not sure if this issue is caused by this change, but please check it.

In v19, when you create a new contract without a customer, the invoice count is incorrect. This causes the `Recurrence at line level field` to become readonly because `invoice_count != 0`.
Image

In v18, this works fine.
Image

carlos-lopez-tecnativa: Additionally, if you pass a list of IDs, the operator should be `in` instead of `=`. Please clarify the reason for this change.

carlos-lopez-tecnativa (changes requested):
In my previous comment, I suggested squashing the administrative commits for the same language and the same author, but I noticed that you squashed commits that should not have been squashed and merged them into the previous commit, losing the original author. Please review this and rework the Git history correctly.

You can read this documentation if it helps:
https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests

contract/models/contract.py
BhaveshHeliconia: Done. Changed to in to correctly handle NewId in Odoo 19. This fixes the incorrect count for new contracts and keeps the recurrence field editable. Thanks!

I squashed all translation-related commits together, but I understand now that some commits should have remained separate to preserve the original authorship. I'll rework the Git history accordingly and add it to my TODO list.

pedrobaeza: Yes, you should only merge translation commits coming from the same author, not all the translations commits together.

BhaveshHeliconia: @carlos-lopez-tecnativa I’ve reworked the commit history properly now, preserving the original authors where needed.

MohamedOsman7:
@carlos-lopez-tecnativa can you approve?

contract/models/contract.py
BhaveshHeliconia: ping @carlos-lopez-tecnativa

contract/models/account_move.py
carlos-lopez-tecnativa: Sorry, but this field was removed in this PR: https://github.com/OCA/contract/pull/1330

However, this commit looks strange. Please check whether you mixed multiple commits into a single commit:
https://github.com/OCA/contract/pull/1312/changes/501d3870bcb8b400572a8c0e6f066a00478d34b9#diff-bd92412e26bfd77a1e9fb18d4a493b62b8a9c1ec64cbf57a807f244f12c198fc
image


And in the migration-to-19 commit: https://github.com/OCA/contract/pull/1312/changes#diff-1a8a5c5a17e2198f9684ca9df981f176556dfe610ef4b983c80b99cc679356d9 you added this code again, even though it had already been removed. Please check this, because from my point of view the git history is broken.
image

contract/models/contract.py
carlos-lopez-tecnativa: Thanks, but in that case, I don’t see the need to add the field to the contract template form. If the field is unused in the template because **True is always passed when selecting the contract template**, then displaying it does not seem necessary.

carlos-lopez-tecnativa (changes requested):
Please include https://github.com/OCA/contract/pull/1435

I’m not sure whether this PR should be included here https://github.com/OCA/contract/pull/1429

It was merged into v18, but it is unused in this module. Based on the PR description, it will be used in other modules.

contract/models/account_move.py
BhaveshHeliconia: I've checked it, and you are right — that field was already removed in a previous commit. I've now removed it from the migration commit as well. Thanks for catching this.

BhaveshHeliconia:
@carlos-lopez-tecnativa I've cherry-picked this https://github.com/OCA/contract/pull/1435 as per your suggestion. Thank you!

carlos-lopez-tecnativa (changes requested):
When I go to the contract lines pivot view, I get this error.
Image

Please check this commit: https://github.com/OCA/contract/pull/1312/changes/501d3870bcb8b400572a8c0e6f066a00478d34b9 It seems to be mixing code changes with translation changes.

BhaveshHeliconia:
@carlos-lopez-tecnativa I've solved that issue. Thanks!

bosd:
@carlos-lopez-tecnativa Can we move forward with this one?

carlos-lopez-tecnativa (approved):
@BhaveshHeliconia Please check my comment about this commit: https://github.com/OCA/contract/commit/501d3870bcb8b400572a8c0e6f066a00478d34b9 It seems to be mixing code changes with translation changes.

Everything else looks good. If @pedrobaeza agrees, it can be merged as far as I'm concerned. However, I think the commit mentioned above should be reworked and split into separate commits.

pedrobaeza:
Yes, please, let's fix that commit.

BhaveshHeliconia:
@carlos-lopez-tecnativa @pedrobaeza I've separated the commit. Could you please check whether it looks correct now?

If there are still any improvements needed, please let me know and I'll address them.

pedrobaeza:
Let's merge it, although the separation is not totally clean, but it's not breaking any real attribution.

/ocabot merge nobump

PR #1315, opened 2025-10-07
[18.0][IMP] product_contract - remove unnecessary defaults

anthonissen-a:
The default attributes are no longer necessary since the fields are set by the compute method with a precompute attribute.

### DIFF
--- a/product_contract/models/sale_order_line_contract_mixin.py
+++ b/product_contract/models/sale_order_line_contract_mixin.py
@@ -36,7 +36,6 @@ class SaleOrderLineContractMixin(models.AbstractModel):
             ("semesterly", "Semester(s)"),
             ("yearly", "Year(s)"),
         ],
-        default="monthly",
         help="Specify Interval for contract duration.",
         compute="_compute_product_contract_data",
         precompute=True,
@@ -44,7 +43,6 @@ class SaleOrderLineContractMixin(models.AbstractModel):
         readonly=False,
     )
     recurring_interval = fields.Integer(
-        default=1,
         string="Invoice Every",
         help="Invoice every (Days/Week/Month/Year)",
         compute="_compute_product_contract_data",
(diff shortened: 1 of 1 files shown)


yvaucher:
:+1: 

By the way according to odoo official doc, the default value disables the precompute.

image

yvaucher (approved):

mamcode (approved):

pedrobaeza (changes requested):
Looking at the code, we are not going to have the proper value in the line init, as the compute is not assigning any default value.

anthonissen-a:
Hi @pedrobaeza, thanks for your review. 
Since the fields have `precompute=True`, the compute function will be triggered at line init. As @yvaucher mentioned, the tags `default` and `precompute` can't be used together because they have conflicting behaviors.

pedrobaeza:
I understood the contrary: the default is applied instead of the compute.

anthonissen-a:
Exactly, and therefore the `default` should be removed, to allow a proper use of the `precompute`.

pedrobaeza:
But that can't be done if the compute is not providing the same default value, which is not the case.

yvaucher:
@pedrobaeza Please read again the offical doc warning (see my post upper).

`default` and `precompute` are incompatible, thus one must go.

Are you saying that we should move the default values in the "init" values of the compute?

For instance `'recurring_interval: 1`

It could make sense even for the compute acting as an onchange on the view that if you change the product to a non contract product. As you want plausible default values.

pedrobaeza:
OK to renew the compute method

anthonissen-a:
Do you have a case where these values are used in another context than with products of type contract ? Otherwise, the values of these fields will be overwritten by the compute when a product of type contract is set.

marcos-mendez:
Not just a cleanup — these defaults silently break contract creation from a sale order. On 18.0 a product configured as `yearly` produces an order line with `recurring_rule_type = monthly`, so an annual plan bought on the website generates a monthly contract at the annual price.

`recurrence_number` is the only one of these fields without a `default=`, and the only one that propagates today. Confirmed this PR fixes it end to end.

anthonissen-a:
@pedrobaeza Could you check the 2 last comments and re-evaluate your change request please ?

rousseldenis (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #1318, opened 2025-10-17
[18.0][IMP] product_contract - add class o_row to auto_renew fields

anthonissen-a:
Added `class="o_row"` to the auto-renew divs to ensure consistent layout and styling with the recurrence section.

### DIFF
--- a/product_contract/views/product_template.xml
+++ b/product_contract/views/product_template.xml
@@ -69,7 +69,7 @@
                             string="Renewal Frequency"
                         >
                             <label for="auto_renew_interval" />
-                            <div>
+                            <div class="o_row">
                                 <field
                                     name="auto_renew_interval"
                                     class="oe_inline"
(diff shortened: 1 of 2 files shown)


mamcode (approved):
Thanks for this improvement @anthonissen-a 
At the code level, I understand that this improvement adds consistency in layout and styling with the recurrence section, but visually I can't see any changes. In any case, I believe this is the correct way the code should be.

anthonissen-a:
Hi @mamcode, thanks for the review! In one of our projects, I had to make these fields readonly, and that's where the styling difference becomes noticeable. You can see the contrast in the screenshots below, with and without `class="o_row"` applied : 

image
image

mamcode:
@anthonissen-a Thanks, now it's clear to me.

yvaucher (approved):
Seems consistant with the rest of the module. (I wonder what happens on the responsivness though)

bofiltd (approved):

PR #1321, opened 2025-10-23
[18.0][FIX] contract: Enforce company on tests

etobella:
We should add a validation of company as it might collide if we have multiple companies installed and this companies have a sequence.

One example would be to test contract and l10n_es in the same instance.


carlosdauden (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #1323, opened 2025-10-24
[18.0][IMP] contract_line_successor: allow importing successor_contract_lin…

Rad0van:
…e_id and predecessor_contract_line_id

As outlined in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0 (which I believe applies to 18.0 as well) changing these from readonly=True to be able to  import.


Rad0van:
No comments on this? @sbejaoui ?

PR #1324, opened 2025-10-24
[17.0][IMP] contract_sale_invoicing: Add option to include sale lines in th…

carlos-lopez-tecnativa:
Before this commit, when a sale order was invoiced from a contract, it created a new invoice, resulting in two invoices, one for the contract and another for the sale order. After this commit, when this new field is checked in the contract, the sale order lines are invoiced in the same invoice as the contract.

TT58137
@Tecnativa @pedrobaeza could you please review this?

### DIFF
--- a/contract_sale_invoicing/models/contract.py
+++ b/contract_sale_invoicing/models/contract.py
@@ -12,9 +12,17 @@ class ContractContract(models.Model):
         help="If checked include sales with same analytic account to invoice "
         "in contract invoice creation.",
     )
+    invoicing_sales_into_contract = fields.Boolean(
+        help="If checked, include the sales order lines "
+        "in the contract invoice instead of creating separate invoices.",
+    )
 
     def _recurring_create_invoice(self, date_ref=False):
         invoices = super()._recurring_create_invoice(date_ref)
+        invoice_by_contract = {}
+        for invoice in invoices:
+            contract = invoice.invoice_line_ids.contract_line_id.contract_id
+            invoice_by_contract[contract.id] = invoice
         for contract in self:
             if not contract.invoicing_sales or not contract.recurring_next_date:
                 continue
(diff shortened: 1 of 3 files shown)


contract_sale_invoicing/models/contract.py
```diff
@@ -12,44 +12,78 @@ class ContractContract(models.Model):
         help="If checked include sales with same analytic account to invoice "
         "in contract invoice creation.",
     )
+    invoice_into_contract = fields.Boolean(
```
pedrobaeza: For having the same field prefix to ease recognition:

```suggestion
    invoicing_sales_into_contract = fields.Boolean(
```

contract_sale_invoicing/models/contract.py
```diff
@@ -12,44 +12,78 @@ class ContractContract(models.Model):
...
-                ]
-            )
-            sales |= self.env["sale.order"].search(
-                common_domain
-                + [
-                    (
-                        "order_line.distribution_analytic_account_ids",
-                        "in",
-                        self.group_id.ids,
+            sales = contract._get_sales_to_invoice()
```
pedrobaeza: Why extracting this to a method if it's only called once? Avoid diffs that make the PR review more difficult, or at least do it in a separate `[REF]` commit.

contract_sale_invoicing/tests/test_contract_sale_invoicing.py
```diff
@@ -77,11 +89,30 @@ def test_not_sale_invoicing(self):
...
+        Do invoice Sale Order that matches on analytic account
+        this must add the sale order lines into the contract invoice
+        """
+        self.contract.invoicing_sales = True
+        self.contract.invoice_into_contract = True
+        self.sale_order.action_confirm()
+        self.contract.recurring_create_invoice()
+        self.assertEqual(self.sale_order.invoice_status, "invoiced")
+        self.assertEqual(
+            self.sale_order.invoice_ids, self.contract._get_related_invoices()
```
pedrobaeza: I think you should test the number of invoice lines, and remember to put fixed numbers, not calling other methods.

pedrobaeza (changes requested):

contract_sale_invoicing/models/contract.py
carlos-lopez-tecnativa: Done

carlos-lopez-tecnativa:
ping @pedrobaeza

pedrobaeza (approved):

christian-ramos-tecnativa (approved):
LGTM!

pedrobaeza:
/ocabot merge minor

Please fw-port it to 18

PR #1329, opened 2025-10-29
[16.0][ADD] subscription_oca_recurrent_payment

adasatorres:
This addon adds an option to subscription templates so that a recurring payment is made using the customer's latest payment_token_id.


chrisandrewmann:
@adasatorres thanks for your contribution. 
I had been working on the same feature, but for V18 and have my own working prototype with the code directly inside subscription_oca

I feel that it should not be a seperate module but should be part of subscription_oca as it's a critical feature which is not really optional. I'd say we should avoid what has happened with the contract OCA module, which has "dependency hell" with modules for every possible minor feature making it hard to maintain.

adasatorres:
@chrisandrewmann It seems like a good idea to add this functionality to the base addon; it will be easier to maintain a single addon rather than several. I’m closing this PR and will open another one once I have the implementation in the base addon.

chrisandrewmann:
> @chrisandrewmann It seems like a good idea to add this functionality to the base addon; it will be easier to maintain a single addon rather than several. I’m closing this PR and will open another one once I have the implementation in the base addon.

Thanks @adasatorres that's great. If you want I can provide you a copy of my own code for reference (V18) to see how it compares to your own? No point me submitting my own PR.

adasatorres:
> Thanks @adasatorres that's great. If you want I can provide you a copy of my own code for reference (V18) to see how it compares to your own? No point me submitting my own PR.

Yes, perfect, that way I’ll have another perspective @chrisandrewmann .

chrisandrewmann:
@adasatorres I've sent you a LinkedIn connection request. Let's chat on there and sort out how best to send code

PR #1330, opened 2025-10-30
[18.0][IMP] contract: Remove account.move old_contract_id field

qgroulard:
This field is often searched which may be slow on DB with a lot of account moves.


flotho (approved):
Code review,
LGTM
Great thanks

FrancoMaxime (approved):
lgtm

pedrobaeza:
There's only one search:

https://github.com/OCA/contract/blob/1df7f34a14fc23e5deef03eed84d316f2292f5ed/contract/models/contract.py#L426

and that one is only for retro-compatibility with 11-. I'm wondering if we should still provide that compatibility...

qgroulard:
> There's only one search:
> 
> https://github.com/OCA/contract/blob/1df7f34a14fc23e5deef03eed84d316f2292f5ed/contract/models/contract.py#L426
> 
> and that one is only for retro-compatibility with 11-. I'm wondering if we should still provide that compatibility...

I agree we should consider removing this feature.

pedrobaeza:
Go ahead then.

etobella (approved):

pedrobaeza:
The field hasn't been removed as talked.

qgroulard:
That's still on my todo list but the go-live of a major customer in a few days leaves me little time.

qgroulard:
@pedrobaeza Better late than never, sorry for the delay.

pedrobaeza (approved):

PR #1341, opened 2025-11-18
[18.0][IMP] contract: hook for line is invoiceable

tobiaszehntner:
Make can_be_invoiced available on contract_line, so other modules can modify conditions.

PR #1342, opened 2025-11-18
[18.0][ADD] contract_line_defer: Avoid invoicing line until activated

tobiaszehntner:
Defer Contract Lines to avoid invoicing while start date unknown

Enable to defer contract line's invoicing. If a contract line start date is not known, it can be deferred. It will not be invoiced until manually activated.

- Based on https://github.com/OCA/contract/pull/1341
- contract_line_defer: module to defer contract lines
- product_contract_line_defer: bridge module for product_contract
- contract_line_defer_successor: bridge module for contract_line_successor


contract_line_defer_successor/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+    "author": "ACSONE SA/NV,Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/contract",
+    "depends": [
+        "contract_line_defer",
+        "contract_line_successor",
+    ],
+    "data": [
+        "views/contract_line.xml",
+    ],
+    "auto_install": True,
```
acsonefho, line 18: Make sense for OCA module?

contract_line_defer/wizards/contract_line_activate_wizard.py
```diff
@@ -0,0 +1,27 @@
...
+    )
+    date_end = fields.Date(
+        help="Optional contract termination date (limits recurrence)"
+    )
+
+    def activate_line(self):
+        self.ensure_one()
+        self.contract_line_id.is_deferred = False
+        self.contract_line_id.date_start = self.date_start
+        self.contract_line_id.date_end = self.date_end
```
acsonefho: Can you do this update in a single `write(...)` :pray: 
If another custom module do a `flush` after each `write`, you probably will trigger it 3 times instead of once.
IMO a good practice is to keep a single `write` when it's possible.

product_contract_line_defer/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+        "contract_line_defer",
+    ],
+    "data": [
+        "views/contract_line.xml",
+        "views/product_template.xml",
+        "views/sale_order.xml",
+        "wizards/product_contract_configurator.xml",
+    ],
+    "assets": {"web.assets_backend": ["product_contract_line_defer/static/src/js/*"]},
+    "auto_install": True,
```
acsonefho, line 22: Same question?

product_contract_line_defer/models/sale_order_line.py
```diff
@@ -0,0 +1,45 @@
...
+        values["is_deferred"] = self.is_deferred
+        return values
+
+    @api.depends(
+        "is_deferred",
+    )
+    def _compute_name(self):
+        res = super()._compute_name()
+        for line in self:
+            if line.is_contract:
```
acsonefho: Could be a single `if`.

product_contract_line_defer/models/sale_order_line_contract_mixin.py
```diff
@@ -0,0 +1,49 @@
...
+            if company.defer_contract_line_start:
+                result["is_deferred"] = True
+        return result
+
+    @api.depends("product_id")
+    def _compute_product_contract_data(self):
+        res = super()._compute_product_contract_data()
+        for rec in self:
+            if rec.product_id.is_contract and rec.product_id.is_deferred:
+                rec.update({"is_deferred": True})
```
acsonefho: You can do the update in one shot:
`self.filtered(lambda rec: rec.product_id.is_contract and rec.product_id.is_deferred).update({"is_deferred": True})`

acsonefho (changes requested):
Only minor stuff!
Really good job :muscle:

contract_line_defer_successor/__manifest__.py
tobiaszehntner: Makes sense for a bridge module. If both `contract_line_defer` and `contract_line_successor` are installed, this module *should* be installed to make sure everything works properly.

From [Odoo docs](https://www.odoo.com/documentation/18.0/developer/reference/backend/module.html):

> auto_install (bool or list(str), default: False)
> If True, this module will automatically be installed if all of its dependencies are installed.
> 
> It is generally used for “link modules” implementing synergetic integration between two otherwise independent modules.

product_contract_line_defer/__manifest__.py
tobiaszehntner: Same as above (for bridge modules)

acsonefho (approved):

bosd:
@tobiaszehntner  Can you please fix the tests?

tobiaszehntner:
> @tobiaszehntner Can you please fix the tests?

rebased branch which fixed the tests. Added extra tests for improving coverage.

acsonefho (approved):

contract_line_defer/models/contract_line.py
```diff
@@ -0,0 +1,44 @@
...
+
+    @api.onchange("is_deferred")
+    def _onchange_is_deferred(self):
+        """
+        If activating a contract line, make sure start/end dates are compatible
+        """
+        self.ensure_one()
+        if not self.is_deferred:
+            today = fields.date.today()
+            if self.date_start < today:
```
bosd: A deferred line typically has no `date_start` (that's the point of deferring — the start date is unknown). When the user un-checks **Deferred** on such a line, this onchange runs with `date_start = False`, and `False < today` raises `TypeError: '<' not supported between instances of 'bool' and 'datetime.date'`.

```suggestion
            if self.date_start and self.date_start < today:
```

The existing tests don't catch this because `test_contract_line_deferred_disable` sets `date_start` before toggling — worth adding a case where `date_start` is empty.

tobiaszehntner: Thanks! didn't expect that, date_start is a required field and we don't set it to False to not break other functionalities and modules inheriting contract. But I was indeed able to reproduce this when opening a line and unchecking deferred. I added a test catching this and fixed the onchange.

bosd (approved):
LGTM

bosd:
@gurneyalex Can you please merge?

qgroulard (approved):
LGTM (code review)

victor-champonnois (commented):
Functional test OK. Thank you for the great readme. 

Just two non-blocking comments :
- I note that the module contract_service_dates is not affected by the defered field. It could be nice to have a glue module at some point. 
- Also, the field "Next Period Start" is still filled and visible even the line is deferred. This could generate confusion.

victor-champonnois (approved):

tobiaszehntner:
> * I note that the module contract_service_dates is not affected by the defered field. It could be nice to have a glue module at some point.
> * Also, the field "Next Period Start" is still filled and visible even the line is deferred. This could generate confusion.

@victor-champonnois Thanks! I fixed the next period start (now False if deferred) and added a point to the readme/Roadmap regarding `contract_service_dates` module.

contract_line_defer/models/contract_line.py
```diff
@@ -0,0 +1,44 @@
...
+    @api.onchange("is_deferred")
+    def _onchange_is_deferred(self):
+        """
+        If activating a contract line, make sure start/end dates are compatible
+        """
+        self.ensure_one()
+        if not self.is_deferred:
+            today = fields.date.today()
+            if not self.date_start or self.date_start < today:
+                self.date_start = today
```
rousseldenis: @tobiaszehntner Contract line `date_start` and `date_end` are computed. Don't use onchanges.

tobiaszehntner: fixed for date_start and left for date_end as it's not computed on the contract mixin

tobiaszehntner: @rousseldenis hah that doesn't actually work. The onchange was for the single purpose when the user de-selects is_deferred in the view. but if I move it to the view it triggers on any contract.line that has is_deferred False. I reverted to onchange
```
2026-09-22 09:36:11,339 337 ERROR odoo odoo.addons.contract_sale_generation.tests.test_contract_sale: FAIL: TestContractSale.test_contract
Traceback (most recent call last):
  File "/__w/contract/contract/contract_sale_generation/tests/test_contract_sale.py", line 25, in test_contract
    self.assertEqual(self.contract_line.recurring_next_date, recurring_next_date)
AssertionError: datetime.date(2026, 10, 22) != datetime.date(2020, 2, 15)
 
2026-09-22 09:36:11,492 337 ERROR odoo odoo.addons.contract_sale_generation.tests.test_contract_sale: FAIL: TestContractSale.test_contract_autoconfirm
Traceback (most recent call last):
  File "/__w/contract/contract/contract_sale_generation/tests/test_contract_sale.py", line 40, in test_contract_autoconfirm
    self.assertEqual(self.contract_line.recurring_next_date, recurring_next_date)
AssertionError: datetime.date(2026, 10, 22) != datetime.date(2020, 2, 15)
```

PR #1343, opened 2025-11-19
[18.0][FIX] contract: automatic pricing should resptect pricelist quantity-based rules

RicardCForgeFlow:
__get_product_price_ was sent with quantity=1 instead of the actual quantity of the line. This was causing incorrect prices when a pricelist has quantity-based rules

See the example added in the test

### DIFF
--- a/contract/models/contract_template_line.py
+++ b/contract/models/contract_template_line.py
@@ -194,8 +194,9 @@ class ContractTemplateLine(models.Model):
                         line.contract_id.company_id
                     ).property_product_pricelist
                 )
+                qty = line.env.context.get("contract_line_qty", line.quantity)
                 product = line.product_id.with_context(
-                    quantity=line.env.context.get("contract_line_qty", line.quantity),
+                    quantity=qty,
                     pricelist=pricelist.id,
                     partner=line.contract_id.partner_id.id,
                     uom=line.uom_id.id,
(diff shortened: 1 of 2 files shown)


contract/models/contract_template_line.py
```diff
@@ -194,16 +194,17 @@ def _compute_price_unit(self):
                         line.contract_id.company_id
                     ).property_product_pricelist
                 )
+                qty = line.quantity
```
MiquelRForgeFlow:
```suggestion
                qty = line.env.context.get("contract_line_qty", line.quantity)
```

RicardCForgeFlow: Okay!

AaronHForgeFlow (approved):
Code review LGTM

MiquelRForgeFlow (approved):

contract/tests/test_contract.py
```diff
@@ -244,6 +262,34 @@ def test_automatic_price(self):
...
+                "Price should be 100 for quantity 1 based on pricelist rule",
+            )
+            self.acct_line.quantity = 10
+            self.assertEqual(
+                self.acct_line.price_unit,
+                50,
+                "Price should be 50 for quantity 10 based on pricelist rule",
+            )
+        finally:
+            # Restore original values
```
pedrobaeza: Why are you doing this try/finally? These are tests, and after each method call, things are restored automatically.

RicardCForgeFlow: Oops! My bad. I have now deleted the try/finally

pedrobaeza (approved):
/ocabot merge patch

PR #1344, opened 2025-11-19
[18.0][IMP] product_contract: Allow creating sales orders with lines that lack a contract template

carlos-lopez-tecnativa:
Before this commit, confirming a sales order would raise an exception if any line lacked a configured contract template, preventing the creation of the contract.

Now, sales order lines without a contract template are automatically added to a new contract without a template, allowing the process to continue seamlessly.

FWP of https://github.com/OCA/contract/pull/1185

TT54544
@Tecnativa @pedrobaeza @CarlosRoca13 @sergio-teruel  could you please, review this

### DIFF
--- a/product_contract/models/sale_order.py
+++ b/product_contract/models/sale_order.py
@@ -48,8 +48,11 @@ class SaleOrder(models.Model):
 
     def _prepare_contract_value(self, contract_template):
         self.ensure_one()
+        name = (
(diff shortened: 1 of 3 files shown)


pedrobaeza (approved):
Fast-tracking as direct fw-port:

/ocabot merge patch

PR #1313, opened 2025-10-02
[17.0][IMP] contract,contract_sale_invoicing: improve search for Sale Order Lines with analytic account

carlos-lopez-tecnativa:
Partial backport of https://github.com/OCA/contract/pull/1206

See each commit description for more information.

TT57029
@Tecnativa @pedrobaeza @victoralmau  @tobiaszehntner @sbejaoui @rousseldenis Could you please review this?

PR #1345, opened 2025-11-19
[18.0][IMP] contract: compute group_id based on analytic accounts in contract lines

carlos-lopez-tecnativa:
This field is unused in this module, but in the contract_sale_invoicing module it is used to search Sales Orders. However, many times this field goes unnoticed and users set the analytic distribution directly on the lines.

After this commit, the field is computed automatically based on the contract lines.

Partial FWP of https://github.com/OCA/contract/pull/1313

TT57029
@Tecnativa @pedrobaeza @victoralmau  @tobiaszehntner @sbejaoui @rousseldenis Could you please review this?

### DIFF
--- a/contract/models/contract.py
+++ b/contract/models/contract.py
@@ -42,6 +42,9 @@ class ContractContract(models.Model):
     group_id = fields.Many2one(
         string="Group",
         comodel_name="account.analytic.account",
+        compute="_compute_group_id",
+        store=True,
+        readonly=False,
         ondelete="restrict",
     )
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):
Yes, as working on 17.

victoralmau (approved):

FrancoMaxime:
@carlos-lopez-tecnativa @pedrobaeza Hello, is it not better to move the group_id field in the contract_sale_invoicing module especially since it is required but not used?

PR #1346, opened 2025-11-19
[18.0][IMP] contract_sale_invoicing: Add option to include sale lines in the same contract invoice

carlos-lopez-tecnativa:
FWP of https://github.com/OCA/contract/pull/1324
Before this commit, when a sale order was invoiced from a contract, it created a new invoice, resulting in two invoices, one for the contract and another for the sale order. After this commit, when this new field is checked in the contract, the sale order lines are invoiced in the same invoice as the contract.

TT58137
@Tecnativa @pedrobaeza @christian-ramos-tecnativa  could you please review this?


pedrobaeza (approved):
Same as in 17, so fast-tracking this:

/ocabot merge minor

PR #1347, opened 2025-11-19
[17.0][IMP] contract: enhance search filters and add a pivot view for contract lines

carlos-lopez-tecnativa:
- store additional fields for analysis purposes

TT59300
@Tecnativa @pedrobaeza @victoralmau could you please review this?

### DIFF
--- a/contract/models/contract_line.py
+++ b/contract/models/contract_line.py
@@ -32,6 +32,8 @@ class ContractLine(models.Model):
         auto_join=True,
         ondelete="cascade",
     )
+    # replace from abstract to add the store=True
+    partner_id = fields.Many2one(store=True)
     currency_id = fields.Many2one(related="contract_id.currency_id")
     date_start = fields.Date(required=True)
     date_end = fields.Date(compute="_compute_date_end", store=True, readonly=False)
--- a/contract/views/contract_line.xml
+++ b/contract/views/contract_line.xml
@@ -230,8 +230,16 @@
                 <field name="recurring_next_date" />
                 <field name="last_date_invoiced" />
                 <separator />
-                <filter name="is_canceled" domain="[('is_canceled', '=', True)]" />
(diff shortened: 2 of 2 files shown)


contract/models/abstract_contract_line.py
```diff
@@ -49,6 +49,7 @@ class ContractAbstractContractLine(models.AbstractModel):
     price_subtotal = fields.Monetary(
         compute="_compute_price_subtotal",
         string="Sub Total",
+        store=True,
```
pedrobaeza: This is going to provoke a huge computation on existing DBs. Please add migration script for computing this by SQL.

carlos-lopez-tecnativa: I reverted the subtotal field to non-stored because its computation depends on the price unit, which itself depends on the pricelist, and computing this in SQL would be too complex.
In the pivot view, I only added the quantity and specific_price fields.
Please let me know if this is okay for you now.

contract/migrations/17.0.1.4.7/pre-migration.py
```diff
@@ -0,0 +1,19 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    _new_columns = [("contract.line", "partner_id", "many2one")]
+    openupgrade.add_columns(env, _new_columns)
+    openupgrade.logged_query(
```
pedrobaeza: You don't need this, as related fields with only one dot are automatically populated by SQL in the ORM framework.

pedrobaeza: And please revert the changes in the manifest.

carlos-lopez-tecnativa: Done

pedrobaeza (approved):

victoralmau (approved):

PR #1353, opened 2025-11-29
[MIG] subscription_oca: Migration to 19.0

yanchirino:


yanchirino:
Could someone help me understand what the issue is? From what I can tell, tests in v19 require an additional parameter to be initialized. What should I do in this case?
@OCA/project-service-maintainers

danielkauffman:
Regarding the failing checks:

> ValueError: External ID not found in the system: base.demo_user0

See #1312 which creates a user for testing:

        user_portal = self._create_new_portal_user(
            partner_id=partner.id,
            login="portal",
        )

Additionally those tests once refer to `cls.env.user`.

The above works for the tests in the contract module but I've not reviewed how to create user(s) for the particular tests in this module.

Note per ChatGPT:

> Starting progressively in 16–18 and fully in Odoo 19:
> Demo users are no longer defined as fixed XML records like base.demo_user0.
> Demo users are now generated programmatically during demo database initialization.

danielkauffman:
Same applies to products for testing. See https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-19.0

> If your tests rely on demo data, now they are not installed by default, as in fact, Odoo advises not to rely on them, so it's the occasion to create the data on the tests themselves.

PR #1359, opened 2025-12-06
[17.0][FIX] contract: Visual hint for cancelled lines in portal

pedrobaeza:
On portal, the cancelled lines are not distinguished from running ones, as the next invoice date is still displayed, and the color is the same (on contrary than the backend, that appears greyed).

So let's do something similar in portal:

- Put the row as grey.
- Hide the nexte invoice date on that case.

Before:

image

After:

image

@Tecnativa TT59704

### DIFF
--- a/contract/views/contract_portal_templates.xml
+++ b/contract/views/contract_portal_templates.xml
@@ -241,7 +241,7 @@
                                         t-as="line"
                                     >
                                         <tr
-                                            t-att-class="'bg-200 font-weight-bold o_line_section' if line.display_type == 'line_section' else 'font-italic o_line_note' if line.display_type == 'line_note' else ''"
+                                            t-att-class="'bg-200 font-weight-bold o_line_section' if line.display_type == 'line_section' else 'font-italic o_line_note' if line.display_type == 'line_note' else 'text-muted' if line.is_canceled else ''"
                                         >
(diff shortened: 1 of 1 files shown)


contract/views/contract_portal_templates.xml
```diff
@@ -241,7 +241,7 @@
                                         t-as="line"
                                     >
                                         <tr
-                                            t-att-class="'bg-200 font-weight-bold o_line_section' if line.display_type == 'line_section' else 'font-italic o_line_note' if line.display_type == 'line_note' else ''"
+                                            t-att-class="'bg-200 font-weight-bold o_line_section' if line.display_type == 'line_section' else 'font-italic o_line_note' if line.display_type == 'line_note' else 'text-secondary' if line.is_canceled else ''"
```
pilarvargas-tecnativa: Just a non-blocking comment, the primary and secondary colors can be changed, so it may not be gray. If you want it to always appear in gray, it is better to use .text-muted.

pilarvargas-tecnativa (approved):

contract/views/contract_portal_templates.xml
pedrobaeza: Changed

victoralmau (approved):

PR #1364, opened 2025-12-29
[19.0][MIG] contract_variable_quantity

bosd:
Standard migration using odoo-module-migrator 
/ocabot migration contract_variable_quantity .
depends on #1312


contract_variable_quantity/models/contract_line.py
```diff
@@ -0,0 +1,54 @@
...
+        vals = super()._prepare_invoice_line()
+        if (
+            "quantity" in vals
+            and self.contract_id.skip_zero_qty
+            and float_is_zero(
+                vals["quantity"],
+                self.env["decimal.precision"].precision_get("Product Unit of Measure"),
+            )
+        ):
+            vals = {}
```
carlos-lopez-tecnativa, line 54: Please take sections and notes into account. These lines should also be included, even though their quantity is zero.

When I enable this option, those lines are currently not included in the invoice.

**Contract**
Image

**Invoice**
Image

carlos-lopez-tecnativa (changes requested):
Please rebase the branch and remove the last commit, as the dependency has already been merged.

TT58613

contract_variable_quantity/models/contract_line.py
bosd: Good catch — fixed. The `skip_zero_qty` guard now excludes `display_type` lines, so sections and notes (which have no quantity) are always kept on the invoice. Added a test (`test_skip_zero_qty_keeps_sections_and_notes`) covering it.

bosd:
Thanks @carlos-lopez-tecnativa! Done:
- Rebased onto current `19.0` and dropped the trailing `[TEST REQUIREMENTS]` commit (dependency is merged now).
- Fixed the sections/notes issue you flagged — `_prepare_invoice_line` no longer drops `display_type` lines when *Skip zero-quantity lines* is enabled, with a test to cover it.

carlos-lopez-tecnativa (approved):
LGTM
@pedrobaeza @victoralmau could you please review this?

pedrobaeza (approved):
/ocabot merge nobump

PR #1365, opened 2025-12-29
[19.0][MIG] contract_variable_qty_prorated

bosd:
Standard migration using odoo-module-migrator 
/ocabot migration contract_variable_qty_prorated .

Depends on #1364
#1312

PR #1366, opened 2025-12-29
[19.0][ADD] contract_invoice_offset

bosd:
depends on #1372 and #1365

(#1312 and #1364 are merged.)


contract_invoice_offset/models/contract_recurring_mixin.py
```diff
@@ -0,0 +1,147 @@
...
+    invoicing_offset_type = fields.Selection(
+        [
+            ("days", "Days"),
+            ("weeks", "Weeks"),
+            ("months", "Months"),
+            ("years", "Years"),
+        ],
+        default="days",
+        required=True,
+    )
```
anthonissen-a, line 21: IMO, using "daily", "weekly", ..., would fit better with the naming patterns used elsewhere in `contract`. It would allow usage of helper methods like `get_relative_delta`.

contract_invoice_offset/models/contract.py
```diff
@@ -0,0 +1,87 @@
...
+        "recurring_rule_type",
+        "recurring_interval",
+        "date_end",
+        "recurring_next_date",
+        "invoicing_offset_type",
+        "invoicing_offset_value",
+    )
+    def _compute_next_period_date_end(self):
+        """Compute the end date of the next billing period."""
+        # Override to pass offset settings
```
anthonissen-a: typo: overwrite

contract_invoice_offset/__manifest__.py
```diff
@@ -0,0 +1,20 @@
...
+    "name": "Contract Invoice Offset",
+    "summary": "Flexible invoicing offsets (e.g. 1 month in advance)",
+    "version": "19.0.1.0.0",
+    "category": "Accounting",
+    "website": "https://github.com/OCA/contract",
+    "author": "bosd, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "depends": [
+        "contract",
+        "contract_invoice_align_start",
```
anthonissen-a, line 11: IMO this dependency should be removed to be more generic.

anthonissen-a (commented):
Very useful module! My main concern is compatibility with other modules, given the number of overwrites.

anthonissen-a:
Hey @bosd, I suggest some improvements in https://github.com/OCA/contract/pull/1498. Do you agree with my commits ?

contract_invoice_offset/__manifest__.py
bosd: Done — `contract_invoice_align_start` is off the `depends` list, and the `align_billing_cycle` plumbing (the `_get_offset_kwargs()` entry and the `api.depends` on it) is gone with it. `_get_offset_kwargs()` stays as the extension point, so a glue module can add the alignment back for installations that want both.

Side effect: this was the last unmerged dependency in the chain (it pulled in `contract_variable_qty_prorated` too), so the `[DO NOT MERGE]` commit with the `test-requirements.txt` pins is dropped and the PR only needs `contract` now.

contract_invoice_offset/models/contract_recurring_mixin.py
bosd: Agreed — this already landed in the "review feedback" commit: the selection is now `daily`/`weekly`/`monthly`/`yearly`, matching `recurring_rule_type`, and the hand-written if/elif chain is replaced by `get_relative_delta()`.

contract_invoice_offset/__manifest__.py
bosd: Reopening this one — I tried it and had to back it out.

Dropping `contract_invoice_align_start` from `depends` is not just a manifest change: `align_billing_cycle` is that module the offset chain currently rides on. Without it installed, the contract falls back to `min(line.recurring_next_date)` and the line-level pre-paid/post-paid values win over the contract-level ones, so `test_postpaid_delayed_one_month` and `test_days_fallback` both shift by a full period. The offset behaviour is deliberately crafted here, so I would rather not change it as a side effect of a dependency cleanup.

Happy to make it generic, but it needs to be its own piece of work: the alignment kwarg has to move out of `_get_offset_kwargs()` and into a glue module with tests covering both the with- and without-alignment cases. Do you want that as a follow-up PR, or folded into the refactor you are doing in #1498?

PR #1370, opened 2025-12-29
[19.0][MIG] contract_line_successor

bosd:
Standard migration using odoo-module-migrator /ocabot migration contract_line_successor .

PR #1367, opened 2025-12-29
[19.0][MIG] contract_termination

bosd:
Standard migration using odoo-module-migrator  .
contract_line_successor #1370
#1312

PR #1371, opened 2025-12-29
[19.0][MIG] product_contract

bosd:
Standard migration using odoo-module-migrator .

This version is severly cleaned according to:
https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests#mergesquash-the-commits-generated-by-bots-or-weblate

Total reduction from 202 to 105 commits (44% reduction)

Also lists the depends properly.

depends on:
#1312
#1367
#1370


pedrobaeza:
/ocabot migration product_contract

@bosd I see that you have generated `Merged bot/translation changes` commits, while the idea is to squash that ones into the real commit that generates them (i. e. `[IMP]`, `[FIX]` or `[MIG]` commits).

bosd:
> /ocabot migration product_contract
> 
> @bosd I see that you have generated `Merged bot/translation changes` commits, while the idea is to squash that ones into the real commit that generates them (i. e. `[IMP]`, `[FIX]` or `[MIG]` commits).

Yeah, I noticed that the custom script i used, created this..
(On this module it was a but to much to do manually). 
It's not optimal oca style, but decided to open this pr anyway, and see how this "lands".

pedrobaeza:
Uhm, I don't think it's OK, as now it's even more confusing about its source. I would prefer to let them as it was than this way.

bosd:
It is ok now

raulovallet:
Hey @bosd, I found this issue https://github.com/raulovallet/contract/commit/67bfd99c887fe4315b41c3f5672e686c49c05079. It causes the default values not to be set correctly and throws an error when closing the window. I’m not sure how I can contribute—could you make a cherry-pick or I have to do a pull request here?

wpichler:
Hi @bosd , I am facing an issue here with this module - the required="is_contract" on date_start seems to be wrong. In state draft,sent there must not be a date_start specified. Why is this required ?

wpichler:
Hi @bosd , Please ignore my last comment - i found the real issue...

I have been working on an migrated database (migrated from v16) - i had sale.order.line's with contract_start_date_method='manual' and date_start=null

For the record - i've fixed it with:

```sql
 UPDATE sale_order_line sol                                                                                                                                                                                                                                                                                                                                                                                                                                                                 
  SET date_start = COALESCE(                                                                                                                                                                                                                                                                                                                                                                                                                                                               
      (SELECT cl.date_start FROM contract_line cl                                                                                                                                                                                                                                                                                                                                                                                                                                            
        WHERE cl.sale_order_line_id = sol.id AND cl.date_start IS NOT NULL LIMIT 1),
      (SELECT so.date_order::date FROM sale_order so WHERE so.id = sol.order_id),                                                                                                                                                                                                                                                                                                                                                                                                            
      sol.create_date::date                                                                                                                                                                                                                                                                                                                                                                                                                                                                  
  )                                                                                                                                                                                                                                                                                                                                                                                                                                                                                          
  FROM product_product pp                                                                                                                                                                                                                                                                                                                                                                                                                                                                    
  JOIN product_template pt ON pt.id = pp.product_tmpl_id                                                                                                                                                                                                                                                                                                                                                                                                                                   
  WHERE sol.product_id = pp.id                                                                                                                                                                                                                                                                                                                                                                                                                                                               
    AND pt.is_contract IS TRUE
    AND sol.date_start IS NULL                                                                                                                                                                                                                                                                                                                                                                                                                                                               
    AND COALESCE(sol.contract_start_date_method, 'manual') = 'manual';
```

raulovallet:
Any update for this?

test-requirements.txt
```diff
@@ -0,0 +1 @@
+odoo-addon-contract_termination @ git+https://github.com/OCA/contract@refs/pull/1367/head#subdirectory=contract_termination
```
etobella, line 1: This should be removed from the migration commit 😉

bosd:
@tobiaszehntner an you please review?

product_contract/models/product_template.py
```diff
@@ -154,10 +154,21 @@ class ProductTemplate(models.Model):
 
     def write(self, vals):
         if "is_contract" in vals and vals["is_contract"] is False:
-            for company in self.env["res.company"].search([]):
-                self.with_company(company).write(
-                    {"property_contract_template_id": False}
+            # Get only the IDs to avoid loading all company records into memory
+            # Use a reasonable batch size to avoid performance issues
```
tobiaszehntner, line 158: I'd suggest a comment on why we have to do it in the first place
```
# property_contract_template_id is a company-dependent field and we have to remove the template for all companies that used this product

tobiaszehntner (approved):

PR #1377, opened 2026-01-12
[17.0][FIX] contract: Use another condition for forcing a failure

etobella:
Odoo forced a change that makes the test fail:

https://github.com/odoo/odoo/pull/240230

This probably needs to be ported to 18 and 19 :disappointed: however, the change is not there yet.


pedrobaeza (approved):
/ocabot merge patch

PR #1380, opened 2026-01-19
[17.0][IMP] product_contract: New feature to avoid create contract line from the marked line

CarlosRoca13:
With this changes we add a new way to avoid create a contract from the line marked with this new option. There are some cases where we want to invoice a contract product but we don't want to create that contract line.

cc @Tecnativa TT60386

ping @juancarlosonate-tecnativa @sergio-teruel

### DIFF
--- a/product_contract/i18n/es.po
+++ b/product_contract/i18n/es.po
@@ -9,8 +9,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Odoo Server 10.0\n"
 "Report-Msgid-Bugs-To: \n"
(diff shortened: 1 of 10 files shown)


luisDIXMIT (approved):
Code review and tested locally, LGTM!

As a suggestion, the ‘Create Contracts’ button on sales can be hidden if all lines are set to avoid creating contracts, as it doesn’t make sense to show it.

pedrobaeza:
/ocabot merge minor

@CarlosRoca13 please fw-port it

PR #1381, opened 2026-01-21
[18.0][IMP] contract: Improve contract lines view with section/note support

cristina-hidalgo-tecnativa:
This PR improves the contract form view:
- Apply section_and_note_text widget to name field
- Apply product_label_section_and_note_field widget to product_id

@Tecnativa TT60295
@christian-ramos-tecnativa

### DIFF
--- a/contract/views/contract.xml
+++ b/contract/views/contract.xml
@@ -105,7 +105,7 @@
                             <field
                                 name="contract_line_fixed_ids"
                                 invisible="line_recurrence"
-                                widget="section_and_note_one2many"
+                                widget="product_label_section_and_note_field_o2m"
                                 context="{'default_contract_type': contract_type, 'default_recurring_rule_type': recurring_rule_type, 'default_recurring_invoicing_type': recurring_invoicing_type, 'default_recurring_interval': recurring_interval, 'default_date_start': date_start, 'default_recurring_next_date': recurring_next_date}"
                             >
                                 <list
@@ -129,12 +129,20 @@
                                         column_invisible="True"
                                     />
                                     <field name="sequence" widget="handle" />
-                                    <field name="product_id" />
+                                    <field
+                                        name="product_id"
+                                        widget="product_label_section_and_note_field"
+                                    />
+                                    <field
+                                        name="name"
+                                        widget="section_and_note_text"
+                                        optional="show"
+                                    />
                                     <field
                                         name="product_uom_category_id"
                                         column_invisible="True"
                                     />
-                                    <field name="name" widget="section_and_note_text" />
+
                                     <field
                                         name="analytic_distribution"
                                         widget="analytic_distribution"
@@ -187,7 +195,7 @@
                             <field
                                 name="contract_line_ids"
                                 invisible="not line_recurrence"
-                                widget="section_and_note_one2many"
+                                widget="product_label_section_and_note_field_o2m"
                                 context="{'default_contract_type': contract_type}"
                             >
                                 <list decoration-info="create_invoice_visibility">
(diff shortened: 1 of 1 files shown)


contract/views/contract.xml
```diff
@@ -208,8 +228,28 @@
...
+                                        context="{
+                                            'partner_id': parent.partner_id,
+                                            'quantity': quantity,
+                                            'company_id': parent.company_id,
+                                            'pricelist_id': parent.pricelist_id,
+                                            'fiscal_position_id': parent.fiscal_position_id,
+                                            'currency_id': parent.currency_id,
+                                            'uom': uom_id
+                                        }"
+                                        domain="['|', ('sale_ok', '=', True), ('purchase_ok', '=', True)]"
```
christian-ramos-tecnativa: IMO you are adding too much extra stuff, Can we keep it simple? Or there is a reason for it? 
```suggestion
                                        widget="product_label_section_and_note_field"
```

christian-ramos-tecnativa (changes requested):

contract/views/contract.xml
cristina-hidalgo-tecnativa: All good

contract/views/contract.xml
```diff
@@ -129,12 +129,20 @@
...
                                     <field name="sequence" widget="handle" />
-                                    <field name="product_id" />
+                                    <field
+                                        name="product_id"
+                                        widget="product_label_section_and_note_field"
+                                    />
+                                    <field
+                                        name="name"
+                                        widget="section_and_note_text"
+                                        optional="show"
```
christian-ramos-tecnativa: You should also remove the optional in the name, the widget handle the display
```suggestion
```

contract/views/contract.xml
```diff
@@ -208,8 +216,16 @@
...
-                                    <field name="product_id" />
-                                    <field name="name" widget="section_and_note_text" />
+                                    <field
+                                        name="product_id"
+                                        widget="product_label_section_and_note_field"
+                                    />
+                                    <field
+                                        name="name"
+                                        widget="section_and_note_text"
+                                        optional="show"
```
christian-ramos-tecnativa: Same here
```suggestion
```

christian-ramos-tecnativa (changes requested):

contract/views/contract.xml
cristina-hidalgo-tecnativa: All good now

contract/views/contract.xml
cristina-hidalgo-tecnativa: All good now

christian-ramos-tecnativa (approved):

contract/views/contract.xml
carlos-lopez-tecnativa: > You should also remove the optional in the name, the widget handle the display

Sorry, but I do not agree with this. I prefer to keep the optional attribute, as in Sales.
The widget does not handle this. Imagine a product with a large description and you want to hide it; currently, this is impossible.

Sale
image

Contract
image

carlos-lopez-tecnativa (changes requested):
Please check my comment
https://github.com/OCA/contract/pull/1381#discussion_r2737016316

pedrobaeza:
There's something weird trying in runboat: now the list when you uncheck "Recurrence at line level?" is no more editable, while the idea was to have this "easy mode" for doing the edition directly.

christian-ramos-tecnativa:
I think is because when you set the "Recurrence at line level?" the line no longer is `editable bottom` and the form pop up for you to edit it

pedrobaeza:
When I designed this in previous versions, there were 2 one2many fields and different page visibility for handling this, and it was working properly...

carlos-lopez-tecnativa:
> There's something weird trying in runboat: now the list when you uncheck "Recurrence at line level?" is no more editable, while the idea was to have this "easy mode" for doing the edition directly.

It seems that the `contract_update_last_date_invoiced` module is the one that changes the editable behavior here: https://github.com/OCA/contract/blob/5f6692ff77399ef04f40aeb8637341f0e576724e/contract_update_last_date_invoiced/views/contract_views.xml#L12

In v17 it works the same way, so this change is unrelated.

carlos-lopez-tecnativa (approved):
LGTM

pedrobaeza (approved):
Thanks!

/ocabot merge minor

Next time, please don't introduce empty lines.

PR #1383, opened 2026-01-26
[18.0][REF] product_contract: simplify contract line upsell/downsell code

anthonissen-a:
The suppressed code is actually already present inside the `stop` function.

Also add function _check_set_successor to be able to change this condition in custom modules.

### DIFF
--- a/product_contract/models/contract_line.py
+++ b/product_contract/models/contract_line.py
@@ -75,3 +75,6 @@ class ContractLine(models.Model):
     )
     def _compute_recurring_interval(self):
         return super()._compute_recurring_interval()
+
+    def _check_set_successor(self):
+        return self.state == "closed"
--- a/product_contract/models/sale_order_line.py
+++ b/product_contract/models/sale_order_line.py
@@ -88,7 +88,7 @@ class SaleOrderLine(models.Model):
         contract_line = self.env["contract.line"]
         for rec in self:
             rec.contract_line_id.stop(rec.date_start - relativedelta(days=1))
-            if rec.contract_line_id.state == "closed":
+            if rec.contract_line_id._check_set_successor():
                 new_contract_line = contract_line_model.create(
                     rec._prepare_contract_line_values(contract, rec.contract_line_id.id)
                 )
--- a/product_contract/models/sale_order_line.py
+++ b/product_contract/models/sale_order_line.py
@@ -86,30 +86,13 @@ class SaleOrderLine(models.Model):
     def create_contract_line(self, contract):
         contract_line_model = self.env["contract.line"]
         contract_line = self.env["contract.line"]
-        predecessor_contract_line = False
         for rec in self:
-            if rec.contract_line_id:
-                # If the upsell/downsell line start at the same date or before
-                # the contract line to replace supposed to start, we cancel
-                # the one to be replaced. Otherwise we stop it.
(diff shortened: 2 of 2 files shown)


product_contract/models/sale_order_line.py
```diff
@@ -86,30 +86,13 @@ def _prepare_contract_line_values(
     def create_contract_line(self, contract):
         contract_line_model = self.env["contract.line"]
         contract_line = self.env["contract.line"]
-        predecessor_contract_line = False
         for rec in self:
-            if rec.contract_line_id:
-                # If the upsell/downsell line start at the same date or before
-                # the contract line to replace supposed to start, we cancel
-                # the one to be replaced. Otherwise we stop it.
-                if rec.date_start <= rec.contract_line_id.date_start:
```
qgroulard, line 95: We used to compare `rec.date_start`, now we compare `rec.date_start - relativedelta(days=1)` (the argument `date_end` passed to `stop()`).

Do you think this 1 day difference could have an impact for some corner cases?

product_contract/models/sale_order_line.py
```diff
@@ -86,30 +86,13 @@ def _prepare_contract_line_values(
...
-                    # An exception will be raised if we try to cancel an
-                    # invoiced contract line
-                    rec.contract_line_id.cancel()
-                elif (
-                    not rec.contract_line_id.date_end
-                    or rec.date_start <= rec.contract_line_id.date_end
-                ):
-                    rec.contract_line_id.stop(rec.date_start - relativedelta(days=1))
-                    predecessor_contract_line = rec.contract_line_id
-            if predecessor_contract_line:
```
qgroulard, line 106: The `predecessor_contract_line` used to be reused through the iterations of the `for` loop.
Hence, if `predecessor_contract_line` was set for record1, then not set for record2, the `if predecessor_contract_line` would still be satisfied, considering `record1.contract_line_id` as the predecessor line.

Do you think this was a bug or a feature? 😅 

It looks like a bug to me, but the way it was coded makes me think that it was made on purpose.

qgroulard (commented):
Hi,
Thanks for the contribution.
I like this refactoring but I notice some small behavior changes.

product_contract/models/sale_order_line.py
anthonissen-a: It should be equivalent since the `stop` uses a strict `<`.

qgroulard: Indeed, my bad !

product_contract/models/sale_order_line.py
anthonissen-a: @sbejaoui Could you review this please ?

product_contract/models/contract_line.py
```diff
@@ -75,3 +75,6 @@ def _compute_recurring_invoicing_type(self):
     )
     def _compute_recurring_interval(self):
         return super()._compute_recurring_interval()
+
+    def _check_set_successor(self):
```
sbejaoui:
```suggestion
    def _check_can_have_successor(self):
```

sbejaoui (approved):
LGTM, code review

qgroulard (approved):

bosd (approved):
LGTM

Issue #1385, opened 2026-02-03
Adding optional fields on tree view


noah-lemarinel:
**Is your feature request related to a problem?**
No

**Describe the solution you'd like**
Having the option to toggle additional fields on contract tree view to display more information. It wouldn't change any logic, just couple more lines on tree view.

**Describe alternatives you've considered**
I think it would be a waste of resources to make a new module just for that, therefore, this seems like the best option

**Additional context**
Ideally we would have these fields :
<img width="1854" height="269" alt="Image" src="https://github.com/user-attachments/assets/2f897889-a244-4a8e-9953-70fcbf595072" />

PR #1386, opened 2026-02-03
[18.0][IMP] contract: Adding optional fields on tree view

noah-lemarinel:
- code
- date_start
- date_end
- recurring_next_date
- invoice_partner_id
- commercial_partner_id
- invoice_count

refs #1385


pedrobaeza:
At least `invoice_count` shouldn't be added, as it's not stored, and it will require a lot of computations for showing it up. Some of the rest of the fields are not shown because they may not be valid if the recurrence is set by line.

contract/views/contract.xml
```diff
@@ -425,6 +425,11 @@
                     options="{'color_field': 'color'}"
                 />
                 <field name="company_id" groups="base.group_multi_company" />
+                <field name="code" optional="hide" />
```
carlosdauden: This field is showed in:
https://github.com/OCA/contract/pull/1386/changes#diff-824d27cca50c885b1fb2a66f39e66227a4ea197d6be188550c503be3c1e9e75dR419
Add optional="show" in existing field if you want hide:

contract/views/contract.xml
```diff
@@ -425,6 +425,11 @@
...
+				<field name="date_start" optional="hide" />
+				<field name="date_end" optional="hide" />
+				<field name="invoice_partner_id" optional="hide" />
+				<field name="commercial_partner_id" optional="hide" />
```
carlosdauden: It is best to add these fields after partner_id to keep all related fields together.

carlosdauden (changes requested):

luisDIXMIT (commented):
Can you fix pre-commit issues and squash commits into a single one commit.

PR #1388, opened 2026-02-04
19.0 mig contract fixed

stferraro:


bosd:
I think this one can be closed. As the last commit is only a dotfile improvement on top of the original 19.0 migration on #1312

PR #1399, opened 2026-03-06
[18.0][IMP] subscription_oca: propagate salesman from sale order to subscri…

cvinh:
…ption
Before this commit, the salesman of the subscription is the user who validates the sale order. If the sale order is confirmed by the customer, it's Administrator instead of the sale order salesman

### DIFF
--- a/subscription_oca/models/sale_order.py
+++ b/subscription_oca/models/sale_order.py
@@ -57,7 +57,7 @@ class SaleOrder(models.Model):
             rec = self.env["sale.subscription"].create(
                 {
                     "partner_id": self.partner_id.id,
-                    "user_id": self.env.context.get("uid", self.env.uid),
+                    "user_id": self.user_id.id or self.env.uid,
                     "template_id": subscription_tmpl.id,
                     "pricelist_id": self.partner_id.property_product_pricelist.id,
                     "date_start": date.today(),


luisDIXMIT (approved):
Code review, LGTM!

subscription_oca/models/sale_order.py
```diff
@@ -57,7 +57,7 @@ def create_subscription(self, lines, subscription_tmpl):
             rec = self.env["sale.subscription"].create(
                 {
                     "partner_id": self.partner_id.id,
-                    "user_id": self.env.context.get("uid", self.env.uid),
+                    "user_id": self.user_id.id,
```
tarteo: Should it default to self.env.uid if saleperson is empty? something like this:
`self.user_id or self.env.uid`

cvinh: done

cvinh:
2026-04-21 18:42:58,359 332 WARNING odoo odoo.addons.contract.tests.test_portal.TestContractPortal.test_tour: Killing chrome descendants-or-self of 341: 5 remaining

luisDIXMIT:
Hi @cvinh , this issue was addressed in this PR. I hope it helps you. https://github.com/OCA/web/pull/3507 https://github.com/OCA/oca-addons-repo-template/pull/344

tarteo (approved):
Thanks for making the change!

qgroulard (approved):
LGTM (code review)

PR #1400, opened 2026-03-06
[18.0][FIX] subscription_oca: fix multi-company where the subscription id s…

cvinh:
…is set to the user's company instead of the sale order company

In multi company environment, it can be a mess if the subscription is not in the same company as the sale order's one

It can happen if the user who confirms the sale order is not in the same company as the sale order's


luisDIXMIT (approved):
LGTM!

PR #1401, opened 2026-03-06
[18.0][IMP] subscription_oca: removing automatic date_start to today when subscription goes i…

cvinh:
…n progress. This could be useful when we launch a subscription with a date start different from today

### DIFF
--- a/subscription_oca/models/sale_subscription.py
+++ b/subscription_oca/models/sale_subscription.py
@@ -453,7 +453,8 @@ class SaleSubscription(models.Model):
                     if record.stage_id.type == "in_progress":
                         record.in_progress = True
                         today = date.today()
-                        record.date_start = today
+                        if not record.date_start:
+                            record.date_start = today
                         record.calculate_recurring_next_date(today)
                     elif record.stage_id.type == "post":
                         record.close_reason_id = values.get("close_reason_id", False)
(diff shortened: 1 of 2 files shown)


luisDIXMIT (approved):
LGTM!

rrebollo:
@cvinh PR title should be something along [18.0][IMP] subscription_oca: ....

subscription_oca/models/sale_subscription.py
```diff
@@ -453,7 +453,6 @@ def write(self, values):
                     if record.stage_id.type == "in_progress":
                         record.in_progress = True
                         today = date.today()
-                        record.date_start = today
```
rrebollo, line 456: My concern is: what if `record.date_start` is not set? I checked and it's not a required field.

Maybe you could remove the assignment only for records that already have a value set for the `date_start` field. Though it's possible the scenario I'm describing isn't a valid one.

rrebollo (commented):
Technically correct. Would you consider covering the modification with some tests? I'm guessing a small adjustment to one of the existing tests would work. Ping me then.

cvinh:
> Technically correct. Would you consider covering the modification with some tests? I'm guessing a small adjustment to one of the existing tests would work. Ping me then.

tests added

cvinh:

rrebollo (approved):
Code Review. LGTM!

PR #1404, opened 2026-03-12
[18.0][IMP] subscription_oca : add a setting to choose if the subscription …

cvinh:
…will start automatically when sale order is confirmed or not


luisDIXMIT (approved):
Code review and LGTM!

rrebollo (commented):
From a technical perspective, LGTM.

Would you be open to adding a test covering the change you introduced to the "standard" feature? Before this PR, my understanding from the code is that every newly created subscription automatically started. Now, with your setting, that won't happen by default.

I think adding a test would be a good safety measure—let's see how it goes. You can ping me then.

Also, would you be so kind to review my #1410 in return?

cvinh:
> From a technical perspective, LGTM.
> 
> Would you be open to adding a test covering the change you introduced to the "standard" feature? Before this PR, my understanding from the code is that every newly created subscription automatically started. Now, with your setting, that won't happen by default.
> 
> I think adding a test would be a good safety measure—let's see how it goes. You can ping me then.
> 
> Also, would you be so kind to review my #1410 in return?

tests added

cvinh:

subscription_oca/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 {
     "name": "Subscription management",
     "summary": "Generate recurring invoices.",
-    "version": "18.0.1.0.0",
+    "version": "18.0.1.1.0",
```
rrebollo, line 6: I believe the version bump should only apply at merge time—though I could be wrong.

rrebollo (approved):
Code Review. LGTM!

rrebollo:
@cvinh did you rebase? The tests might pass then.

subscription_oca/__manifest__.py
cvinh: Right

PR #1407, opened 2026-03-17
[18.0][FIX] contract: recurring_next_date : update lines when changed on contract

jbaudoux:
### Issue
The recurring next date is editable on the contract but changing it has no impact.
The invoicing is based on the lines recurring next date which is not updated. At invoicing, a new recurring next date is recomputed based on the lines and the contract recurring next date is recomputed. So the new date is not taken into account and is lost.

### Fix 
Update lines recurring next date when the contract recurring next date is modified

cc @sbejaoui @florian-dacosta @pedrobaeza @etobella

### DIFF
--- a/contract/models/contract.py
+++ b/contract/models/contract.py
@@ -5,6 +5,7 @@
 # Copyright 2016-2017 LasLabs Inc.
 # Copyright 2018 ACSONE SA/NV
 # Copyright 2021 Tecnativa - Víctor Martínez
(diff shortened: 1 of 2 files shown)


jbaudoux:
As that date has an impact on the next period end date, it make sense the date is not updated on the lines. Otherwise the period is not respected anymore. I assume the next invoice date can be changed at contract level just to set the generated invoice date

PR #1391, opened 2026-02-16
[18.0][FIX] contract_invoice_manually: Fix wizard button

tobiaszehntner:
Wizard button is not visible due to xpath 'move' not working on entire object

image

PR #1408, opened 2026-03-17
[18.0][FIX] contract_manually_invoiced: wizard invoice manually view

jbaudoux:
The wizard to invoice manually was not displayed properly.
I also allow to choose to invoice only the manually invoiced contracts or not.
I removed the button to view the contracts as this is there by default in contract module.

cc @tobiaszehntner

### DIFF
--- a/contract_invoice_manually/wizards/contract_manually_create_invoice.py
+++ b/contract_invoice_manually/wizards/contract_manually_create_invoice.py
@@ -1,14 +1,18 @@
 # Copyright 2026 ACSONE SA/NV
+# Copyright 2026 Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 
 from odoo import api, fields, models
-from odoo.osv import expression
 from odoo.tools.safe_eval import safe_eval
 
 
 class ContractManuallyCreateInvoice(models.TransientModel):
     _inherit = "contract.manually.create.invoice"
+
+    manually_invoiced = fields.Boolean(
+        default=lambda self: self.env.company.enable_contract_invoice_manually
+    )
 
     filter_domain = fields.Char(
         string="Domain",
@@ -18,7 +22,7 @@ class ContractManuallyCreateInvoice(models.TransientModel):
         help="Filter/Domain to apply on contracts to invoice",
     )
 
-    @api.depends("invoice_date", "contract_type")
+    @api.depends("invoice_date", "contract_type", "manually_invoiced")
     def _compute_filter_domain(self):
         for wizard in self:
             domain = [
@@ -28,9 +32,8 @@ class ContractManuallyCreateInvoice(models.TransientModel):
                     fields.Datetime.to_string(wizard.invoice_date),
                 ),
                 ("contract_type", "=", wizard.contract_type),
+                ("is_manually_invoiced", "=", wizard.manually_invoiced),
             ]
-            if self.env.company.enable_contract_invoice_manually:
-                domain = expression.AND([domain, [("is_manually_invoiced", "=", True)]])
             wizard.filter_domain = str(domain)
 
     @api.depends("invoice_date", "filter_domain")
--- a/contract_invoice_manually/wizards/contract_manually_create_invoice.xml
+++ b/contract_invoice_manually/wizards/contract_manually_create_invoice.xml
@@ -9,22 +9,21 @@
             ref="contract.contract_manually_create_invoice_form_view"
         />
         <field name="arch" type="xml">
-            <xpath expr="//form/group[last()]" position="after">
-                <notebook invisible="not invoice_date">
-                    <page name="contracts" string="Contracts">
-                        <button
-                            name="action_show_contract_to_invoice"
-                            position="move"
-                        />
-                    </page>
-                    <page name="domain" string="Contract Filter">
-                        <field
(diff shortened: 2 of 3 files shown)


contract_invoice_manually/wizards/contract_manually_create_invoice.xml
```diff
@@ -9,22 +9,21 @@
...
-                    </page>
-                </notebook>
+            <field name="invoice_date" position="after">
+                <field name="manually_invoiced" />
+            </field>
+            <xpath expr="//form/group" position="after">
+                <group
+                    string="Contract Filter"
+                    invisible="not invoice_date"
+                    groups="base.group_no_one"
```
tobiaszehntner, line 19: I find it's too limiting to put the filter only for debug-mode. Our clients use filtering regularly but aren't necessarily familiar with debug mode. However, I agree it's a bit of an extra technical thing, hence my reasoning to put it away in a second tab: accessible but not confusing.

contract_invoice_manually/wizards/contract_manually_create_invoice.xml
```diff
@@ -9,22 +9,21 @@
             ref="contract.contract_manually_create_invoice_form_view"
         />
         <field name="arch" type="xml">
-            <xpath expr="//form/group[last()]" position="after">
-                <notebook invisible="not invoice_date">
-                    <page name="contracts" string="Contracts">
-                        <button
```
tobiaszehntner, line 15: the button is fixed here: https://github.com/OCA/contract/pull/1391

contract_invoice_manually/wizards/contract_manually_create_invoice.py
```diff
@@ -1,15 +1,19 @@
...
 
 from odoo import api, fields, models
-from odoo.osv import expression
 from odoo.tools.safe_eval import safe_eval
 
 
 class ContractManuallyCreateInvoice(models.TransientModel):
     _inherit = "contract.manually.create.invoice"
 
+    manually_invoiced = fields.Boolean(
```
tobiaszehntner, line 13:
```suggestion
    is_manually_invoiced = fields.Boolean(
        "Only Manually Invoiced Contracts
```
Yes that could add some flexibility (even though it's already there via the filter for the technically versed users)

contract_invoice_manually/wizards/contract_manually_create_invoice.py
```diff
@@ -28,9 +32,8 @@ def _compute_filter_domain(self):
                     fields.Datetime.to_string(wizard.invoice_date),
                 ),
                 ("contract_type", "=", wizard.contract_type),
+                ("is_manually_invoiced", "=", wizard.manually_invoiced),
             ]
-            if self.env.company.enable_contract_invoice_manually:
-                domain = expression.AND([domain, [("is_manually_invoiced", "=", True)]])
```
tobiaszehntner, line 33: I don't know what your use case is, but I think it's not clear from the boolean in the view that unticking it would **exclude** the manually invoiced contracts. Maybe only add the domain if it's ticked, and otherwise include *both* types of contracts?

contract_invoice_manually/wizards/contract_manually_create_invoice.xml
```diff
@@ -9,22 +9,21 @@
...
-                    <page name="domain" string="Contract Filter">
-                        <field
-                            name="filter_domain"
-                            widget="domain"
-                            options="{'model': 'contract.contract'}"
-                        />
-                    </page>
-                </notebook>
+            <field name="invoice_date" position="after">
+                <field name="manually_invoiced" />
```
tobiaszehntner, line 13: This should only show when the config setting is active

tobiaszehntner (changes requested):
You hide the filter to debug-mode while adding the boolean. The idea of the filter is that the user can do all specifics they would like, including choosing non-manually-invoiced contracts.

If we add the boolean for user-friendliness, I would suggest to make it clearer ('Only include manually invoice contracts'), and avoid excluding them if it's unticked.

jbaudoux:
> You hide the filter to debug-mode while adding the boolean. The idea of the filter is that the user can do all specifics they would like, including choosing non-manually-invoiced contracts.
> 
> If we add the boolean for user-friendliness, I would suggest to make it clearer ('Only include manually invoice contracts'), and avoid excluding them if it's unticked.

@tobiaszehntner Thanks for the feedback. For the manually invoiced contracts, the most convenient I find is to select some contracts from the list view and launch a "create invoice" action. You can use the standard filters on the list view (To Invoice & Manually Invoiced).
The wizard was used for automaticaly invoiced contracts, to launch the invoicing when you don't use the cron. I found it suddenly unfriendly to have this filter changing the base behavior.
I agree this boolean is not ideal. Maybe an expliclit selection "Only non manually invoiced", "Only manually invoiced", "Both" would be better.
Nevertheless, I still find it strange to invoice in mass manually invoiced contracts, but it's up to the user to decide. It's more usefull for automaticaly invoiced contracts like it's with the base module. And for manually invoiced contracts to manage them one by one without the wizard.

tobiaszehntner:
@jbaudoux you seem to describe the basic behavior from `contract` no? This is why this is a separate module with a company config. It's because our client has some contracts that are *always* manually invoiced (but hundreds of them), and it makes it easier to do that with the boolean on the contract and through the wizard. If you want to manually invoice contracts that you select in list view, and use the wizard occasionally, you indeed might not need this module?

PR #1409, opened 2026-03-17
[19.0][FIX] contract: Add missing translated_product_name field

bosd:
## Summary

Odoo core PR #248401 introduced a new JS widget (`product_label_section_and_note_field`) that requires the `translated_product_name` field on any model using section/note display types. Since `contract.line` uses this widget but does not define the field, opening a contract form view crashes with:

```
KeyError: 'translated_product_name'
```

This PR adds a computed `Text` field on `contract.line` that mirrors the pattern used by `sale.order.line` in Odoo core: it returns the product `display_name` translated into the contract partner's language.

## Changes

- Added `translated_product_name` computed field to `contract.line`
- Compute method translates product name using the partner's language (falls back to env language)

## Test plan

- [x] Open a contract with lines — no more `KeyError`
- [x] Verify translated product names display correctly for partners with different languages
- [x] Pre-commit passes

Based on: #1312


pedrobaeza:
The PR doesn't seem correct due to the number of modified files.

bosd:
There are indeed a lot of modified files. Similar as in #1312

pedrobaeza:
Then instead opening a new PR, just mention the patch in the open migration PR.

bosd:
I am actually trying to create a pr against the migration pr. Somehow it is not working. will ask to cherry-pick instead.

bosd:
Need to check if it still is needed.. Odoo partially erverted the problematic commit: https://github.com/odoo/odoo/commit/199038099f7c6717e2c79ec230562514db52facc

bosd:
Closing — no longer needed on 19.0.

This PR worked around `KeyError: 'translated_product_name'`, which was caused by Odoo core PR #248401 making the `product_label_section_and_note_field` widget depend on a `translated_product_name` field. Odoo has since **partially reverted** that change (odoo/odoo@199038099f), removing both the `translatedProductName` getter and the `fieldDependencies: [{ name: 'translated_product_name', ... }]` declaration from the widget. On `19.0` the widget no longer references the field, so `contract.line` no longer crashes and the field is unnecessary.

No functional impact from closing this.

PR #1060, opened 2024-03-15
[14.0][FIX] subscription_oca: product-subscription_template company issues

HekkiMelody:
Make company required on subscription_template.

Make subcription_template_id a company-dependent field on product_template.

On SO, set a team whose company is consistent with the SO's company.

TODO:
  - [x] must verify migration of existing data

Migration of existing data is necessarily imperfect, because previous data may be inconsistent in the first place.

PR #1410, opened 2026-03-19
[18.0] [FIX] subscription_oca: add migration script for subscription_template_id field on product.template model 

rrebollo:
## Context

We discovered this issue during a customer migration from v14 to v18: the `subscription_template_id` field value for product templates is lost after upgrading to v18.

## Root Cause

#1060 made `subscription_template_id` a company-dependent field on `product.template`, but this change was never forward-ported beyond its target version (v14).

In Odoo 18.0, company-dependent fields are stored in a JSONB column instead of `ir_property` records. Since the forward-port never happened, the automatic migration from `ir_property` to JSONB did not occur, leaving the field empty after upgrade.

## Solution

This pull request adds a pre-migration script for Odoo 18.0 that backfills `product_template.subscription_template_id` from legacy `ir_property` records, ensuring existing data is preserved during the upgrade.

## Scope

This fix applies in a way that **only** instances affected by the company-dependent implementation of the `subscription_template_id` field will be targeted. Nothing changes or backfires if you're not affected.

## Migration and Data Integrity

- Added `pre-migration.py` to backfill `product_template.subscription_template_id` from legacy `ir_property` records
- Documented the root cause and provided context about the storage change for company-dependent fields

### DIFF
--- a/subscription_oca/upgrades/18.0.1.0.0/pre-migration.py
+++ b/subscription_oca/upgrades/18.0.1.0.0/pre-migration.py
@@ -1,0 +1,25 @@
+# Copyright 2026 Binhex (<https://www.binhex.cloud>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    # PR OCA/contract#1060 made subscription_template_id a company-dependent field on
+    # product.template, but was never forward-ported beyond the version it targeted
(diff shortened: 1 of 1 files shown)


luisDIXMIT (approved):
Code review, LGTM!

cvinh (approved):
LGTM code review no test

rrebollo:
@pedrobaeza any chances to get this merged?

pedrobaeza:
Maintainers of this modules should trigger it, as I don't use (and you know that I don't want this duality `contract`/`subscription_oca`.

rrebollo:
@rousseldenis @sbejaoui would you merge this one?

Issue #1412, opened 2026-03-30
[18.0][BUG] contract: future contracts smart button is hidden on contacts without customer_rank or supplier_rank


fcvalgar:
## Module
contract

## Describe the bug
When a future contract is created for a contact/partner, the contract smart button is not shown on the contact form if the partner was created from the Contacts module or through any flow that does not set `customer_rank` or `supplier_rank`.

As a result, the related contract exists, but the user cannot access it from the partner smart buttons.

<img width="3423" height="1182" alt="Image" src="https://github.com/user-attachments/assets/2c4f4adf-e4f8-4fc6-92d9-c479935cb491" />

<img width="3427" height="1181" alt="Image" src="https://github.com/user-attachments/assets/ca1b5073-b87f-4a89-b662-4c91c4e1a482" />

## To Reproduce
**Affected versions**:
18.0

1. Create a new contact from the Contacts module.
2. Make sure the partner is not created through a Sales or Purchase flow, so `customer_rank` and/or `supplier_rank` remain `0`.
3. Create a future contract linked to that partner.
4. Open the partner form.

<img width="1750" height="1139" alt="Image" src="https://github.com/user-attachments/assets/6f9fd8d3-759a-40a1-9409-40a2a3847d8d" />

<img width="1751" height="1140" alt="Image" src="https://github.com/user-attachments/assets/a52dc6d0-a31e-48f5-b817-293ba7b87fb9" />

**Expected behavior**
The smart button for contracts should be visible whenever the partner has related contracts, regardless of whether `customer_rank` or `supplier_rank` is set.

## Additional context
The issue seems to come from the smart button visibility conditions:
- Sale Contracts button: invisible when `customer_rank == 0`
- Purchase Contracts button: invisible when `supplier_rank == 0`

Because of this, partners created outside Sales/Purchases do not show the contract smart button, even if they already have linked contracts.

[MT-14328](https://www.moduon.team/odoo/project.task/project.task/14328) @moduon

PR #1413, opened 2026-03-31
[18.0][FIX] contract: contracts from partners not yet invoiced

chienandalu:
When we have a brand new customer or supplier, we could't see their contracts until those are invoiced and the customer/supplier rank counter is started. Better rely on the existing sale/purchase contracts counter.

image

Should close:

- https://github.com/OCA/contract/issues/1412

cc @moduon MT-14364

please review @fcvalgar @Gelojr

### DIFF
--- a/contract/views/res_partner_view.xml
+++ b/contract/views/res_partner_view.xml
@@ -14,7 +14,7 @@
                     groups="account.group_account_invoice"
                     icon="fa-book"
                     context="{'default_contract_type': 'sale', 'contract_type': 'sale'}"
-                    invisible="customer_rank == 0"
+                    invisible="sale_contract_count == 0"
                     help="Show the sale contracts for this partner"
                 >
                     <field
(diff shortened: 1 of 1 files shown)


fcvalgar (approved):
Tested and working as expected.

I have verified the behavior by creating the contact from Contacts, Sales, and Purchase. In all cases, the contact correctly shows the corresponding smart button.

Thanks for the fix @chienandalu .
LGTM

Image

carlosdauden:
When contacts are created via the Customer or Supplier actions, the x_rank field is set correctly:
https://github.com/odoo/odoo/blob/be579b408547d45503d3953ee2b3696e76d76992/addons/account/views/partner_view.xml#L352
https://github.com/odoo/odoo/blob/74607bc3196667f282795e0335e84ab95040fd55/addons/account/models/partner.py#L832-L835

When you create a Customer/Supplier following the ‘correct workflow’, you expect the smartbutton to be visible so that you can create a contract from there.

With this change, what will happen to the smartbutton when a Customer/Supplier is created?
I think it’s fine to try to solve a problem that users who don’t follow the ‘correct workflow’ might have, but under no circumstances should this be at the sacrifice of functionality for those who work correctly.

If you want to proceed this way, what you should do is use the _increase_rank method when creating a contract in the same way as when creating an invoice:
https://github.com/odoo/odoo/blob/f5501e5c8dcf60444077912db4c87e7a3f2654a6/addons/account/models/account_move.py#L5239-L5241

ping @sergio-teruel

chienandalu:
Well, yes, taking that perspective I think you're absolutly right. Thanks for you review @carlosdauden :wink:

PR #1418, opened 2026-04-27
[19.0][FIX] checklog-odoo.cfg: ignore Killing chrome descendants-or-self warning

BhaveshHeliconia:


pedrobaeza:
Please update copier template instead.

PR #1420, opened 2026-04-27
[18.0] [FIX] subscription_oca: prevent CRM tag context pollution on subscription create

rrebollo:
When confirming a sale.order from a CRM opportunity, sale_crm injects default_tag_ids (containing crm.tag IDs) into the context. Without this fix, those IDs are applied by default_get() to sale.subscription.tag_ids, causing a FK constraint violation on sale_subscription_sale_subscription_tag_rel.

Fix: use clean_context() from odoo.tools to strip all default_* keys from context before calling sale.subscription.create(), following the idiomatic Odoo core pattern used in res_bank, account_move, hr_expense_sheet, etc.

Adds a regression test that reproduces the production scenario end-to-end.

@BinhexTeam T18387

### DIFF
--- a/subscription_oca/models/sale_order.py
+++ b/subscription_oca/models/sale_order.py
@@ -6,6 +6,7 @@ from datetime import date
 from dateutil.relativedelta import relativedelta
 
 from odoo import Command, api, fields, models
+from odoo.tools import clean_context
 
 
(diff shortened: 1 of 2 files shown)


tarteo (approved):

rrebollo:
@ajfebres please review this one

PR #1421, opened 2026-04-28
[16.0][FIX] Upgrade pre-commit dependency setuptools-odoo to v3.3.2

ivilata:
That version avoids `ModuleNotFoundError: No module named 'pkg_resources'` when running the Git pre-commit hook, which causes an error in all CI runs.

See #1402 or #1419 for examples of 16.0 PRs failing because of this error, and acsone/setuptools-odoo#126 for some context on the error (caused by the recent removal on `pkg_resources` from `setuptools`) and how `setuptools-odoo` v3.3.2 fixes it.

Pre-commit hook configurations for 17.0 and 18.0 don't seem to depend on `setuptools-odoo`.  Similar changes may be ported back to 16.0 as a more stable fix, but the current one is a very simple, low-impact fix that may help unlock PRs right now.

PR #1422, opened 2026-04-28
[16.0][FIX] Update dotfiles and unit test code

ivilata:
Mainly to avoid `ModuleNotFoundError: No module named 'pkg_resources'` when running the `setuptools-odoo` pre-commit hook (e.g. run by CI), as v3.1.8 was bitten by the removal of `pkg_resources` from recent versions of `setuptools`.  This was fixed by v1.38 of `oca-addons-repo-template` by updating `setuptools-odoo` to v3.3.2.

See #1421 for more more information.

Update 2026-05-12: This also updates `contract_payment_auto` and `subscription_oca` unit tests to avoid a few assorted errors that crept in with other commits. These should unlock merging in other PRs.


ivilata:
As mentioned in the other ticket, `KeyError: 'test_target_state'` in `TestContract` tests are pre-existing.

pedrobaeza (approved):
Anyway, to merge this, we need to fix the error in the tests.

ivilata:
Thanks @pedrobaeza for the review.

Umm… So a merge is not possible if there are unrelated test failures?

I wonder if creating additional PRs for solving the other issues won't work since they will be blocked by this one, while this one is blocked by the others.

In that case, should the fixes to the other problems be included in this PR as well? :thinking:

pedrobaeza:
Yeah, it should be included in this PR.

ivilata:
@pedrobaeza: It took me some time, but at last I was able to fix the remaining unit test errors. I understand that, since the fixes are unrelated but need to be pushed in a single blow to avoid the chicken-and-egg situation mentioned above, it's better not to squash them together.

contract_payment_auto/tests/test_contract.py
```diff
@@ -136,8 +136,8 @@ def test_pay_invoice_not_open(self):
 
     def test_pay_invoice_no_residual(self):
         """It should return None if no residual on the invoice."""
-        invoice = self._create_invoice()
-        invoice.state = "posted"
+        invoice = self._create_invoice(True)
+        invoice.amount_residual = 0.0
```
pedrobaeza, line 140: This is tricking too much the aim of the tests, as this is saying there's no debt. Not sure though if the tests were OK before.

ivilata: Thanks @pedrobaeza for the comment, I'm struggling quite blindly with these unrelated test errors…

Honestly, I don't know if posting or even sending (i.e. `self._create_invoice(True, True)`) are supposed to clear the residual in the test context (note: they don't), or if some other operation should be issued in the middle (hints welcome!).

At any rate, this `contract_payment_auto` module doesn't even exist in 17.0 nor 18.0, I wonder if just dropping the test would make sense at this point if we fail to make sense out of it. :shrug:

pedrobaeza:
OK, let's continue for now this way, and let that tests as an eccentricity, hehe, but for sure it's not correct (and I'm pretty sure it wasn't before).

/ocabot merge patch

PR #1429, opened 2026-05-13
Add a method to get total value of a contract line

baimont:
The motivation for adding this method to the base `contract` module rather than to extension modules is extensibility: many modules customize contract behavior (custom quantity rules, variable pricing, etc.).
By overriding `_get_contract_line_total_value()` in those modules, they can influence the total value used downstream by any module that needs it — such as some modules we'll announce soon with @sbidoul  — without creating unwanted cross-dependencies.

### DIFF
--- a/contract/models/contract_line.py
+++ b/contract/models/contract_line.py
@@ -8,8 +8,8 @@ import warnings
 
 from dateutil.relativedelta import relativedelta
 
-from odoo import api, fields, models
-from odoo.exceptions import ValidationError
+from odoo import _, api, fields, models
+from odoo.exceptions import UserError, ValidationError
 
 
 class ContractLine(models.Model):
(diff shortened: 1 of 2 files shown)


baimont:
@qgroulard @sbejaoui take a look please :-)

contract/models/contract_line.py
```diff
@@ -312,3 +312,25 @@ def _get_quantity_to_invoice(
     ):
         self.ensure_one()
         return self.quantity if not self.display_type else 0.0
+
+    def _get_contract_line_total_value(self):
+        """Return the total value of this contract line over its full period.
+
+        Raises NotImplementedError if the line has no date_end, since a
```
sbejaoui:
```suggestion
        Raises UserError if the line has no date_end, since a
```

sbejaoui (approved):
LGTM,

sbidoul (approved):

contract/models/contract_line.py
```diff
@@ -318,30 +307,30 @@ def get_view(self, view_id=None, view_type="form", **options):
...
-            )
-            line.monthly_recurring = line.price_subtotal / months if months else 0.0
-
     def _get_quantity_to_invoice(
         self, period_first_date, period_last_date, invoice_date
     ):
         self.ensure_one()
         return self.quantity if not self.display_type else 0.0
+
+    def _get_contract_line_total_value(self):
```
carlos-lopez-tecnativa, line 316: This method is not called from any code and is only used in tests. What is its purpose?

contract/models/contract_line.py
```diff
@@ -318,30 +307,30 @@ def get_view(self, view_id=None, view_type="form", **options):
...
+                    "no end date is set.",
+                    self.display_name,
+                )
+            )
+        quantity = self._get_quantity_to_invoice(
+            self.date_start,
+            self.date_end,
+            self.date_start,
+        )
+        return quantity * self.price_unit
```
carlos-lopez-tecnativa, line 336: You didn’t consider the discount. Is this correct?

contract/models/contract_line.py
baimont: hello. indeed it is not called from contract module. it was put there because some modules depending on contract may have to get this information. this seemed like the right place

contract/models/contract_line.py
baimont: correct

PR #1435, opened 2026-05-21
[18.0][IMP] contract: add monthly_recurring field to contracts and lines

cristina-hidalgo-tecnativa:
Add a new stored monetary field `monthly_recurring` on `contract.line` and `contract.contract` that normalises each line's amount to its monthly equivalent based on its recurrence rule and interval.

It is optional (hidden by default) in all list views and shows a sum in the footer, allowing users to see the total recurring monthly amount across all contracts and lines at a glance. Non-active lines can be excluded using the existing "In progress" filter.

@Tecnativa TT62575
@pedrobaeza @christian-ramos-tecnativa

### DIFF
--- a/contract/models/contract_line.py
+++ b/contract/models/contract_line.py
@@ -11,6 +11,17 @@ from dateutil.relativedelta import relativedelta
 from odoo import api, fields, models
 from odoo.exceptions import ValidationError
 
+# Months equivalent per recurrence unit (to normalise amounts to monthly)
+_MONTHS_PER_RULE = {
+    "daily": 1.0 / 30.0,
+    "weekly": 7.0 / 30.0,
+    "monthly": 1.0,
+    "monthlylastday": 1.0,
+    "quarterly": 3.0,
+    "semesterly": 6.0,
+    "yearly": 12.0,
+}
+
 
 class ContractLine(models.Model):
     _name = "contract.line"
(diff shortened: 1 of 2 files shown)


pedrobaeza:
You should remove the conditions for not started/cancelled lines, as if not, data won't be refreshed due to the time condition (there's no refresh when a line starts).

We should control what we see/sum through the filters in the pivot removing non valid lines.

cristina-hidalgo-tecnativa:
> You should remove the conditions for not started/cancelled lines, as if not, data won't be refreshed due to the time condition (there's no refresh when a line starts).
> 
> We should control what we see/sum through the filters in the pivot removing non valid lines.

Fixed. Removed the time-based conditions.

contract/models/contract.py
```diff
@@ -126,6 +126,12 @@ class ContractContract(models.Model):
     # === Dates ===
     date_end = fields.Date(compute="_compute_date_end", store=True, readonly=False)
 
+    monthly_recurring = fields.Monetary(
```
pedrobaeza: I don't think we should include this one, as this total will be incorrect with cancelled lines, deprecated, etc.

cristina-hidalgo-tecnativa: Removed the field from `contract.contract`, keeping it only at line level.

pedrobaeza (approved):
Maybe we should update the Reporting > Customer Contract Lines for having better defaults in the pivot view with the MRR measure and removing creation date columns.

cristina-hidalgo-tecnativa:
> Maybe we should update the Reporting > Customer Contract Lines for having better defaults in the pivot view with the MRR measure and removing creation date columns.

Done, removed `date_start` column grouping as it wasn't meaningful as a default, and set `monthly_recurring` as the default measure so the pivot is immediately useful for analysing MRR per contract.

pedrobaeza:
Thank you for the fine-tuning. For being even more exact, we should filter in that view only the valid lines, and we finally target the goal.

cristina-hidalgo-tecnativa:
> Thank you for the fine-tuning. For being even more exact, we should filter in that view only the valid lines, and we finally target the goal.

Added an 'In progress' filter that excludes cancelled, not yet started, and already ended lines, and set it as the default in both the customer and supplier reporting actions.

christian-ramos-tecnativa (approved):

pedrobaeza:
Good contribution!

/ocabot merge minor

PR #1441, opened 2026-05-27
[19.0][IMP] subscription_oca: track invoice periods on move lines

alvaro-domatix:
Invoice lines generated from a subscription keep no reference to the subscription nor to the period they bill, so there is no way to tell which period an invoice line covers or to analyze invoiced amounts by subscription.

This adds `subscription_id`, `subscription_period_start` and `subscription_period_end` on `account.move.line`, filled when the subscription generates the invoice. The line description now also shows the billed period, formatted with the partner's lang.

It is also a prerequisite to detect already invoiced periods and avoid generating duplicated invoices, which I plan to submit as a follow-up PR.


nachoserra-domatix (approved):

gerard-domatix (approved):

jelenapoblet:
Same here, @alvaro-domatix like the conflicts are limited to:

subscription_oca/__manifest__.py (version bump)
subscription_oca/tests/__init__.py (both test imports should be kept)

Could you please rebase the branch and update the PR?

jorgeglez1990 (commented):
@alvaro-domatix Functional review: it would be advisable to validate the Period Date Format field in the settings, as an incorrect configuration could cause errors during invoice generation.

alvaro-domatix:
Rebased on top of the current 19.0 (`tests/__init__.py` keeps both imports, manifest version aligned with 19.0).

@jorgeglez1990 good point — I've added an `@api.constrains` on `subscription_period_date_format` that renders a sample date with the configured Babel pattern and raises a `ValidationError` on save. An invalid format now fails fast at settings time instead of crashing invoice generation (which the cron would otherwise swallow silently). Empty is still allowed (falls back to the customer language). Two tests cover the rejection and the empty case. CI is green.

alvaro-domatix:
Gentle nudge on this one when there's a window. @jorgeglez1990

alvaro-domatix:
Bumping for a maintainer look. @pedrobaeza

pedrobaeza:
I think @jelenapoblet is better suited, as he uses this module, and you should become maintainer of the module for being more self-independent.

jelenapoblet:
@jorgeglez1990 returns next week from vacations, he was looking at this.

PR #1446, opened 2026-06-03
[18.0][IMP] contract: Contract line total amount

qgroulard:
Improve feature introduced in https://github.com/OCA/contract/pull/1429.

Previous implementation had 3 problems:
* Discounts were not taken into account
* When combined with 'contract_variable_quantity', '_get_quantity_to_invoice()' returns 1 for fixed quantities
* When combined with 'contract_variable_quantity', it is wrong to assume that the quantity to invoice of one big period is equal to the quantity to invoice of the sum of smaller chunks

Because of the two last problems, we are forced to iterate over each periods and sum the subtotals.

### DIFF
--- a/contract/models/contract_line.py
+++ b/contract/models/contract_line.py
@@ -361,9 +361,32 @@ class ContractLine(models.Model):
                     self.display_name,
                 )
             )
-        quantity = self._get_quantity_to_invoice(
-            self.date_start,
-            self.date_end,
-            self.date_start,
+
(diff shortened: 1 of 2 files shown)


qgroulard:
@baimont

qgroulard:
Hi @carlos-lopez-tecnativa, you seemed interested in this.

baimont (approved):

rousseldenis (approved):

PR #1449, opened 2026-06-09
[16.0][FIX] contract: preserve manually set recurring_next_date on lines

### LINKED ISSUE #1411: [16.0] contract: Editing `recurring_next_date` on a modal `contract.line` view resets the value when saving
## Module
`contract`

## Describe the bug
Whenever I go on a `contract.contract` form view, for a contract with line recurrence which was already invoiced (`last_date_invoiced` set), I try to open one line in a modal view, to modify its `recurring_next_date` and to Save it.
However, upon closing, the `recurring_next_date` value of the line is reset to the `next_period_date_start` value.

I'm not exactly sure why it doesn't save the `recurring_next_date` value on the `contract.line` record when clicking Save - perhaps since the Edit mode is still on, it doesn't save the value in the database, however it gets overwritten anyways.

## To Reproduce
**Affected versions**: 16.0 (afaik)

1. Create a `contract.contract` entity from a form view
2. Check "Line recurrence"
3. Create a `contract.line` with an arbitrary product
4. Set a `date_start` before today, to set the `next_period_date_start` and `recurring_next_date` fields
5. Save and exit
6. Click on the newly created `contract.line`
7. Set the `recurring_next_date` to a later date
8. Click on `Save`

**Expected behavior**
I would like for the `recurring_next_date` value to not get overwritten and reset back to the `next_period_date_start` value

**Additional context**
I've found various workarounds, however they're not handy for reasons : 
 
- I can go on a dedicated tab to display the `contract.line` form view, edit the `recurring_next_date` value and save, and this value doesn't get overwritten - however, since no web menu allows browsing through `contract.line` entities, it requires knowing the ID of the `contract.line` and setting it in the URL bar, which is not handy for functionnal users
 
- I can install the `contract_update_last_invoice_date` module, and set the value through the wizard (since it directly saves on the `contract.line` record) and quit the modal view without clicking on `Save`.
  However, this is an unstable solution, since modifying the `recurring_next_date` field directly through the modal view after can reset the `recurring_next_date` field.

maisim:
When a user manually modifies recurring_next_date on a contract line that belongs to a contract with line_recurrence=True, the value was overwritten by the cross-dependency recomputation chain:

1. Writing line.recurring_next_date triggers contract._compute_recurring_next_date() (depends on contract_line_ids.recurring_next_date)
2. Contract recomputation triggers line._compute_recurring_next_date() (depends on contract_id.recurring_next_date)
3. super()._compute_recurring_next_date() unconditionally recomputes from next_period_date_start, overwriting the user's value

The fix adds a guard (similar to the existing one in the contract model for date_start) that detects manual modifications by comparing rec._origin.recurring_next_date with the current value, and skips the recomputation for records where the user has explicitly set the value.

Fixes #1411

### DIFF
--- a/contract/models/abstract_contract_line.py
+++ b/contract/models/abstract_contract_line.py
@@ -171,7 +171,24 @@ class ContractAbstractContractLine(models.AbstractModel):
     # pylint: disable=missing-return
     @api.depends("contract_id.recurring_next_date", "contract_id.line_recurrence")
     def _compute_recurring_next_date(self):
-        super()._compute_recurring_next_date()
+        # Preserve manually set recurring_next_date when line_recurrence is enabled.
+        # The cross-dependency chain (contract depends on line's recurring_next_date,
+        # and line depends on contract's recurring_next_date) would otherwise
+        # overwrite user-modified values with the computed value from
+        # next_period_date_start. We detect manual modifications by comparing
+        # with _origin and skip recomputation for those records.
+        to_compute = self.filtered(
+            lambda r: not (
+                r.contract_id.line_recurrence
+                and r._origin
+                and r._origin.recurring_next_date
+                and r._origin.recurring_next_date != r.recurring_next_date
+            )
+        )
+        if to_compute:
+            super(
+                ContractAbstractContractLine, to_compute
+            )._compute_recurring_next_date()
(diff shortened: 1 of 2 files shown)


maisim:
replaced by https://github.com/OCA/contract/pull/1462

PR #1443, opened 2026-05-27
[IMP] subscription_oca: compute MRR and ARR

alvaro-domatix:
## Problem

`sale.subscription.recurring_total` is the sum of the line `price_subtotal` without normalising to a single cadence. A yearly subscription that bills 1 200 € once a year and a monthly one that bills 100 € a month look very different in the list view even though they generate identical recurring revenue. Aggregating across cadences is impossible — the basic management question "how much do I bill per month?" can only be answered by exporting to a spreadsheet and dividing case by case.

## Solution

Add two new stored computed monetary fields on `sale.subscription`:

- `recurring_monthly` — monthly recurring revenue (MRR), summed from a per-line `recurring_monthly`.
- `recurring_yearly` — twelve times the monthly figure (ARR).

The line-level `recurring_monthly` divides `price_subtotal` by the number of months in the line's period. The factor is `1 / 30.4375` for days, `7 / 30.4375` for weeks, `1.0` for months and `12.0` for years, multiplied by `recurring_interval`. So a 100 €/week line yields ≈ 434.82 €/month, a 1 200 €/year line yields 100 €/month, a 300 € line every 3 months yields 100 €/month.

UI changes:

- Form view: new stat button showing MRR and ARR side by side.
- List view: `recurring_monthly` (`sum` footer, optional `show`) and `recurring_yearly` (optional `hide`).
- Search view: new "With recurring revenue" filter (`recurring_monthly > 0`).

No existing field is renamed, dropped or otherwise altered (see §0.4 of the internal compatibility checklist). Pure additive change.

## Migration

`-u subscription_oca` triggers a normal ORM recompute. No data script. Estimate roughly two seconds per ten thousand subscription lines on commodity hardware.

## How to test

1. Install the module and create a template with `recurring_rule_type = months`, `recurring_interval = 1`. Add a subscription line with a 100 € product → `recurring_monthly = 100`, `recurring_yearly = 1200`.
2. Switch the template to `years` → `recurring_monthly = 100 / 12 ≈ 8.33`.
3. Switch to `weeks`, interval 1 → `recurring_monthly = 100 * 30.4375 / 7 ≈ 434.82`.
4. Switch to `months`, interval 3 → `recurring_monthly = 100 / 3 ≈ 33.33`.
5. Open the subscription list view → MRR column is summable.
6. Apply the search filter "With recurring revenue" → only subscriptions with positive MRR remain.

[...]

PR #1451, opened 2026-06-09
[19.0][IMP] subscription_oca: compute MRR and ARR

alvaro-domatix:
Adds MRR/ARR to subscriptions:

- `recurring_monthly` on each subscription line: `price_subtotal` normalised to a monthly amount from the template recurrence (30.4375 days/month average for daily/weekly cadences), converted to the company currency so subscriptions in different currencies can be compared and summed.
- `recurring_monthly` / `recurring_yearly` on the subscription: sum of the lines, ARR being 12 x MRR.

They are shown as a stat button on the form, optional sum columns on the list view and a "With recurring revenue" search filter.

Supersedes #1443.


szalatyzuzanna (commented):
I have a question regarding the scope of this feature. Since MRR/ARR are reporting-oriented KPIs rather than core subscription functionality, was a separate extension module considered instead of adding them directly to subscription_oca?

alvaro-domatix:
Thanks for raising the scope question. I considered a separate extension module, but I kept MRR/ARR in `subscription_oca` on purpose, for a few reasons:

1. **They are fields on the record itself, not a reporting overlay.** `recurring_monthly`/`recurring_yearly` are computed fields on `sale.subscription`, exposed directly in the form via stat buttons alongside `recurring_total`. They are part of the contract's own figures, the same way `recurring_total`/`recurring_next_date` already are — both are computed on the model today. Splitting MRR out would mean an extra module for two fields that belong to the same family of native computes.

2. **No new dependency is introduced.** The rating and SMS features were split into bridge modules precisely because they pull in `rating.mixin` / the SMS stack. MRR/ARR add none of that — it is a pure Python normalisation of the existing line subtotals (plus a currency conversion). So the usual reason to factor a feature out (to avoid forcing a heavy dependency on the core) does not apply here.

3. **Downstream depends on the field.** The reporting analysis (#1471) and the spreadsheet dashboards read `recurring_monthly`. Moving it to an extension would turn all of those into dependent modules and add an install step for a feature that is just two computed fields.

If anyone feels strongly that the KPI should not be forced on everyone, I can gate the *display* (stat buttons / list columns) behind a group or a setting rather than splitting the module — but the compute itself stays on the model. Happy to go either way; my recommendation is to keep it as is.

PR #1452, opened 2026-06-09
[19.0][IMP] subscription_oca: change customer wizard

alvaro-domatix:
Adds a wizard to change the customer of one or more subscriptions from the list/form action menu.

It refuses closed subscriptions, skips records whose customer does not actually change, optionally updates the pricelist and the recomputed fiscal position in the same write, and posts a note in the chatter of each subscription. Invoices already issued keep their original partner; only the subscription and its future invoices and orders are affected.

The action is restricted to sale managers, since changing the customer of a running contract is a sensitive operation.


jorgeglez1990 (commented):
A functional review has been carried out, and it was detected that when changing the customer to another one with a specific pricelist assigned, the line price is updated correctly, but the invoice total is not recalculated or updated.

It would be preferable to develop this functionality in a separate addon that inherits from the corresponding module, in order to avoid modifying the base module. 

Image

alvaro-domatix:
Hi @jorgeglez1990, thanks for catching this. I investigated and the root cause is in the core, not in the wizard itself: `sale.subscription._compute_total` only depended on the list of lines (`sale_subscription_line_ids`) and not on the lines' `price_subtotal`. Since `recurring_total`/`amount_total` are stored, any change to an **existing** line's price (a pricelist change, a manual edit, a discount update) recomputed the line subtotals but did not refresh the subscription totals. The change-customer wizard exposes it because it reassigns the pricelist on a subscription whose lines already exist.

The fix is in #1474 — `_compute_total` now depends on `price_subtotal`/`amount_tax_line_amount`, with a regression test. Once it lands, the wizard's totals will stay correct.

On the separate-addon point: the wizard is a small, self-contained extension of the base, and the bug above turned out to be pre-existing in the core rather than introduced by it; I'd lean towards keeping it here, but happy to split it out if you still see value in it.

PR #1453, opened 2026-06-09
[19.0][IMP] subscription_oca: support invoice and delivery addresses

alvaro-domatix:
Adds invoice and delivery addresses on subscriptions, defaulting to the customer's corresponding addresses and overridable per subscription, mirroring sale orders.

Both addresses are propagated to the recurring invoices (the move is addressed to the invoice address and records the delivery address; `commercial_partner_id` still rolls up to the contracting company, so the receivable stays on the parent) and to the sale orders generated by the subscription.

The fields show under the standard "Display Delivery / Invoice addresses" group.


jelenapoblet:
This improvement makes sense, please rebase the branch and update the PR

szalatyzuzanna (approved):
LGTM

jorgeglez1990 (approved):
LGTM

rrebollo:
@alvaro-domatix I think rebasing would help resolve the conflicts with the base branch.

subscription_oca/tests/test_subscription_partner_addresses.py
```diff
@@ -0,0 +1,118 @@
...
+            }
+        )
+        cls.shipping_address = cls.env["res.partner"].create(
+            {
+                "name": "Warehouse",
+                "type": "delivery",
+                "parent_id": cls.customer.id,
+            }
+        )
+        cls.pricelist = cls.env["product.pricelist"].create(
```
rrebollo: You are overwriting the pricelist variable from the core ProductCommon class (defined [here](https://github.com/odoo/odoo/blob/fb7fb13185f699a839b58a598d50fcc71c089380/addons/product/tests/common.py#L31)). Instead, you could either modify it in place (which is not recommended) or, better yet, create your own pricelist using a different variable name, such as test_pricelist, to avoid unintended side effects on the parent class's behavior.

subscription_oca/tests/test_subscription_partner_addresses.py
```diff
@@ -0,0 +1,118 @@
...
+            {"name": "Addr PL", "currency_id": cls.company.currency_id.id}
+        )
+        cls.template = cls.env["sale.subscription.template"].create(
```
rrebollo: I would prefix all the names of these variables to avoid collisions with variables defined higher up in the testing class hierarchy (Accounting < Product < UOM < BaseCommon).

subscription_oca/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 {
     "name": "Subscription management",
     "summary": "Generate recurring invoices.",
-    "version": "19.0.1.0.0",
+    "version": "19.0.1.1.0",
```
rrebollo: Generally, this is handled automatically by the OCA merge bot.

rrebollo (approved):
Code Review: Great work! The code looks good to me (LGTM). Thank you for your contribution! I've provided a few suggestions for your consideration—feel free to address them as you see fit.

alvaro-domatix:
Rebased on top of the current 19.0 (`tests/__init__.py` keeps both imports, manifest version aligned with 19.0).

@rrebrollo good catch — I've prefixed the test setUp variables (`addr_pricelist`, `addr_template`, `addr_customer`, ...) so they no longer shadow the inherited `ProductCommon.pricelist` or any other attribute of the common class hierarchy. CI is green.

pedrobaeza:
With all of these changes, you are converting `subscription_oca` into `contract` in complexity terms... I think we should focus in having just one module for the same feature.

jelenapoblet:
I don't think it will reach the same level of complexity as contract. To me, these improvements are still part of the core subscription domain and make subscription_oca more complete rather than turning it into a contract management module.

pedrobaeza:
I said "complexity" meaning the same feature set, and yes, you are doing just exactly that: putting the features that `contract` already has.

jelenapoblet:
I see your point. My understanding, however, is that the goal is to strengthen subscription_oca as the subscription engine rather than to keep it intentionally limited. I don't think the current improvements bring it to the same scope as contract, but I'd be interested in aligning on the long-term vision for both modules.

alvaro-domatix:
Rebased on top of the current 19.0 (`tests/__init__.py` now keeps `test_subscription_recurrence_dates` as well). The test variables are still prefixed with `addr_` as discussed. CI is green.

PR #1454, opened 2026-06-09
[19.0][ADD] subscription_oca_rating

alvaro-domatix:
New bridge module between `subscription_oca` and the `rating` addon.

`sale.subscription` inherits `rating.mixin`, so the standard rating statistics (last value, average, count, satisfaction percentage) become available on subscriptions. A header button sends a configurable rating request email to the customer, and the same request is sent automatically when the subscription is closed (both the close wizard and the cron go through it); a failure to send never blocks the close. The form gets a stat button to the ratings of the contract and the search view a "With ratings" filter.

Shipped as a separate addon to keep the `rating` dependency out of the core module.


jorgeglez1990 (commented):
LGTM Functional Review

jorgeglez1990 (approved):

pablolopez-domatix (approved):
LGTM

Issue #1091, opened 2024-06-10
[16.0] subscription_oca: error when creating subscription invoice (manually or cron) with "Next Invoice Date" value is empty


cuongnmtm:
## Module
subscription_oca

## Describe the bug
Error when the Next Invoice Date [recurring_next_date] field value is empty.

## To Reproduce
**Affected versions**: 16.0

1. Create a subscription and set value for fields.
2. Change the stage to "In progress" by clicking on the status bar. The field "Next Invoice Date" appears.
3. Empty the value of "Next Invoice Date", the field disappears permanently.
4. When the subscription cron running or when pressing the button "Create invoice", face the error below:

```
  File ".../oca-contract/subscription_oca/models/sale_subscription.py", line 219, in calculate_recurring_next_date
    self.recurring_next_date = start_date + relativedelta(
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: unsupported operand type(s) for +: 'bool' and 'relativedelta'
```

So there are two issues when the "Next Invoice Date" value is empty:
1. The field "Next Invoice Date" disappears permanently.
2. Error when creating subscription invoice (manually or cron).

Another weird behavior:
- When clicking on the button "Close subscription", a popup appears to input the close reason, then I cancel the popup, and the "Next invoice date" value is emptied. Then the above issue happens.

**Expected behavior**

- No error.
- Either "Next Invoice Date" should not be emptied when the subscription is not closed or a value for "Next Invoice Date" is set automatically to avoid issues.

**Additional context**

This is how the field is displayed on the view:
```xml
<field
    name="recurring_next_date"
    attrs="{'invisible': ['|', ('recurring_next_date', '=', False), ('in_progress', '=', False)]}"
/>
```

**Workaround**
Create a new subscription from scratch (duplicate subscription does not help).

rrebollo:
I think the real issue here is that even when a subscription is in progress, it still allows the recurring_next_date field to be emptied. Something very similar is still happening on version 18.

PR #1456, opened 2026-06-10
[18.0][FIX] subscription_oca: prevent cron crash and enforce data consistency

rrebollo:
### Problem

During a database upgrade, a subscription was found with `in_progress=True`
but no valid `recurring_next_date`. When the cron job `cron_subscription_management()`
processed this record, the comparison `False <= date.today()` raised `TypeError`,
crashing the entire cron and rolling back all invoices generated in that run.

### Root Cause (Potentially)

`close_subscription()` wrote `recurring_next_date = False` as a standalone
assignment before the `write({'stage_id': ...})` call, creating a split-write
window that left the record in an inconsistent state.

### What This Does

- Consolidates `close_subscription()` into a single atomic `write()` call
- Adds `@api.constrains` enforcing `in_progress=True` requires a consistent
  stage type and a valid `recurring_next_date`
- Hardens the cron with per-record savepoint isolation so a single failure
  never kills the entire job
- Adds a domain filter to skip closed subscriptions, tighter to
  `pre`/`in_progress` stages only
- Adds defensive guards on date comparisons to prevent `TypeError` on
  legacy bad data
- Aligns test fixtures (`sub7`/`sub8`/`sub9`) with correct `in_progress`
  semantics (stage type `"in_progress"`)
- Fixes `write()` override timing: inject `in_progress=False` before
  `super().write()` when target stage is not `"in_progress"`, so the
  constraint sees consistent state during flush

### Commits

1. `[FIX] subscription_oca: fix test fixtures and cron error test`
   Add `stage_in_progress` with correct type, update error test for new
   cron behavior.

2. `[FIX] subscription_oca: consolidate close_subscription atomic write`
   Merge all three fields into one `write()` to prevent split-write
   inconsistencies.

3. `[FIX] subscription_oca: add constraint for in_progress consistency`
   Enforce `in_progress=True` requires `stage_id.type="in_progress"`
   and a valid `recurring_next_date`.

4. `[FIX] subscription_oca: harden cron with savepoint and domain filter`
   Per-record savepoint isolation, domain filter for `pre`/`in_progress`
   stages, defensive guards on date comparisons.

5. `[FIX] subscription_oca: set in_progress=False in close_subscription`
   Ensure constraint sees consistent state during `write()`.

6. `[FIX] subscription_oca: prevent ValidationError when changing stage via write()`
   Inject `in_progress=False` before `super().write()` when the target
   stage type is not `"in_progress"`, preventing constraint errors on any
   stage change for in-progress subscriptions.

### Testing

All existing tests pass. Updated `test_subscription_oca_sub_cron_error`
to verify the cron catches and logs exceptions instead of propagating them.

@BinhexTeam T18387

### DIFF
--- a/subscription_oca/models/sale_subscription.py
+++ b/subscription_oca/models/sale_subscription.py
@@ -287,6 +287,7 @@ class SaleSubscription(models.Model):
         )
         self.write(
             {
+                "in_progress": False,
                 "recurring_next_date": False,
                 "close_reason_id": close_reason_id,
                 "stage_id": closed_stage.id,
--- a/subscription_oca/models/sale_subscription.py
+++ b/subscription_oca/models/sale_subscription.py
@@ -476,6 +476,12 @@ class SaleSubscription(models.Model):
         return False
 
     def write(self, values):
+        if "stage_id" in values and "in_progress" not in values:
+            stage = values["stage_id"]
+            if isinstance(stage, int):
+                stage = self.env["sale.subscription.stage"].browse(stage)
+            if stage.type != "in_progress":
+                values["in_progress"] = False
         res = super().write(values)
         if "stage_id" in values:
             for record in self:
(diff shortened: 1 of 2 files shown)


rrebollo:
Maybe this is related to #1091.

PR #1457, opened 2026-06-11
[19.0][IMP] contract: redesign the PDF report

bosd:
Split out from #1427 to keep changes reviewable in isolation (per @pedrobaeza's request to break the big PR into focused ones).

Redesigns the contract PDF report:
- structured info block (reference, start/end dates, responsible, payment terms)
- clearer section headings and proper HTML note rendering
- meaningful download filename (`Contract - <name>`)

**before:**
image

**After**
info block pdf renders correctly now in bubbule layout
image


pedrobaeza:
Can you put a screenshot of before/after?

PR #1459, opened 2026-06-11
[19.0][IMP] contract: harden the recurring-invoice cron

bosd:
Split out from #1427 (per @pedrobaeza's request to break the big PR into focused ones).

Hardens the recurring-invoice cron so one bad contract can no longer block a whole company's batch.

**Two-tier execution**
- Fast path — one batched `account.move.create([...])` per company (unchanged behaviour/performance).
- On failure — fall back to per-contract processing inside `cr.savepoint()`, so a single contract with bad data doesn't poison the transaction for the rest (SQL-level errors included).

**Per-contract failure trail**
- New fields `invoice_generation_error` / `invoice_generation_error_date` / `has_invoice_generation_error`.
- "Invoice Generation Failed" search filter, a warning icon + `decoration-warning` row tint in the list, and a form banner with a **Dismiss** button.
- Chatter post + de-duplicated TODO activity for the responsible user.
- Auto-clears on the next successful run.

**Tests:** isolation of a failing contract, error/activity population, recovery clearing, no duplicate chatter/activity, fast-path retention, and the dismiss/view actions.

### DIFF
--- a/contract/models/contract.py
+++ b/contract/models/contract.py
@@ -125,6 +125,22 @@ class ContractContract(models.Model):
     # === Dates ===
     date_end = fields.Date(compute="_compute_date_end", store=True, readonly=False)
 
+    # === Cron status ===
+    invoice_generation_error = fields.Text(
+        readonly=True,
+        copy=False,
+        help="Error message from the most recent failed scheduled invoicing run.",
+    )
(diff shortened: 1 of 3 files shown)


bofiltd (approved):
LGTM 👍

PR #1460, opened 2026-06-11
[19.0][FIX] contract: exclude archived products from the line picker

bosd:
Split out from #1427 (per @pedrobaeza's request to break the big PR into focused ones).

Excludes archived products from the contract line product picker.

Context: the `active_test: False` on `contract_line_ids` — added in d25115df ("No contract line and invoices when contract is archived") to keep archived contract *lines* visible — leaks into the embedded product search and surfaces archived products. Because a view-level `domain` attribute *replaces* the model field domain (they don't AND together), each customer/supplier line view re-asserts `('active', '=', True)`. Happy to switch to re-asserting `active_test: True` on the product field instead if preferred.

### DIFF
--- a/contract/models/contract_template_line.py
+++ b/contract/models/contract_template_line.py
@@ -33,7 +33,11 @@ class ContractTemplateLine(models.Model):
         comodel_name="res.partner", related="contract_id.partner_id"
     )
     # === Product & UOM ===
-    product_id = fields.Many2one("product.product", string="Product")
+    product_id = fields.Many2one(
+        "product.product",
+        string="Product",
+        domain=[("active", "=", True)],
+    )
     name = fields.Text(
         string="Description",
         required=True,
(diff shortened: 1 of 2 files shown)


contract/tests/test_contract.py
```diff
@@ -533,6 +533,16 @@ def test_last_invoice_pre_paid(self):
             "Should not create a new invoice after the last one",
         )
 
+    def test_archived_product_not_selectable(self):
```
pedrobaeza: This test is not adding value.

bosd: Agreed — removed it. The fix is UI-only (a field `domain` filters the picker but isn't enforced on write, so there's no server-side behaviour to assert), which made the test tautological. Dropped it.

BhaveshHeliconia (approved):
LGTM!

pedrobaeza (approved):

pedrobaeza:
Ouch, my bad, I have merged directly without ocabot (mixed with opened tabs in OpenUpgrade). Sorry.

PR #1462, opened 2026-06-12
[16.0][FIX] contract: remove circular dependency on recurring_next_date

maisim:
Break the cascade that overwrites user-edited recurring_next_date on contract lines whenever the contract was saved.

Three changes (aligned with the 18.0 logic):

1. abstract_contract_line: remove contract_id.recurring_next_date from @api.depends of _compute_recurring_next_date. Only contract_id.line_recurrence remains.

2. contract_line: remove _compute_next_period_date_start override that had cross-line dependencies (contract_id.last_date_invoiced, contract_id.contract_line_ids.last_date_invoiced).

3. contract_line: make date_end a plain (non-computed) field and remove _compute_date_end, breaking the cascade: contract.date_end -> line.date_end -> npds -> rnd.

Add regression tests.

Fixes https://github.com/OCA/contract/issues/1411

### DIFF
--- a/contract/models/abstract_contract_line.py
+++ b/contract/models/abstract_contract_line.py
@@ -168,8 +168,8 @@ class ContractAbstractContractLine(models.AbstractModel):
     def _compute_date_start(self):
         self._set_recurrence_field("date_start")
 
+    @api.depends("contract_id.line_recurrence")
     # pylint: disable=missing-return
-    @api.depends("contract_id.recurring_next_date", "contract_id.line_recurrence")
     def _compute_recurring_next_date(self):
         super()._compute_recurring_next_date()
         self._set_recurrence_field("recurring_next_date")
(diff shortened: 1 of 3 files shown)


contract/models/abstract_contract_line.py
```diff
@@ -169,10 +169,11 @@ def _compute_date_start(self):
         self._set_recurrence_field("date_start")
 
     # pylint: disable=missing-return
-    @api.depends("contract_id.recurring_next_date", "contract_id.line_recurrence")
+    @api.depends("contract_id.line_recurrence")
     def _compute_recurring_next_date(self):
-        super()._compute_recurring_next_date()
+        res = super()._compute_recurring_next_date()
         self._set_recurrence_field("recurring_next_date")
+        return res
```
Honeyxilia: I don't think the return here is necessary, as the compute method only needs to write the computed field and is not expected to return anything

Honeyxilia (commented):
Live testing OK (as a technical user)
Just a remark about a change

fcayre (approved):
Code review + live test ok, thanks !

Honeyxilia (approved):
Code review + Live test OK

rousseldenis (approved):

Honeyxilia:
thanks for the timely merge

PR #1466, opened 2026-06-19
[19.0][FIX] contract: Make portal test more resilient

eduezerouali-tecnativa:
cc @Tecnativa
If other test have already created that user will fail, better to use an specific portal user.
ping @pedrobaeza @victoralmau


pedrobaeza (approved):
Fast-tracking as obvious.

/ocabot merge patch

PR #1474, opened 2026-06-25
[19.0][FIX] subscription_oca: recompute totals on line subtotal change

alvaro-domatix:
Depends on nothing.

`_compute_total` only declared a dependency on the list of lines (`sale_subscription_line_ids`), not on the lines' subtotals. Since `recurring_total`, `amount_tax` and `amount_total` are stored, any change to an **existing** line's price (pricelist change, manual unit price edit, discount update) recomputed the line subtotals but did not trigger the parent totals, leaving them stale.

This is the root cause behind the issue reported in #1452 (change-customer wizard): when a subscription's pricelist is changed, line prices update correctly but `recurring_total` does not.

The fix declares the dependency on the aggregated line fields (`price_subtotal`, `amount_tax_line_amount`), so the subscription totals are refreshed whenever a line value changes. A regression test (edit the unit price of an existing line and assert the totals are refreshed) is included.

### DIFF
--- a/subscription_oca/models/sale_subscription.py
+++ b/subscription_oca/models/sale_subscription.py
@@ -182,7 +182,10 @@ class SaleSubscription(models.Model):
                 subscription.action_start_subscription()
                 subscription.generate_invoice()
 
-    @api.depends("sale_subscription_line_ids")
+    @api.depends(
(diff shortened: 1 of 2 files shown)


stferraro (approved):
LGTM !

PR #1475, opened 2026-06-26
[18.0][FIX] product_contract: Use the correct language in the sale order line to translate terms correctly

carlos-lopez-tecnativa:
When a product has a translated name, it is not correctly stored in the sale.order.line name field because the language context is lost in this function.

TT63447
@Tecnativa @pedrobaeza @CarlosRoca13 could you please review this?


CarlosRoca13 (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #1477, opened 2026-06-29
Updated from template

SirPyTech:
To include https://github.com/OCA/oca-addons-repo-template/pull/333.

PR #1478, opened 2026-06-29
[FIX] contract_sale_generation: Sale line violates check constraint

SirPyTech:
Steps:
1. Create a contract with "Sale" generation type
2. Add a line without a product
3. Generate a sale from the contract

Before this change:
The sale cannot be created because it raises:
> ValueError: <class 'psycopg2.errors.CheckViolation'>: "new row for relation "sale_order_line" violates check constraint "sale_order_line_accountable_required_fields"

After this change:
The line must have a product in order to save the contract.

Additional context:
The same issue was also addressed in #1100 but from a UI point of view

### DIFF
--- a/contract_sale_generation/models/contract.py
+++ b/contract_sale_generation/models/contract.py
@@ -16,6 +16,13 @@ class ContractContract(models.Model):
 
     sale_count = fields.Integer(compute="_compute_sale_count")
 
+    @api.constrains(
+        "generation_type",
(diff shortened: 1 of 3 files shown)


SirPyTech:
pre-commit will succeed after https://github.com/OCA/contract/pull/1477.

Tests fail because of https://github.com/OCA/contract/pull/1479.

PR #1482, opened 2026-07-02
[IMP]product_contract

bjouini-acsone:
this is a refactoring of the action_create_contract on sale order , so that it calls methods that we could inherit without touching the action_create_contract itself

### DIFF
--- a/product_contract/models/sale_order.py
+++ b/product_contract/models/sale_order.py
@@ -67,7 +67,7 @@ class SaleOrder(models.Model):
     def action_create_contract(self):
         contract_model = self.env["contract.contract"]
         contracts = []
-        lines_without_contract = self.env["sale.order.line"].browse()
+
         for rec in self.filtered("is_contract"):
             line_to_create_contract = rec.order_line.filtered(
                 lambda r: not r.contract_id and r.product_id.is_contract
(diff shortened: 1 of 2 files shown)


product_contract/models/sale_order.py
```diff
@@ -79,40 +79,52 @@ def action_create_contract(self):
...
+        for contract_template in contract_templates:
+            order_lines = (
+                lines_with_contract_template._filter_order_lines_from_contract_template(
+                    contract_template
+                )
+            )
+            contract = contract_model.create(
+                self._prepare_contract_value(contract_template)
+            )
+            contracts.append(contract.id)
```
tobiaszehntner, line 113: you gather all ids only to then send back browse(ids). Just gather the records directly
```suggestion
            contracts += contract
```

product_contract/models/sale_order.py
```diff
@@ -79,40 +79,52 @@ def action_create_contract(self):
...
+            contract._onchange_contract_template_id()
+            order_lines.create_contract_line(contract)
+            order_lines.write({"contract_id": contract.id})
```
tobiaszehntner, line 116: contract.id is already set as far as I can see: sale_order_line.py:80

tobiaszehntner (approved):
my comments are to improve, so maybe they shouldn't be done in this PR that is only about separating out a hook

product_contract/models/sale_order.py
sbejaoui, line 74: in cases like this, when needed to have a method per record to do 

```python
def action_create_contract(self):
    for rec in self.filtered("is_contract"):
        contracts += rec._create_contract()
    return contracts

def _create_contract(self):
    ...
    return contract
```

sbejaoui (changes requested):
Please check oca guidelines for commit message

https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#id57

PR #1485, opened 2026-07-15
[19.0][MIG] contract_sale_generation: Migration to 19.0

Rad0van:
Migration of `contract_sale_generation` to 19.0.

This continues the work started in #1333 (draft, CI red, marked stale and pending auto-close) by @Wodran14, doing the actual 19.0 API migration that was missing there.

### Changes

- `sale.order.line`: rename `product_uom` → `product_uom_id` (field renamed in 19.0).
- `_prepare_sale_line`: drop the `.new()` / onchange / `_convert_to_write` dance that relied on `sale.order.line._onchange_product_id_warning()` (removed in 19.0). The line values are now built directly, mirroring the base module's `contract.line._prepare_invoice_line()`; taxes and other computed values are resolved by the sale order on creation.
- Use the lazy `self.env._` translation helper.
- Tests:
  - `res.users.groups_id` → `group_ids` and `sale.order.line.tax_id` → `tax_ids` (fields renamed in 19.0).
  - Build the test data instead of relying on demo data (create the product locally instead of using `product.product_product_2`, and drop the chart template loading since `account` already provides a chart of accounts on install).
  - Keep the sale tax in the test fiscal position, because since 19.0 an empty fiscal position removes all taxes.

### Tests

Run locally on Odoo 19.0, all green **both with and without demo data** (`0 failed, 0 error(s) of 13 tests`).


FrancoMaxime (approved):
LGTM: code review

RaphaelVicini:
Hi, LGTM.

Can you update with fr.po traduction file from V18? thanks

tobiaszehntner (approved):
Code looks good to me

PR #1486, opened 2026-07-16
[18.0][IMP] contract_line_successor - stop() returns self in place of True

anthonissen-a:
This allows more flexibility when overriding the method.


bosd (approved):

qgroulard (approved):

bosd:
@sbejaoui :rocket:  ?

sbejaoui (changes requested):
stop() is already a multi-record method, not a model-level method, so self is available when overriding it, I don’t see what additional flexibility this change provides, could you clarify the use case you have in mind?

anthonissen-a:
If stopping one line automatically triggers stopping other related lines, returning self passes back the full list of everything actually stopped.

PR #1489, opened 2026-07-20
[19.0][MIG] contract_payment_mode: Migration to version 19.0

carlos-lopez-tecnativa:
TT58612
@Tecnativa @pedrobaeza @victoralmau @CarlosRoca13 could you please review this?


contract_payment_mode/hooks.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+def post_init_hook(env):
+    """Copy payment mode from partner to the new field at contract."""
+    m_contract = env["contract.contract"]
+    contracts = m_contract.search([("payment_mode_id", "=", False)])
+    if contracts:
+        _logger.info(f"Setting payment mode: {len(contracts)} contracts")
+    for contract in contracts:
+        payment_mode = contract.partner_id.customer_payment_mode_id
```
victoralmau:
```suggestion
        if contract.contract_type == "purchase":
           payment_mode = contract.partner_id.supplier_payment_mode_id        
        else:
           payment_mode = contract.partner_id.customer_payment_mode_id
```
IMO, it should be consistent with the `_compute_payment_mode_id()` method.

victoralmau (changes requested):
Code review OK

contract_payment_mode/hooks.py
carlos-lopez-tecnativa: I didn't take this code into account. Thanks for pointing it out.
I don't completely understand the purpose of this hook. If its purpose is to populate the field, the ORM already does that automatically when the new computed field is added.

So, if you agree, we can remove this code. What do you think?

CarlosRoca13 (approved):

eduezerouali-tecnativa (approved):
LGTM 

Image

contract_payment_mode/hooks.py
pedrobaeza: @carlos-lopez-tecnativa indeed the hook is not right, as it's not speeding up anything, but I think it's correct to have a hook that makes the installation fast enough if you have a lot of contracts.

carlos-lopez-tecnativa: I updated the hook to create and prefill the `payment_mode_id` field, and adapted the test accordingly. Please check it and let me know if it is okay with you.

pedrobaeza (approved):
/ocabot merge nobump

PR #1490, opened 2026-07-24
[18.0][FIX] contract: do not suggest archived products in contract lines

renzo1974:
The contract_line_ids / contract_line_fixed_ids one2many fields define context={"active_test": False} so that archived (closed) contract lines stay visible in the contract form. 

However, the web client propagates this context to every RPC issued from within the lines — including name_search on the product_id and uom_id many2one fields. As a result, archived products and UoMs show up in the dropdowns when adding or editing a contract line.

Steps to reproduce: 
archive a product → open a contract → add a line → search the product: the archived product is suggested.

This PR restores the default behaviour by explicitly setting active_test: True on those fields in the embedded line lists and in the line form dialog, while archived contract lines remain visible as intended.

### DIFF
--- a/contract/views/contract.xml
+++ b/contract/views/contract.xml
@@ -132,6 +132,7 @@
                                     <field
                                         name="product_id"
                                         widget="product_label_section_and_note_field"
+                                        context="{'active_test': True}"
                                     />
                                     <field
                                         name="name"
@@ -149,7 +150,10 @@
                                         groups="analytic.group_analytic_accounting"
                                     />
                                     <field name="quantity" />
-                                    <field name="uom_id" />
+                                    <field
+                                        name="uom_id"
+                                        context="{'active_test': True}"
+                                    />
                                     <field
                                         name="automatic_price"
                                         column_invisible="parent.contract_type == 'purchase'"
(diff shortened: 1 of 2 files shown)


contract/__manifest__.py
```diff
@@ -11,7 +11,7 @@
 
 {
     "name": "Recurring - Contracts Management",
-    "version": "18.0.2.5.0",
+    "version": "18.0.2.5.1",
```
pedrobaeza: Please don't change manually the module version. It can conflict if multiple PRs are done against the same module. This will be bumped by the bot on merging.

pedrobaeza (approved):

contract/__manifest__.py
renzo1974: Hi Pedro
Sorry for that! I have reset the version.

pedrobaeza:
Please squash both commits into one. And FYI, this is still modifying the version (first putting a version, and the reverting back), so the conflict will happen the same.

renzo1974:
> Please squash both commits into one. And FYI, this is still modifying the version (first putting a version, and the reverting back), so the conflict will happen the same.

done

pedrobaeza (approved):

victoralmau (approved):

pedrobaeza:
/ocabot merge patch

Can you fw-port it to 19?

PR #1492, opened 2026-07-29
[19.0][FIX] contract: do not suggest archived products in contract lines

renzo1974:
The contract_line_ids and contract_line_fixed_ids one2many fields set active_test=False in their context so that archived (closed) contract lines remain visible in the contract form. The web client propagates this context to every RPC issued from within the lines, including the name_search of the product_id and uom_id many2one fields. As a result, archived products and units of measure were suggested in the dropdowns.

Restore the default behaviour by explicitly setting active_test=True on these fields in the embedded line lists and in the line form dialog.

(cherry picked from commit 6459aa9b58c365ce52eaa9ae3448914b974b7625)

### DIFF
--- a/contract/views/contract.xml
+++ b/contract/views/contract.xml
@@ -132,6 +132,7 @@
                                     <field
                                         name="product_id"
                                         widget="product_label_section_and_note_field"
+                                        context="{'active_test': True}"
                                     />
                                     <field
                                         name="name"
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):
Thanks for the fw-port. We can fast-track it:

/ocabot merge patch

PR #1494, opened 2026-08-04
[18.0][ADD]contract_invoice_suspended

bjouini-acsone:


contract_invoice_suspended/README.rst
```diff
@@ -0,0 +1,6 @@
+==========================
+Contract Invoice Suspended
+==========================
+
+This module provides functionality to suspend automatic invoicing for
+contracts with configurable reasons and tracking.
```
qgroulard, line 6: Please follow oca guidelines: https://www.odoo-community.org/readme-structure

Could you please also elaborate? What's the point of this module? How to configure it? etc.
Try to think of this module as if it was completely unknown to you, what would you like to know about it?

contract_invoice_suspended/__manifest__.py
```diff
@@ -0,0 +1,25 @@
...
+    "version": "18.0.1.0.0",
+    "license": "AGPL-3",
+    "author": "ACSONE SA/NV,Odoo Community Association (OCA)",
+    "category": "contract",
+    "website": "https://github.com/OCA/contract",
+    "depends": [
+        # OCA
+        "contract_line_successor",
+        # OCA/contract
+        "contract_sale",
```
qgroulard, line 17: Why are these two separated?

contract_invoice_suspended/security/acl_auto_invoice_suspension_reason.xml
```diff
@@ -0,0 +1,47 @@
...
+            name="name"
+        >contract.automatic.invoice.suspension.reason multi-company rule</field>
+        <field
+            name="model_id"
+            ref="contract_invoice_suspended.model_contract_automatic_invoice_suspension_reason"
+        />
+        <field
+            name="domain_force"
+        >['|', ('company_id','=',False), ('company_id','parent_of',company_ids)]</field>
+    </record>
```
qgroulard, line 46: The new convention for multi-company rule is:
```
        [('company_id', 'in', company_ids + [False])]
```

contract_invoice_suspended/models/auto_invoice_suspension_reason.py
```diff
@@ -0,0 +1,83 @@
...
+        for suspension_reason in self:
+            fetched = d.get(suspension_reason.id)
+            if fetched is not None:
+                suspension_reason.suspended_reason_category_id = fetched
+            elif not suspension_reason.parent_id:
+                suspension_reason.suspended_reason_category_id = suspension_reason
+            else:
+                suspension_reason.suspended_reason_category_id = (
+                    suspension_reason.parent_id.suspended_reason_category_id
+                )
```
qgroulard, line 83: This looks super complicated to me.
Why can't we do something like:


```suggestion
    @api.depends("parent_id.suspended_reason_category_id")
    def _compute_suspended_reason_category_id(self):
        for suspension_reason in self:
            suspension_reason.suspended_reason_category_id = (
                suspension_reason.parent_id.suspended_reason_category_id
                or suspension_reason
            )
```

qgroulard (commented):
Hi, thanks for this contribution !

It would be nice to provide a bit more context 🙂

contract_invoice_suspended/__manifest__.py
bjouini-acsone: fixed

contract_invoice_suspended/security/acl_auto_invoice_suspension_reason.xml
bjouini-acsone: refactored

contract_invoice_suspended/models/auto_invoice_suspension_reason.py
bjouini-acsone: yes , refactored

contract_invoice_suspended/README.rst
bjouini-acsone: is it okay like this ?

qgroulard: Not quite, please see the documentation: https://www.odoo-community.org/readme-structure

There are dedicated fragments for "Usage" and "Configure", they should not be sub-section of the "Description" fragment.

contract_invoice_suspended/models/auto_invoice_suspension_reason.py
qgroulard: This was a genuine question. Are you 100% sure this is correct and efficient?

contract_invoice_suspended/README.rst
bjouini-acsone: refactored this , is this okay now ?

contract_invoice_suspended/models/auto_invoice_suspension_reason.py
bjouini-acsone: i tested it works fine also if i understood well the functionality also this test is paasing fine ,test_compute_suspended_reason_category_id

georgnume (commented):
Tested, seems working fine functionally

georgnume (approved):
Tested, seems working fine functionally

bosd:
Functional testing.. LGTM, few remarks.

Is there a specific reason why the `Automatic Invoice Suspension Reason` is placed under the sales menu?
The other contract modules, place it under invoicing / accounting.
Can you please place it there as well?

From UX perspective, one wants to see in the list view which contract is suspended, and be able to filter on them.

WDYT, Should suspended contracts still be shown under "In Progress"? Or better only in their own search filter "Suspended" contracts. 


What is the use case for having a Suspension reason an not being able to select if?
IMO, would propose to drop the field "Can be selected". Instead allow a suspension reason to be Archived.
More standard odoo flow/ux.

contract_invoice_suspended/models/auto_invoice_suspension_reason.py
```diff
@@ -0,0 +1,53 @@
...
+
+    name = fields.Char(required=True)
+    sequence = fields.Integer(default=10)
+    parent_id = fields.Many2one(
+        comodel_name="contract.automatic.invoice.suspension.reason",
+        string="Parent Auto Invoice Suspension Reason",
+        index=True,
+    )
+    suspended_reason_category_id = fields.Many2one(
+        "contract.automatic.invoice.suspension.reason",
```
tobiaszehntner, line 21: I don't understand the difference between parent_id and suspended_reason_category_id. They both point to a `contract.automatic.invoice.suspension.reason` record..? If it's needed it should be another model?

contract_invoice_suspended/models/contract.py
```diff
@@ -0,0 +1,62 @@
...
+        inverse="_inverse_is_auto_invoice_suspended",
+    )
+    auto_invoice_suspended_user_id = fields.Many2one(
+        string="Automatic invoicing suspended by",
+        comodel_name="res.users",
+        readonly=True,
+    )
+    auto_invoice_suspended_date = fields.Date(
+        string="Automatic invoicing suspended on", readonly=True
+    )
```
tobiaszehntner, line 24: why do we need these fields? is_auto_invoice_suspended is tracked, so the chatter will mention which user and at what date/time it was changed

contract_invoice_suspended/models/auto_invoice_suspension_reason.py
```diff
@@ -0,0 +1,53 @@
...
+        compute="_compute_suspended_reason_category_id",
+        string="Auto Invoice Suspension Reason Category",
+        store=True,
+        index=True,
+        recursive=True,
+    )
+    can_be_selected = fields.Boolean(default=True)
```
tobiaszehntner, line 28: Use odoo's standard field for this, then you don't have to add it in the domains
```suggestion
    active = fields.Boolean(default=True)
```

contract_invoice_suspended/models/auto_invoice_suspension_reason.py
```diff
@@ -0,0 +1,53 @@
...
+                rec.display_name = f"{rec.parent_id.name} - {rec.name}"
+            else:
+                rec.display_name = rec.name
+
+    @api.depends("parent_id.suspended_reason_category_id")
+    def _compute_suspended_reason_category_id(self):
+        for suspension_reason in self:
+            suspension_reason.suspended_reason_category_id = (
+                suspension_reason.parent_id.suspended_reason_category_id
+                or suspension_reason
```
tobiaszehntner, line 52: suspended_reason_category_id or suspension_reason 

that's where the category field doesn't make sense

contract_invoice_suspended/models/contract.py
```diff
@@ -0,0 +1,62 @@
...
+        return res
+
+    @api.depends("auto_invoice_suspended_user_id")
+    def _compute_is_auto_invoice_suspended(self):
+        self.filtered("auto_invoice_suspended_user_id").update(
+            {"is_auto_invoice_suspended": True}
+        )
+
+    def _inverse_is_auto_invoice_suspended(self):
+        if self.is_auto_invoice_suspended:
```
tobiaszehntner, line 49: use chatter for this (field is already tracked)

tobiaszehntner (changes requested):
the functionality of the module could be achieved with contract_invoice_manually (https://github.com/OCA/contract/tree/18.0/contract_invoice_manually) where you can mark contracts so they're not automatically invoiced (but there wouldn't be suspension reasons)

PR #1497, opened 2026-08-12
[REF] contract - pass self instead of relying on @api.model

anthonissen-a:
Replaces `@api.model` methods with instance methods across contract methods to improve extensibility and allow overrides to access record data in `self`.

Method parameters are now optional to ensure full backward compatibility.

### DIFF
--- a/contract/models/contract.py
+++ b/contract/models/contract.py
@@ -185,14 +185,7 @@ class ContractContract(models.Model):
                 and contract._origin.date_start != contract.date_start
                 or not recurring_next_date
             ):
-                contract.recurring_next_date = self.get_next_invoice_date(
-                    contract.next_period_date_start,
-                    contract.recurring_invoicing_type,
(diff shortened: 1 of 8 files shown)


qgroulard (approved):
Great changes 👍 

This is backward compatible and it highly improves the extensibility of `get_next_invoice_date` and `get_next_period_date_end` for other OCA modules and custom code.

tobiaszehntner (approved):
LGTM

sbejaoui (changes requested):
let see what @pedrobaeza thinks obout this

pedrobaeza (changes requested):
I don't think it's a good idea, as this is mixing both approaches: if you use `self`, then you should use it for all, but if there's going to be options for still passing arguments, then keep it `@api.model`, and you will get a predictive behavior, as having some arguments that if omitted, are taken from `self` implies some kind of "black magic".

PR #1498, opened 2026-08-12
[18.0][ADD] contract_invoice_offset

anthonissen-a:
This PR proposes an improvement to https://github.com/OCA/contract/pull/1366 by @bosd, building on top of the refactor https://github.com/OCA/contract/pull/1497 (which updated contract.recurring.mixin date methods to use self).

Key Changes:
- Replace overwrites with standard overrides: Utilizing self in mixin methods allows proper method overriding, improving modularity and maintainability for downstream modules.
- Remove contract_invoice_align_start dependency: Decoupled this dependency from the core module. It should instead be handled via a glue module.
- Layout improvements: Updated views to follow standard layout patterns used in other contract modules.


contract/models/contract_template_line.py
```diff
@@ -253,12 +253,6 @@ def _compute_recurring_interval(self):
     def _compute_date_start(self):
         self._set_recurrence_field("date_start")
 
-    @api.depends("contract_id.line_recurrence")
-    def _compute_recurring_next_date(self):
-        res = super()._compute_recurring_next_date()
-        self._set_recurrence_field("recurring_next_date")
-        return res
-
```
anthonissen-a, line 261: My understanding is that the parent contract's `recurring_next_date` is driven by the minimum `recurring_next_date` across its child lines. Why do individual contract lines need to pull or sync this value from the parent contract via `_set_recurrence_field`? What specific edge case or behavior does this pattern handle? I tested removing it, and no existing tests failed.

qgroulard: Hi @sbejaoui, could you please have a look at this question?

anthonissen-a:
Hi @bosd, if you are at the OCA days, we are at the contract table and would like to discuss this PR with you to try to merge it.

contract/models/contract_template_line.py
sbejaoui: This is needed when setting the `contract.line_recurrence field` to override recurring_next_date on all lines according to the contract, nothing happens when setting it to False

Maybe there is no test covering this case, which would explain why the tests didn't fail

PR #1501, opened 2026-08-20
[ADD] contract_invoice_auto_validate_send

tobiaszehntner:
This module extends *Contract Invoice Auto Validate*. When a contract
invoice is automatically validated, it is also sent to the customer
using their preferred method (email or Peppol), as configured on the
partner.

AI:
- Claude Opus 4.8: Conception and writing of code


acsonefho (approved):

bosd (commented):
Small and focused, and the core API use is right: the `sending_data` payload is exactly what core's own `account.move.send.batch.wizard` writes, and `is_sale_document(include_receipts=True)` correctly keeps vendor bills from purchase contracts out. Two things I would like to see addressed before this goes in.

### The send cron is never checked for being active

`models/contract_contract.py` writes `sending_data` and then calls `_trigger()`. When the cron is inactive, `ir.cron._trigger_list` drops the trigger without a word:

```python
if not self.sudo().active:
    at_list = [at for at in at_list if at > now]   # at_list == [now] -> becomes []
if not at_list:
    return self.env['ir.cron.trigger']
```

So on any database where `account.ir_cron_account_move_send` has been deactivated - routine in production - every auto-validated contract invoice gets `sending_data` written, `is_being_sent` stays True forever, and nothing is ever sent. No error, no log, no trace. Silent no-op is the worst outcome for an unattended path, and it is why core blocks on this: `account_move_send_batch_wizard.action_send_and_print` checks `account_move_send_cron.sudo().active` and raises a `RedirectWarning`/`UserError` *before* writing `sending_data`.

Raising is not right here, since it would break the invoicing cron for everyone. But the check should still happen:

```python
cron = self.env.ref("account.ir_cron_account_move_send", raise_if_not_found=False)
if not cron or not cron.sudo().active:
    _logger.warning(
        "Contract invoices are not sent: the 'Send invoices automatically' "
        "cron is not active."
    )
    return moves
to_send.sending_data = {...}
cron._trigger()
```

Invoicing keeps working, the invoices stay in a clean state, and the misconfiguration becomes visible. Worth a test with the cron deactivated.

### Recording the cron user as the sending author

`author_user_id` / `author_partner_id` are taken from `self.env.user`. On the path this module actually runs on - the contract invoicing cron - that is OdooBot, and `_cron_account_move_send` groups by `sending_data['author_partner_id']` to bus-notify the outcome. So every success and every failure notification is addressed to OdooBot, i.e. to nobody.

Attributing the send to OdooBot is not good practice: it puts a system account on record as the author of customer-facing correspondence, and it routes the only real-time feedback channel into a void. `_hook_if_errors` does post errors to the move chatter, so failures are recoverable after the fact, but nobody is told at the time.

The author should be a real, configurable partner - a company-level "invoice sender" setting alongside `auto_send_contract_invoice`, falling back to the company partner, would fit the existing configuration nicely.

### Checked and cleared

Access rights are fine. `ir.cron` is `group_system`-only in `ir.model.access.csv`, but `_trigger_list` never reads the un-sudoed record (`ensure_one()`, then `self.sudo().active` and `sudo().create`), so a billing user manually creating invoices on a contract will not hit an `AccessError`.

contract_invoice_auto_validate_send/models/contract_contract.py
```diff
@@ -0,0 +1,48 @@
...
+            lambda move: move.state == "posted"
+            and move.is_sale_document(include_receipts=True)
+            and move.company_id.auto_send_contract_invoice
+        )
+        if not to_send:
+            return moves
+        send_cron = self.env.ref(
+            "account.ir_cron_account_move_send", raise_if_not_found=False
+        )
+        if not (send_cron and send_cron.sudo().active):
```
acsonefho, line 26: IMO a cron is not a business object (but a technical object who call scheduled business code).
And the user is not supposed to enable/disable the cron depending on the behavior he wants.

Instead, you can create a configuration (on company maybe) to know if you have to send or not.

contract_invoice_auto_validate_send/models/contract_contract.py
```diff
@@ -0,0 +1,48 @@
...
+            # Queue the move for the standard asynchronous "Send & Print" cron,
+            # which sends the invoice with the customer's own method (email or
+            # Peppol) without blocking the invoicing cron.
+            move.sending_data = {
+                "author_user_id": self.env.user.id,
+                "author_partner_id": (
+                    company.auto_send_author_partner_id or company.partner_id
+                ).id,
+            }
+        send_cron._trigger()
```
acsonefho: That's strange to call a cron.
You probably should call the function called by the cron

acsonefho (changes requested):

contract_invoice_auto_validate_send/models/contract_contract.py
tobiaszehntner: We have a setting (auto_send_contract_invoice) and the user is not encouraged to change any cron. This is purely technical.

`move.sending_data` (json) is a technical field that is being treated by a cron. Here we just make sure to not set the field if the cron isn't running to treat it.

contract_invoice_auto_validate_send/models/contract_contract.py
tobiaszehntner: it's [what odoo does](https://github.com/odoo/odoo/blob/943a9429d7b8d859a6426145ffaa1b19f38879c1/addons/account/wizard/account_move_send_batch_wizard.py#L109) but I could change it to `self.env['account.move']._cron_account_move_send()`

tobiaszehntner: Just looked more into it and calling the function directly removes the chance of doing this asynchronously. `_trigger` actually queues this and wakes up cron workers. This is Odoo's built-in way while otherwise we would depend on queue_job. So considering how simple an intervention this is I would stick with the `send_cron._trigger()`

acsonefho (approved):

contract_invoice_auto_validate_send/models/res_company.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    auto_send_contract_invoice = fields.Boolean(
+        help="Send auto-validated contract invoices to the customer using their "
+        "preferred method (email or Peppol).",
+    )
+    auto_send_author_partner_id = fields.Many2one(
```
bosd: Great addition..
Automatic sended invoice can come from, an accounting team, instead of a user.
Like disucussed this is very depening on the use case.

Would propose to add a domain on this. Now you can select any partner. Also customers.

bosd:
My functional test of the invoice sending partner on runboat failed.

To replicate:
- Set an auto send partner `eg amy watson`
- Create contract with auto sent. (as admin user)
- Trigger invoicing cron.

Result: 
Invoice got sent out as admin user.
(which is desired behaviour. as it is the responsible on the contract.)
Bu that is actually not what is to be expected. When we force/override a user in the settings.

image
 

Proposed behaviour:
When auto send invoice, do it by default as the contract responsible.
(Standard expected odoo behaviour)

Optionally: 
Allow a user override to be defined in the config settings.
Since we do not have a business case for this. 
Please either drop the Sender override in the setting, or make it functional :)

PR #1504, opened 2026-08-25
[FIX] contract_line_successor: always call _prepare_value_for_stop() …

bjouini-acsone:
…on stop()

Before, when the date_end we want to set was not smaller than the line's current date_end, stop did not call _prepare_value_for_stop. It only wrote is_auto_renew and manual_renew_needed by hand instead.

This is a problem because other modules can override _prepare_value_for_stop() to add their own logic  If stop() does not call this method, all that logic is skipped and their values are never saved.

Now _prepare_value_for_stop is always called. We still never make date_end bigger than what it already is  we just make sure the method is called in all cases so custom values from other modules are not lost.

### DIFF
--- a/contract_line_successor/models/contract_line.py
+++ b/contract_line_successor/models/contract_line.py
@@ -399,21 +399,36 @@ class ContractLine(models.Model):
                 }
             )
 
-    def _prepare_value_for_stop(self, date_end, manual_renew_needed):
+    def _prepare_value_for_stop_base(self, date_end, manual_renew_needed):
+        """
+        Values that must always be applied when a line is stopped, whether
+        or not date_end itself is being pushed earlier.
+
+        This is the extension point for any module that needs to set values
+        on stop() even when the line's date_end doesn't move.
+        Modules should override this method to add their own
+        fields here rather , so those fields
+        are set in every case, not only when date_end is shortened.
+        """
         self.ensure_one()
         return {
-            "date_end": date_end,
             "is_auto_renew": False,
             "manual_renew_needed": manual_renew_needed,
-            "recurring_next_date": self.get_next_invoice_date(
-                self.next_period_date_start,
-                self.recurring_invoicing_type,
-                self.recurring_invoicing_offset,
-                self.recurring_rule_type,
-                self.recurring_interval,
-                max_date_end=date_end,
-            ),
         }
+
+    def _prepare_value_for_stop(self, date_end, manual_renew_needed):
+        self.ensure_one()
+        vals = self._prepare_value_for_stop_base(date_end, manual_renew_needed)
+        vals["date_end"] = date_end
+        vals["recurring_next_date"] = self.get_next_invoice_date(
+            self.next_period_date_start,
+            self.recurring_invoicing_type,
+            self.recurring_invoicing_offset,
+            self.recurring_rule_type,
+            self.recurring_interval,
+            max_date_end=date_end,
+        )
+        return vals
 
     def stop(self, date_end, manual_renew_needed=False, post_message=True):
         """
@@ -449,10 +464,7 @@ class ContractLine(models.Model):
                         rec.contract_id.message_post(body=msg)
                 else:
                     rec.write(
-                        {
-                            "is_auto_renew": False,
-                            "manual_renew_needed": manual_renew_needed,
-                        }
+                        rec._prepare_value_for_stop_base(date_end, manual_renew_needed)
                     )
         return True
 


contract_line_successor/models/contract_line.py
```diff
@@ -427,33 +427,31 @@ def stop(self, date_end, manual_renew_needed=False, post_message=True):
...
+                continue
+
+            old_date_end = rec.date_end
+            effective_date_end = date_end
+            if rec.date_end and rec.date_end <= date_end:
+                effective_date_end = rec.date_end
+
+            rec.write(
+                rec._prepare_value_for_stop(effective_date_end, manual_renew_needed)
+            )
```
qgroulard: I am confused, why does the call to `_prepare_value_for_stop()` has to change and be called with `effective_date_end` ?

qgroulard (commented):
Hi, thanks for the PR.
I agree that we would gain in extensibility by always calling  `_prepare_value_for_stop`. I believe the reasons for not calling it in the first place are not valid anymore (it doesn't arm anymore to write the `date_end` if it doesn't change).

However I wouldn't expect the call to `_prepare_value_for_stop` to change, see my comment.

contract_line_successor/models/contract_line.py
bjouini-acsone: we change it because _prepare_value_for_stop just take the date_end we give it and use it, it does not check anything but stop has a rule ,it should never make date_end bigger than what it already is 
before, this rule was done by skipping _prepare_value_for_stop when date_end was not smaller
so now we always call _prepare_value_for_stop, but to keep the rule, we compute effective_date_end first and give that instead of the raw date_end. Like this the rule is still respected, and all overrides still go correctly always

qgroulard: Ok then maybe the cleaner fix is to keep the current structure but take `date_end` out of `_prepare_value_for_stop`.

We could even have another method for the actual earlier stop which calls `_prepare_value_for_stop` and add `date_end` to the dict.
This way, we have two possible extension points: one generic and one for the case where we stop before the end date.

bjouini-acsone: @qgroulard you can  check this version

qgroulard: Yes I like it better but I wouldn't name this method "with_date_end" since the goal is to make it extensible in order to add more potential fields to update.

Btw, `recurring_next_date` is also part of this flow only. Although I am not sure we still need to update this field since it is automatically recomputed.

bjouini-acsone: changed the naming to _prepare_common_value_for_stop, is that okay ?

about the recurring_next_date, for me we should not remove this , it's okay that it's a computed field but also it's not a Readonly field , so for me the reason is that when we stop the line we should recalculate the value of the field even if the user have set the alue manually before (not sure if that's the reason of this implementation or not )

qgroulard: I agree to not touch `recurring_next_date`, however:

1. `recurring_next_date` should be updated in the "update date end" workflow, not the other (in order to keep the logic as before).
2. The "update date end" workflow should call the method named `_prepare_value_for_stop` in order to keep backward compatibility. Your new method should be the base one, and not the other way around.

bjouini-acsone: can you see this version

bofiltd (approved):

sbejaoui (approved):

contract_line_successor/models/contract_line.py
```diff
@@ -449,10 +464,7 @@ def stop(self, date_end, manual_renew_needed=False, post_message=True):
                         rec.contract_id.message_post(body=msg)
                 else:
                     rec.write(
-                        {
-                            "is_auto_renew": False,
-                            "manual_renew_needed": manual_renew_needed,
-                        }
+                        rec._prepare_value_for_stop_base(date_end, manual_renew_needed)
                     )
```
tobiaszehntner, line 445: @qgroulard I would suggest just one function but with a parameter, like `set_date_end` **unless** your client needs to specifically only call the `_base` function (adding values to stop-values without also setting the end_date).

In any case I would name the functions so it's clear **when** (what use-case) they're called.

Small improvement: only call the prepare-value in if/else, the write outside

```
if not rec.date_end or rec.date_end > date_end:
    values_for_stop = self._prepare_value_for_stop_base(date_end, manual_renew_needed, set_date_end=True)
    if post_message:
        ...
else:
    values_for_stop = self._prepare_value_for_stop_base(date_end, manual_renew_needed)
rec.write(values_for_stop)
```

tobiaszehntner (commented):
I would rename the functions so it's clear which one should be inherited for which case. Better still to use one with a parameter, but that depends on the use-case

contract_line_successor/models/contract_line.py
tobiaszehntner: ok we couldn't get to a better function. as long as it's clear it's the function being called *without setting an end date*

tobiaszehntner (approved):

contract_line_successor/models/contract_line.py
qgroulard: Yep, I agree the naming could be improved but IMO the method that would deserve a renaming is `_prepare_value_for_stop()`, but this one should stay backward compatible.

qgroulard (approved):

contract_line_successor/models/contract_line.py
```diff
@@ -449,10 +464,7 @@ def stop(self, date_end, manual_renew_needed=False, post_message=True):
                         rec.contract_id.message_post(body=msg)
                 else:
                     rec.write(
-                        {
-                            "is_auto_renew": False,
-                            "manual_renew_needed": manual_renew_needed,
-                        }
+                        rec._prepare_value_for_stop_base(date_end, manual_renew_needed)
```
sbejaoui: IMO, we don't need a new method, just call _prepare_value_for_stop with the same date_end

```suggestion
                        rec._prepare_value_for_stop(rec.date_end, manual_renew_needed)
```

contract_line_successor/models/contract_line.py
```diff
@@ -399,22 +399,37 @@ def _delay(self, delay_delta):
...
+        vals = self._prepare_value_for_stop_base(date_end, manual_renew_needed)
+        vals["date_end"] = date_end
+        vals["recurring_next_date"] = self.get_next_invoice_date(
+            self.next_period_date_start,
+            self.recurring_invoicing_type,
+            self.recurring_invoicing_offset,
+            self.recurring_rule_type,
+            self.recurring_interval,
+            max_date_end=date_end,
+        )
```
sbejaoui: this can be removed, it will be recomputed if needed

sbejaoui (changes requested):

contract_line_successor/models/contract_line.py
bjouini-acsone: refactored

contract_line_successor/models/contract_line.py
bjouini-acsone: done

contract_line_successor/models/contract_line.py
qgroulard: Hi @bjouini-acsone, the goal is to get rid of `_prepare_value_for_stop_base()`. Could you please remove it?

bjouini-acsone: you can check

PR #1507, opened 2026-08-31
[17.0][MIG] contract_sale_generation: Migration to version 17.0

adasatorres-tecnativa:
@Tecnativa TT64328
@christian-ramos-tecnativa 

Superseed #1234


christian-ramos-tecnativa (changes requested):
The "CREATE SALE" button style and text seems off
image
And the "Create invoices" action works even if the contract `generation_type` is set to sale, you can extend `_compute_create_invoice_visibility` to hide it and in `recurring_create_invoice` filter out sale contracts before calling `super()`

christian-ramos-tecnativa (changes requested):
I would use secondary button style so we keep it similar to the Create invoices, also the chatter message is not properly displayed:
Image

christian-ramos-tecnativa (approved):

contract_sale_generation/models/abstract_contract.py
```diff
@@ -12,4 +12,5 @@ class ContractAbstractContract(models.AbstractModel):
 
     generation_type = fields.Selection(
         selection_add=[("sale", "Sale")],
+        ondelete={"sales": "cascade"},
```
pedrobaeza: Incorrect key. It should be without final `s`.

contract_sale_generation/models/contract.py
```diff
@@ -131,3 +132,8 @@ def _get_recurring_create_func(self, create_type="invoice"):
     @api.model
     def cron_recurring_create_sale(self, date_ref=None):
         return self._cron_recurring_create(date_ref, create_type="sale")
+
+    def recurring_create_invoice(self):
+        return super(
+            ContractContract, self.filtered(lambda c: c.generation_type == "invoice")
+        ).recurring_create_invoice()
```
pedrobaeza: I like more this one for being more compact:

```suggestion
        self = self.filtered(lambda c: c.generation_type == "invoice")
        return super().recurring_create_invoice()
```

contract_sale_generation/tests/common.py
```diff
@@ -5,34 +5,33 @@
...
-            coa.try_loading(company=cls.env.company, install_demo=False)
+        chart_template = cls.env["account.chart.template"]._guess_chart_template(
+            cls.env.company.country_id
+        )
+        cls.env["account.chart.template"].try_loading(
+            chart_template, company=cls.env.company, install_demo=False
+        )
         cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
+        analytic_plan = cls.env["account.analytic.plan"].create({"name": "Test Plan"})
+
```
pedrobaeza, line 30: Don't add empty lines inside methods.

pedrobaeza (commented):
I suppose part of these improvements should go also to 18/19.

pedrobaeza (approved):
/ocabot merge nobump

PR #1511, opened 2026-09-02
[19.0][FIX] subscription_oca: fall back to journal account when product has no income account

alvaro-domatix:
A subscription line on a product without income account (neither the product nor its category defines one) previously forced account_id=False on the generated invoice line, which violates the database constraint account_move_line_check_accountable_required_fields (account_id IS NOT NULL). The management cron then failed five times in a row and Odoo deactivated it silently.

With this change, when no income account is found the key is omitted so the standard fallback applies: the line gets the journal default account, like any other sale invoice line.

- [x] Module installs and upgrades cleanly on 19.0
- [x] Existing tests pass (0 failed, 0 error)
- [ ] Reviewed by a maintainer

### DIFF
--- a/subscription_oca/models/sale_subscription_line.py
+++ b/subscription_oca/models/sale_subscription_line.py
@@ -304,7 +304,7 @@ class SaleSubscriptionLine(models.Model):
             self.product_id.property_account_income_id
             or self.product_id.categ_id.property_account_income_categ_id
         )
-        return {
+        vals = {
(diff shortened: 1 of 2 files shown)


alvaro-domatix:
Bumping for a maintainer look. @pedrobaeza

pedrobaeza:
I'm not a maintainer nor user of this module. You should become that figure IMO. Others using the module can review.

PR #1512, opened 2026-09-02
[18.0][IMP] contract - invoice lines up to date_ref

anthonissen-a:
Allows invoicing all recurring occurrences up to `date_ref` in a single invoice, rather than limiting the process to one occurrence per run.

Previously, the cron processed only a single occurrence per contract line execution. This change eliminates the need for multiple manual or scheduled runs when a contract has multiple backlog occurrences to process up to a given date.

Additionally, this decouples the single-occurrence limitation from `_prepare_recurring_invoices_values`, providing a cleaner entry point for custom overrides and tailored business logic.


victor-champonnois (approved):
Functional review OK
- If the Next period start is "far" in the past, the invoice generated from the cron accounts for all the period from the  Next period start to now. There is one invoice line per period. 
- The "Create invoices" button only invoice one at a time. 
- Cases where there is several invoice lines with different "Next period start" date are correctly handled

bosd (approved):
LGTM

PR #1513, opened 2026-09-04
[17.0][IMP] contract: New markers X-NEXT-MONTH

christian-ramos-tecnativa:
New markers to display next month dates, useful if you invoice pre-paid in the previous month

@tecnativa TT64062


bosd (changes requested):
@christian-ramos-tecnativa We solved this issue differently by using: https://github.com/OCA/contract/pull/1366

That way, one can configure the pre-payment. (We had a customer with in the month, pre-paid and pre-oaid+1 so all scenarios are covered). 
That implementation left it easier for the end user. No need to change the markers. The default ones just show the correct invoicing period.

pedrobaeza:
AFAIK, the offset already exists in the main module... @christian-ramos-tecnativa did you check it?

FrancoMaxime (approved):
LGTM: code review

FrancoMaxime (commented):
Wrong review from my part. I agree with Pedro's comment

PR #1515, opened 2026-09-05
[19.0][IMP] contract: label the date fields "Start Date" and "End Date"

bosd:
## Problem

`date_start` and `date_end` on `contract.recurring.mixin` carry no `string`, so Odoo derives the label from the field name. Every contract, contract line, contract template and contract template line form therefore reads **"Date Start"** and **"Date End"**.

That is the wrong word order in English, and it is not what Odoo does with the same pair of fields where it defines them itself:

```python
# fleet/models/fleet_vehicle_assignation_log.py
date_start = fields.Date(string="Start Date")
date_end = fields.Date(string="End Date")

# event/models/event_event.py
date_end = fields.Datetime(string='End Date', ...)
```

## Change

Set `string="Start Date"` / `string="End Date"` on the mixin, so all four models that inherit it pick the labels up in one place.

Three spots spelled the old wording out by hand and are updated to match:

- the **Date End** group-by filter on the contract search view
- the **Date Start** column header in the contract PDF report
- *Contract Date Start* in the contract reference mail template

Labels only - no behaviour change, no field renames, no view restructuring.

---

## Second commit: the columns could not be hidden either

Beyond the wording, neither date could be controlled from the line lists:

- `date_end` carried no `optional` attribute, so it sat outside Odoo's column selector and was **always shown, with no way to switch it off**.
- `date_start` was `column_invisible="True"`, so it was **never shown and could not be switched on**.
- 
The recurring line list is the exception: `date_start` is `required="1"` there and stays outside the selector, since hiding a required column would leave no way to fill it in an editable list.

---

## Third commit: the line dates read as plain dates

The contract list renders `date_end` and `recurring_next_date` with the `remaining_days` widget, so a date reads as "In 5 days". The contract line lists showed plain dates - most visibly in the report view behind **Contracts > Reporting > Contract Lines**, a whole table of deadlines with no sense of which are close.

The widget now applies to `date_end` in both standalone line lists, and to `recurring_next_date` in the report view where it is visible.

The lists embedded on the contract form are left alone on purpose: they sit next to the contract's own dates and already carry the lifecycle colours, so relative dates on every line would be noise.

### DIFF
--- a/contract/data/mail_template.xml
+++ b/contract/data/mail_template.xml
@@ -25,7 +25,7 @@
                     &amp;nbsp;&amp;nbsp;<strong>REFERENCES</strong><br />
                     &amp;nbsp;&amp;nbsp;Contract: <strong t-out="object.name" /><br />
                     <t t-if="object.date_start">
-                        &amp;nbsp;&amp;nbsp;Contract Date Start: <t
+                        &amp;nbsp;&amp;nbsp;Contract Start Date: <t
                     t-out="object.date_start or ''"
                 /><br />
                     </t>
--- a/contract/models/contract_recurring_mixin.py
+++ b/contract/models/contract_recurring_mixin.py
@@ -19,12 +19,15 @@ class ContractRecurringMixin(models.AbstractModel):
     _description = "Contract Recurring Mixin"
 
     date_start = fields.Date(
+        string="Start Date",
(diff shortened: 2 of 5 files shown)


FrancoMaxime (approved):
LGTM: code review

PR #1517, opened 2026-09-05
[19.0][IMP] contract: optional Reference/Company columns + optional Payment Terms column

bosd:
Spotted on the 19.0 runboat: the contract list has columns that cannot be toggled, and lacks a payment-terms column.

## Change

- `code` (Reference) and `company_id` (Company) get `optional="show"` — they render exactly as today, but a user can now drop them from the column selector (remembered per user). Previously they sat outside the selector with no way off.
- `payment_term_id` (Payment Terms) is added as an **optional column, hidden by default**, so it is available from the selector without cluttering the list.

`payment_mode_id` has the same toggle problem but is added by `contract_payment_mode`, so it is handled separately in #1518.

### DIFF
--- a/contract/views/contract.xml
+++ b/contract/views/contract.xml
@@ -419,6 +419,7 @@
                 <field name="code" optional="show" />
                 <field name="journal_id" groups="account.group_account_user" />
                 <field name="partner_id" />
+                <field name="payment_term_id" optional="hide" />
                 <field
                     name="tag_ids"
                     widget="many2many_tags"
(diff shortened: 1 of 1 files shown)


pedrobaeza (changes requested):
The company is hidden if you don't have multi-company group, so that should be enough. Put them `optional="show"`.

bosd:
Done, both are `optional="show"` now, so the list is unchanged and the columns just become droppable. Rewrote the commit message, which argued for the old default.

Fair point on Company - `base.group_multi_company` already keeps it away from everyone who does not need it, so there was no case for hiding it a second time.

pedrobaeza (approved):

PR #1518, opened 2026-09-05
[19.0][IMP] contract_payment_mode: hide Payment Mode by default in the contract list

bosd:
Spotted on the 19.0 runboat, alongside #1517 - same problem, different module.

## Problem

This module adds Payment Mode to the contract list without an `optional` attribute:

```xml
<field name="partner_id" position="after">
    <field name="payment_mode_id" />
</field>
```

Without `optional`, the column sits outside Odoo's column selector. Every database that installs this glue module gets it added to the contract list permanently, with no way to switch it off.

## Change

`optional="hide"` on the list column.

The payment mode is a detail of how a contract is billed rather than something you scan a list for, and a glue module should not spend a list column by default. Users who want it turn it on from the column selector, and the choice is remembered per user.

The form and search views keep the field exactly as they are.

### DIFF
--- a/contract_payment_mode/views/contract_view.xml
+++ b/contract_payment_mode/views/contract_view.xml
@@ -20,7 +20,7 @@
         <field name="inherit_id" ref="contract.contract_contract_tree_view" />
         <field name="arch" type="xml">
             <field name="partner_id" position="after">
-                <field name="payment_mode_id" />
+                <field name="payment_mode_id" optional="hide" />
             </field>
         </field>
     </record>


pedrobaeza (commented):
OK to be optional, but I think it should be `optional="show"` to keep it as it was before.

bosd:
Actually wanted to unify the contract view with the invoice view

https://github.com/OCA/bank-payment/blob/408d7e1c5ea74f79aebc4fa3bfbde43f893fc5ee/account_payment_mode/views/account_move_view.xml#L72-L75

pedrobaeza (approved):
OK, you have convinced me.

bofiltd (approved):

ajimenez-pyming (commented):
LGTM!

PR #1521, opened 2026-09-05
[19.0][IMP] contract: colour the contract list by term (expired muted, upcoming info)

bosd:
Spotted on the 19.0 runboat: the contract list gives no visual cue about a contract's term. An expired contract looks the same as a running one.

## Change

Colour the contract list by term, matching Odoo's draft/expired conventions:

- end date in the past → **muted** (expired)
- start date in the future → **info** (blue, like a draft order)

`date_start` / `date_end` are added as **hidden** columns so the decorations can evaluate; the visible layout is unchanged.

Companion to #1520 (which colours the contract *lines* by their lifecycle state).

### DIFF
--- a/contract/views/contract.xml
+++ b/contract/views/contract.xml
@@ -414,7 +414,13 @@
         <field name="name">contract.contract list view (in contract)</field>
         <field name="model">contract.contract</field>
         <field name="arch" type="xml">
-            <list>
+            <!-- Colour contracts by their term: an expired contract (end date in
+                 the past) is muted, one that has not started yet (start date in
+                 the future) is shown in the info colour, like a draft order. -->
+            <list
+                decoration-muted="date_end and date_end &lt; current_date"
+                decoration-info="date_start and date_start &gt; current_date"
+            >
                 <field name="name" string="Name" />
                 <field name="code" />
                 <field name="journal_id" groups="account.group_account_user" />


contract/views/contract.xml
```diff
@@ -414,7 +414,15 @@
...
         <field name="arch" type="xml">
-            <list>
+            <!-- Colour contracts by their term: an expired contract (end date in
+                 the past) is muted, one that has not started yet (start date in
+                 the future) is shown in the info colour, like a draft order. -->
+            <list
+                decoration-muted="date_end and date_end &lt; current_date"
+                decoration-info="date_start and date_start &gt; current_date"
+            >
+                <field name="date_start" column_invisible="True" />
```
pedrobaeza: I think you don't need to add these fields in this version.

contract/__manifest__.py
```diff
@@ -11,7 +11,7 @@
 
 {
     "name": "Recurring - Contracts Management",
-    "version": "19.0.1.0.4",
+    "version": "19.0.1.1.0",
```
pedrobaeza: Don't change module version (this smells to AI...)

contract/views/contract.xml
pedrobaeza: What about this?

bosd:
Good point, thanks, removed. The `decoration-muted`/`decoration-info` expressions auto-fetch `date_start`/`date_end` in the list, so the hidden columns weren't needed. The visible layout is unchanged.

pedrobaeza (approved):

bofiltd (approved):

pedrobaeza:
@bosd do you know who is @bofiltd ? It's a bit suspicious...

bosd:
Yes! We once did some projects together

pedrobaeza:
OK, but he/she can't come without history and approve without more comments 20 commits cc @etobella

sbejaoui (approved):

PR #1524, opened 2026-09-07
[ADD] contract_forecast: option to stop forecast at line date_end

bjouini-acsone:
Add a new company setting `contract_forecast_stop_at_line_date_end` that, when enabled, forces forecast period generation to stop at a contract line's date_end even when the line is set to auto-renew (tacit renewal)


georgnume (commented):
The job for forecasting does not run completely, cannot test to the end thus.
It has been tested internally on another DB, and it works fine there

FrancoMaxime (approved):
LGTM: code review

sbejaoui (approved):

PR #1526, opened 2026-09-15
[17.0][FIX] contract: Allow enabling Recurrence at line level

christian-ramos-tecnativa:
IMO this readonly restriction was mean to allow you changing from `general recurrence` to `at line level` even if there were invoices linked to the contract. The way it is right now allows you to check it but on save the field `line_recurrence` goes back to `False` moving the restriction to the `write` method fixes the issue.

@Tecnativa TT63901


pedrobaeza (commented):
Do it with a Python constraint instead.

pedrobaeza:
Anyway, keep the UI hint, but put `force_save=1`.

contract/views/contract.xml
```diff
@@ -141,7 +141,7 @@
                             <field
                                 name="line_recurrence"
                                 class="oe_inline"
-                                readonly="line_recurrence and invoice_count != 0"
```
pedrobaeza, line 144: But the readonly is OK.

contract/models/contract.py
```diff
@@ -181,6 +181,18 @@ def create(self, vals_list):
         return records
 
     def write(self, vals):
+        if (
```
pedrobaeza: What about putting this as Python ORM constraint?

christian-ramos-tecnativa:
Not sure if the field change can be handled with a SQL constraint as the `api.constrains` decorator doesn't exist for this version, do you have any example in mind @pedrobaeza?

contract/views/contract.xml
christian-ramos-tecnativa: Seems Ok at first but look at this behavior:
Peek 2026-09-15 13-57

pedrobaeza:
Hey, hey, why do you say `@api.constrains` doesn't exist in 17? This hasn't changed for a lot (since v8 I would say).

pedrobaeza:
Example in Odoo core: https://github.com/odoo/odoo/blob/17.0/addons/account/models/account_account.py#L26

contract/views/contract.xml
pedrobaeza: That's exactly why `force_save` is put, for saving such value. Something odd is happening then. Please debug.

christian-ramos-tecnativa:
Oh right!!!! My bad 😅

pedrobaeza (approved):

marcelsavegnago (approved):
LGTM

eduezerouali-tecnativa (approved):

PR #1527, opened 2026-09-16
[18][FIX] product_contract: pass recurrence_interval

JuanJaraSandoval:
The context sent to the contract configurator was missing default_recurrence_interval. It was lost when the context was moved to the contractContext getter.

As a result, the configurator always showed the model default (Month(s)) for the contract duration unit, and validating it wrote that value back to the sale order line. A product sold for "6 Year(s)" ended up as "6 Month(s)" on the line, and so did the contract end date once it was recomputed, e.g. when the start date is set at confirmation.


FrancoMaxime (approved):
LGTM: code review

sbejaoui (approved):

sbejaoui (approved):
functional test

PR #1528, opened 2026-09-18
[18.0][ADD] contract_minimum_term, product_contract_minimum_term

wpichler:
This PR adds two modules for contracts with a **minimum term**, as used for example by internet service providers and telecom operators: the customer commits to 24 months, after that the contract goes on until it is terminated.

- `contract_minimum_term`: minimum term handling on contracts. Depends on `contract_termination`.
- `product_contract_minimum_term`: glue module with `product_contract`, so the minimum term can be defined on the product and sold through sale orders. It is `auto_install`.

## Why

In the contract modules, a commitment can only be modelled with a **fixed end date** (`date_end`), optionally renewed automatically (`is_auto_renew` in `contract_line_successor`, "Duration" in `product_contract`). That is a different model than a minimum term:

| | Fixed duration + auto renew | Minimum term |
|---|---|---|
| Contract line | ends at `date_end`, a new period starts at each renewal | open-ended, no `date_end` |
| Meaning of "24 months" | the line ends after 24 months | the line can't be **terminated** before 24 months |
| Termination | the line is not renewed | the termination date must respect the minimum term and the termination notice |

Modelling a minimum term with a fixed end date + auto renewal makes the contract lines end and renew for no business reason, and the termination rules are not enforced anywhere. With a minimum term, the contract stays open-ended and only the **termination** is restricted. `contract_termination` already has the right place to enforce it.

## `contract_minimum_term`

**Configuration**: three values, each with a unit (days, weeks, months, years). 0 disables the value.

- **Minimum term**, e.g. 24 months, counted from the start date.
- **Renewal term**, e.g. 12 months: the minimum term is extended by this term if the contract is not terminated in time. Optional: with 0, the contract can be terminated at any time once the minimum term is over, respecting the notice. This is for example what the EU Electronic Communications Code (art. 105) requires for consumer telecom contracts after the initial commitment.
- **Termination notice**, e.g. 3 months: the termination has to be received at least this long before the end of the minimum term.

Like the recurrence fields, the values are defined on the contract, or on each line when `line_recurrence` is set (same `_set_recurrence_field` pattern). They are taken from the contract template and template lines.

**Computed dates** on the lines (stored, the contract shows the latest one):

- `min_term_date_end`: last day of the running minimum term.
- `min_term_notice_date`: last day on which a termination has to be received to end the contract at `min_term_date_end`.

A daily cron extends the terms whose notice deadline is over and logs it in the chatter. The terms are always computed from the start date: no drift on month ends, and a contract that started years ago gets the right current term straight away.

**Termination** (`contract_termination` wizard):

- New field *Termination Received On* (default: today), used for the notice check.
- The wizard proposes the earliest possible termination date and warns when the chosen date is earlier.
- `_terminate_contract` refuses an earlier date, except for the new group **Contract: Terminate Before Minimum Term** (implies *Can Terminate Contracts*). An early termination is flagged on the contract and logged in the chatter.
- The check computes the earliest date from the configuration and the reception date, not from the stored fields. It doesn't depend on the cron having run, and a termination letter received days before it is entered in Odoo is handled correctly.

A line can't have both a minimum term and `is_auto_renew` (constraint).

## `product_contract_minimum_term`

On a contract product, a **radio button** chooses how the contract duration is defined. The two options are exclusive:

- **Fixed Duration**: standard `product_contract` behavior (duration → `date_end`, optional auto renewal). This is the default, so existing products are not affected.
- **Minimum Term**: the duration and renewal fields are hidden, the minimum term, renewal term and termination notice are shown instead.

The choice and the values go to the sale order line and the contract configurator (same radio button, JS patch of the configurator like `product_contract_variable_quantity`). They are then passed to the contract lines created at confirmation: no `date_end`, no auto renewal. Lines with a fixed duration explicitly get no minimum term, even if the contract template defines one. Auto renewal and a minimum term can't be combined on the product or the sale order line.

## Tests

- `contract_minimum_term`: 18 tests. Contract and line level values, template, computation and renewal (compute and cron), month ends, termination refused / allowed / allowed with the group, notice deadline, wizard.
- `product_contract_minimum_term`: 5 tests. Product, sale order line, configurator, contract creation from a sale order with a minimum term line and a fixed duration line.
- The existing tests of `contract`, `contract_line_successor`, `contract_termination` and `product_contract` pass with both modules installed.

The modules are in two separate commits. I'm happy to split them into two PRs if you prefer, the glue module depending on the first one.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


contract_minimum_term/models/contract_contract.py
```diff
@@ -0,0 +1,127 @@
...
+        for rec in self:
+            lines = rec._get_min_term_lines().filtered("min_term_date_end")
+            last_line = lines.sorted("min_term_date_end")[-1:]
+            rec.min_term_date_end = last_line.min_term_date_end
+            rec.min_term_notice_date = last_line.min_term_notice_date
+
+    def _get_min_term_lines(self):
+        self.ensure_one()
+        return self.contract_line_ids.filtered(
+            lambda line: not line.display_type and not line.is_canceled
```
tobiaszehntner: I wonder if this could only concern lines with min_term_date_end, ie. actively adding the ones you want instead of excluding (you filter on those in the compute but you seem to need all of them in `_get_min_term_contract_termination_date`). My thinking is to make it work with other modules without needing a bridge. I have for example a module that introduces 'is_deferred' on the line which would have to be excluded. Of course it's an easy bridge with this hook, but if it could always concern only contract lines with a minimum term it wouldn't need anything. Not entirely sure you can make it work, so leaving that up to you.

contract_minimum_term/models/contract_line.py
```diff
@@ -0,0 +1,81 @@
...
+    def _get_min_term_renewal_domain(self):
+        return [
+            ("min_term_notice_date", "<", fields.Date.context_today(self)),
+            ("min_term_renewal_interval", ">", 0),
+            ("is_canceled", "=", False),
+            ("contract_id.is_terminated", "=", False),
+        ]
+
+    @api.model
+    def cron_renew_minimum_term(self):
```
tobiaszehntner:
```suggestion
    def _cron_renew_minimum_term(self):
```

contract_minimum_term/models/contract_line.py
```diff
@@ -0,0 +1,81 @@
...
+        lines = self.search(self._get_min_term_renewal_domain())
+        old_dates = {line: line.min_term_date_end for line in lines}
+        # The dates depend on the current date: force their recomputation and
+        # propagate the new values to the dependent fields (contract dates)
+        fnames = ["min_term_date_end", "min_term_notice_date"]
```
tobiaszehntner: maybe separate that out into a hook?

contract_minimum_term/models/contract_minimum_term_mixin.py
```diff
@@ -0,0 +1,115 @@
...
+RULE_TYPES = [
+    ("daily", "Day(s)"),
+    ("weekly", "Week(s)"),
+    ("monthly", "Month(s)"),
+    ("yearly", "Year(s)"),
+]
+
+
+class ContractMinimumTermMixin(models.AbstractModel):
+    _name = "contract.minimum.term.mixin"
```
tobiaszehntner, line 31: product_contract and contract_line_successor both already have renewal periods. Could the contract not go to that after the minimum term is fulfilled, in order to not introduce yet another interval?

product_contract_minimum_term/models/product_template.py
```diff
@@ -0,0 +1,49 @@
...
+        default="fixed",
+        required=True,
+        help="Fixed Duration: the contract line ends after the duration, it can "
+        "be renewed automatically.\n"
+        "Minimum Term: the contract line is open-ended, it cannot be "
+        "terminated before the end of the minimum term, which is extended by "
+        "the renewal term if the contract is not terminated in time.",
+    )
+    min_term_rule_type = fields.Selection(default="monthly")
+    min_term_renewal_rule_type = fields.Selection(default="monthly")
```
tobiaszehntner: to only have one source of truth instead of declaring the same string multiple times, I like to declare the selection values as static (in the mixin) and import it here and other places. But not required.

product_contract_minimum_term/models/sale_order_line_contract_mixin.py
```diff
@@ -0,0 +1,88 @@
...
+        store=True,
+        readonly=False,
+    )
+    min_term_renewal_interval = fields.Integer(
+        compute="_compute_product_contract_minimum_term",
+        precompute=True,
+        store=True,
+        readonly=False,
+    )
+    min_term_renewal_rule_type = fields.Selection(
```
tobiaszehntner: sale.order.line.contract.mixin has already 
```
auto_renew_rule_type = fields.Selection(
        [
            ("daily", "Day(s)"),
            ("weekly", "Week(s)"),
            ("monthly", "Month(s)"),
            ("yearly", "Year(s)"),
        ],
```
so it seems weird to have a separate renewal field..?

tobiaszehntner (changes requested):
Main concern is about introducing new renewal fields and not using existing ones. In my opinion, after a contract's minimum term has ended, it should go over to standard (or other contract module's renewal) functions for the continuation. So these modules only handle the minimum term, not the period afterwards.

wpichler:
Thanks a lot for the thorough review!

**Renewal fields.** You are right that the renewal term should not be a new
concept, and I will reuse the existing fields:

- `is_auto_renew` + `auto_renew_interval` / `auto_renew_rule_type` for the
  renewal of the minimum term
- `termination_notice_interval` / `termination_notice_rule_type` (also already
  there) instead of `min_term_notice_*`

That leaves `min_term_interval` / `min_term_rule_type` as the only new fields,
on the contract (line) and, through the glue module, on the product and the
sale order line. The current constraint "a minimum term excludes automatic
renewal" then becomes "a minimum term excludes a fixed end date", which is the
real restriction: with `date_end` set, both mechanisms would fight over the
same line.

**Handing over to the standard renewal after the minimum term.** I looked into
that, but I do not think it can work, because the existing renewal is built
around `date_end`:

- `_compute_termination_notice_date` returns `False` without a `date_end`
- `_contract_line_to_renew_domain` filters on `termination_notice_date <= today`,
  so an open-ended line is never picked up
- `_get_renewal_new_date_end` starts from `self.date_end`

A line with a minimum term has no `date_end` at all -- that is the point of the
module: the contract stays open-ended, only the commitment is renewed. So there
is nothing for `renew()` to extend, and if it ran it would give the line an end
date, which is exactly what we want to avoid. What these modules add is
therefore only "until when can this line not be terminated", never an end date,
and everything after the minimum term is plain open-ended contract behaviour.

If you would rather have no renewal of the commitment at all in this module,
that is possible too, but it would leave out the common case in telecom/ISP
contracts: 24 months, then 12 more unless terminated 3 months in advance.

**`_get_min_term_lines`.** Good point, I will turn it around and select the
lines that actually have a minimum term (`min_term_date_end` set, and not
cancelled) instead of excluding the others. Lines that keep their
`min_term_date_end` in the past are still covered, so the notice period after
the end of the minimum term keeps working, and your `is_deferred` lines need no
bridge. Note that this also means a line without a minimum term is no longer
restricted by a termination notice alone -- which fits the reuse of
`termination_notice_*` above, since every product carries a default notice of
one month.

**`_cron_renew_minimum_term`** and **the hook**: both done. The recomputation
moves into `_renew_minimum_term()` on the lines, and the cron only searches and
calls it. (I had followed `cron_renew_contract_line` from
`contract_line_successor` for the name; happy to use the private one.)

**Selection values.** `CONTRACT_DURATION_TYPES` is already a module-level
constant that the sale order line mixin imports, and the rule types live in
`contract_minimum_term` as `RULE_TYPES`; the lines you pointed at only set a
different default. If you meant something else there, let me know.

wpichler:
Rework pushed (force-push, still two commits, one per module).

**Renewal and notice now reuse the existing fields.** `min_term_renewal_*` and
`min_term_notice_*` are gone. What is left as new configuration is
`min_term_interval` / `min_term_rule_type` on the contract, the template, the
product and the sale order line.

- `is_auto_renew` + `auto_renew_interval` / `auto_renew_rule_type`: on a line
  with a minimum term these renew the **minimum term**, not an end date.
- `termination_notice_interval` / `termination_notice_rule_type`: the notice
  to respect, as before.

That reuse needed two small adaptations in `contract_minimum_term`, both
scoped to lines that actually have a minimum term:

- `_check_allowed` of `contract_line_successor` requires an end date on an
  auto-renewing line. A minimum term line is open-ended on purpose, so the
  check now only applies to the other lines (the "successor" part still
  applies to all).
- `get_allowed()` has no entry for `has_date_end=False, is_auto_renew=True`,
  which until now could not exist, so such a line was allowed nothing at all
  (no stop, no cancel, and therefore no termination). `_compute_allowed`
  evaluates those lines as open-ended lines without auto renewal, which is
  what they are as far as the end date is concerned.

The constraint is now "a minimum term cannot have an **end date that is
renewed automatically**" rather than "cannot have an end date": stopping or
terminating a line sets `date_end` and switches `is_auto_renew` off, and that
has to stay possible.

**`_get_min_term_lines`** now selects the lines that have a minimum term
(`min_term_date_end` set, not cancelled) instead of excluding lines. As
announced, a line without a minimum term is no longer restricted by a
termination notice alone.

**`_cron_renew_minimum_term`** renamed, and the forced recomputation moved
into `_renew_minimum_term()` on the lines; the cron only searches and calls
it.

**Views**: the minimum term of a line sits on the existing *Renewal* page next
to `is_auto_renew` and the notice, and that group is now also shown when there
is a minimum term without auto renewal. On the product, *Renewal* and
*Termination Notice* stay visible in both duration types, only the fixed
*Duration* is hidden for a minimum term.

**Tests**: both test classes are now `@tagged("post_install", "-at_install")`.
Thanks for that catch, it also fixes a failure in a database where the glue
module is installed. 18 + 6 tests, and the tests of `contract`,
`contract_line_successor`, `contract_termination` and `product_contract` pass
with both modules installed (168 in total).

I left the selection constants as they are, as discussed; let me know if you
meant something else there.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

wpichler:
One more commit (no force-push this time): the minimum term now also applies to
a **single contract line**. Until now only `contract.contract._terminate_contract`
was checked, so stopping or cancelling a line was a way around the commitment —
all the more since `_compute_allowed` has to allow those actions on an
open-ended auto renewing line for the termination to work at all.

- `stop()` refuses an end date before the earliest possible termination date,
  computed exactly like in the wizard (`_get_min_term_earliest_termination_date`),
  with the same exception for the group *Terminate Before Minimum Term* and a
  chatter entry. The termination of the whole contract passes its reception date
  down, so both levels agree and only one message is logged.
- `cancel()` is only possible while the line has not started (no start date, or
  a start date in the future). A running line is stopped, not taken back. A user
  of that group may still cancel a started line that was **never invoiced**, to
  correct a mistake; `is_cancel_allowed` itself stays user-independent and only
  carries the "not started yet" rule.
- New context key `min_term_skip_checks` for modules that replace a line by a
  successor instead of ending it. `product_contract_minimum_term` uses it for
  the upsell/downsell of `product_contract`, where the replacement line carries
  the commitment on.

24 + 6 tests, everything green together with `contract`, `contract_line_successor`,
`contract_termination` and `product_contract` (174).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

contract_minimum_term/models/contract_line.py
```diff
@@ -0,0 +1,272 @@
...
+        line. That check only applies to lines without a minimum term, the
+        others stay open-ended on purpose.
+        """
+        min_term_lines = self.filtered(lambda rec: rec.min_term_interval > 0)
+        res = super(ContractLine, self - min_term_lines)._check_allowed()
+        for rec in min_term_lines:
+            if rec.is_auto_renew and rec.successor_contract_line_id:
+                raise ValidationError(
+                    self.env._(
+                        "A contract line with a successor can't be set to " "auto-renew"
```
tobiaszehntner: please review what you push manually
```suggestion
                        "A contract line with a successor can't be set to auto-renew"
```

tobiaszehntner (approved):
I don't mind if you write your code with AI, but it would be great to have a human writing and responding to comments.

contract_minimum_term/models/contract_line.py
wpichler: Hi Tobias,

thanks for your response - you are right - i will take a closer look at what i push. Next Force Push is on the way...

Thanks for your support...

I do also have an other module which could be interesting - contract_periods - this module does force contracts to align to invoicing periods (first of month for example) - and does implement aliquot calculation, and also periods in the invoice (so if you create a new contract which started some months ago, and you selected monthly billing, it would create one invoice - with invoice lines for each period). Is there already such functionality in the repo (i havn't seen something like that) - should i also create a PR for it ?

bosd:
@wpichler Can you please have a look at PR #1374 
It has a overlap in use case. Altough being on odoo 19 and a more simple implementation.
Let's see if we can have some common ground.

IMO, the Termination notice deadline is a function that should be in it's own module. As it is not strictly related to the minimum duration. 
(Also in my use case, we want silent renewal. While UX is already disastrous, on these modules, not willing to have another field added by default, which will never be used. )

This PR is more advanced and supports minimum duration on the line level.


I tried to test this module on the runboat. But was not able to find the button "Terminate Contract" according to the readme. Tried both line level and header contracts.

For now, focused mainly on the `contract_minimum_term`, Is the `product_contract_minimum_term`  adding any additional functionality, which is not already covered by setting these fields in the contract template?

bosd:
Also might want to consider using the widgets on the fields. e.g. 3 days left before min termination date.

PR #1532, opened 2026-09-21
[18.0][FIX] contract_line_successor: create successor line with sudo(…

samirGuesmi:
Creating a successor contract line can trigger internal mail.thread operations during the ORM flush, which may attempt to update protected mail.message fields such as model and res_id.

This previously resulted in the following error:
AccessError(_("Only administrators can modify 'model' and 'res_id' fields."))

The successor line creation is now executed with sudo privileges.


samirGuesmi:
@qgroulard @sbejaoui

samirGuesmi:
this replace PR: https://github.com/OCA/contract/pull/1382

sbejaoui (approved):

